PR #74, opened 2015-08-29
Added FPU IP support to unicorn

mothran:
I added FPIP to the fnstenv write and updated the fpu_helpers.c to only update the FPIP for non-control instructions. I also added a regression test in regress/fpu_ip.py with expect output as a comment.

This enables the unicorn engine to correctly survive the shikata_ga_nai (http://www.rapid7.com/db/modules/encoder/x86/shikata_ga_nai) shellcode encoder and other FPU GETIP methods.

I only added the write for 32 bit mode in and not 16 because I have not been able to read up on 16 bit mode FPU implementation.

### DIFF
--- a/qemu/target-i386/fpu_helper.c
+++ b/qemu/target-i386/fpu_helper.c
@@ -986,7 +986,7 @@ void helper_fstenv(CPUX86State *env, target_ulong ptr, int data32)
             }
         }
     }
-#ifdef TARGET_X86_64
+
     // DFLAG enum: tcg.h, case here to int
     if (env->hflags & HF_CS64_MASK) {
         cpu_stl_data(env, ptr, env->fpuc);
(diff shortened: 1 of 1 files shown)


qemu/target-i386/translate.c
```diff
@@ -6110,6 +6115,7 @@ static target_ulong disas_insn(CPUX86State *env, DisasContext *s,
                         /* fcomp needs pop */
                         gen_helper_fpop(tcg_ctx, cpu_env);
                     }
+                    fpu_update_ip(env, pc_start);
```
wtdcode, line 6118: This fix is wrong according to the discussion at https://lists.nongnu.org/archive/html/qemu-devel/2021-04/msg00307.html.

I have submitted a patch at https://lists.nongnu.org/archive/html/qemu-devel/2021-04/msg02868.html.

Just a ping and reminder for this issue.

PR #198, opened 2015-10-17
tests: create libunicorn.so.0 symlink

JonathonReinhart:
This hack allows the unit tests to link against libunicorn, after
the SONAME was added in 4860fdb3.

Becuase libunicorn.so has an SONAME of libunicorn.so.0, the linker uses
the SONAME for the DT_NEEDED entry of the tests. But because a library
with that name does not exist, they would fail to run. This symlink
works around that issue, without touching the toplevel Makefile.

### DIFF
--- a/tests/unit/Makefile
+++ b/tests/unit/Makefile
@@ -14,7 +14,7 @@ clean:
 	rm -rf ${ALL_TESTS}
 
 .PHONY: test
-test: export LD_LIBRARY_PATH=../../
+test: export LD_LIBRARY_PATH=.
 test: ${ALL_TESTS}
 	./test_sanity
 	./test_x86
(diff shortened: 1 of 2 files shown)


tests/unit/Makefile
```diff
@@ -23,9 +23,12 @@ test: ${ALL_TESTS}
 test_sanity: test_sanity.c
 test_x86: test_x86.c
 test_mem_map: test_mem_map.c
+${ALL_TESTS}: libunicorn.so.0
```
aquynh: have you considered that on OSX, the shared lib extension is `.dylib`, not `.so`?

JonathonReinhart:
This is why PR's are good, so a discussion can be had.

Reasons for not wanting to compile the tests statically:
- Larger binary size (yes these are just for development)
- Require static version of libcmocka to be installed
- I couldn't get it to work :stuck_out_tongue: 

It really doesn't seem like it should be a very big deal to have unit tests that link dynamically against the project's dynamic library. What can we do to make this work?

As @lunixbochs indicated elsewhere, CMake produces the same symlink structure in the build directory that one would expect in `/lib`.  Because it seems like we run into little Makefile-things like this somewhat often, I have submitted #200 for discussion.

Issue #206, opened 2015-10-19
Compilation fails if python3 is the default interpreter


williballenthin:
Compilation via `make.sh` fails if Python 3.x is the default interpreter. Seems qemu requires Python 2.x.

error:
```
./configure --extra-cflags="-DUNICORN_HAS_X86 -DUNICORN_HAS_ARM -DUNICORN_HAS_M68K -DUNICORN_HAS_ARM64 -DUNICORN_HAS_MIPS -DUNICORN_HAS_MIPSEL -DUNICORN_HAS_MIPS64 -DUNICORN_HAS_MIPS64EL -DUNICORN_HAS_SPARC -fPIC -fvisibility=hidden" --target-list="x86_64-softmmu, arm-softmmu, m68k-softmmu, aarch64-softmmu, mips-softmmu, mipsel-softmmu, mips64-softmmu, mips64el-softmmu, sparc-softmmu,sparc64-softmmu," 

ERROR: Cannot use 'python', Python 2.4 or later is required.
       Note that Python 3 or later is not yet supported.
       Use --python=/path/to/python to specify a supported Python.
```

I got it working using the following minor patch:

```
diff --git a/Makefile b/Makefile
index 4ce6473..ff4b15f 100644
--- a/Makefile
+++ b/Makefile
@@ -206,11 +206,11 @@ config:
 qemu/config-host.h-timestamp:
 ifeq ($(UNICORN_DEBUG),yes)
        cd qemu && \
-       ./configure --extra-cflags="$(UNICORN_CFLAGS)" --target-list="$(UNICORN_TARGETS)" ${UNICORN_QEMU_FLAGS}
+       ./configure --python="$(which python2)" --extra-cflags="$(UNICORN_CFLAGS)" --target-list="$(UNICORN_TARGETS)" ${UNICORN_QEMU_FLAGS}
        printf "$(UNICORN_ARCHS)" > config.log
 else
        cd qemu && \
-       ./configure --disable-debug-info --extra-cflags="$(UNICORN_CFLAGS)" --target-list="$(UNICORN_TARGETS)" ${UNICORN_QEMU_FLAGS}
+       ./configure --python="$(which python2)" --disable-debug-info --extra-cflags="$(UNICORN_CFLAGS)" --target-list="$(UNICORN_TARGETS)" ${UNICORN_QEMU_FLAGS}
        printf "$(UNICORN_ARCHS)" > config.log
 endif
```

[...]

PR #211, opened 2015-10-21
COMPILE.TXT: add note to configure Python for build

williballenthin:
Addresses #206


aquynh:
update: i see the PR now.

aquynh:
merged with a bit modification as above. thanks!

Issue #202, opened 2015-10-18
Go bindings mem_unmap


ranmrdrakono:
`mem_unmap` is missing in the go bindings. I tried to add it myself (https://github.com/ranmrdrakono/unicorn/commit/af2a6e7f27e6f78a3a7120b289fcddb9f62f6015 and https://github.com/ranmrdrakono/unicorn/commit/853734724fd80a286b0da44ba9df0d5e67d4d025). However my aproach seems to be of. calling unmap with invalid arguments works as expected (ARG_ERR returned). Calling unmap on an not-yet-mapped pages fails with UC_ERR_NOMEM as expected as well. However calling it on a previously maped page will not unmap the page.  Calling it a second time on the same page will result in a crash:
```
main: /golang/src/github.com/unicorn-engine/unicorn/qemu/memory.c:1596: memory_region_del_subregion_x86_64: Assertion `subregion->container == mr' failed.
```

Any hint would be appreciated

lunixbochs:
worked for me in #203

Maybe make sure you're mapping/unmapping page-sized data?

ranmrdrakono:
Yes I did. Can you try what happens if you add another call to unmap creating a double-unmap?

aquynh:
i merged the PR #203.

ranmrdrakono:
I have absolutely no idea why the exact same patches worked but this did solve the issue indeed.

ranmrdrakono:
Actually it didn't. This is still crashing:

```golang
package main

import (
	uc "github.com/unicorn-engine/unicorn/bindings/go/unicorn"
  "fmt"
  "testing"
)

func TestUnmap(t* testing.T) {
	mu, err := uc.NewUnicorn(uc.ARCH_X86, uc.MODE_64)

	if err != nil {
		fmt.Printf("%v\n",err)
	}

	// map and write code to memory
	if err := mu.MemMap(0x1000, 0x1000); err != nil {
		t.Error(fmt.Sprintf("%v\n",err));
	}

	if err := mu.MemMap(0x4000, 0x1000); err != nil {
		t.Error(fmt.Sprintf("%v\n",err));
	}

	if err := mu.MemUnmap(0x4000, 0x1000); err != nil {
		t.Error(fmt.Sprintf("%v\n",err));
	}

	if err := mu.MemUnmap(0x4000, 0x1000); err != nil {
		t.Error(fmt.Sprintf("%v\n",err));
	}

}

```

[...]

PR #213, opened 2015-10-21
fix for use after free in case of double unmap

ranmrdrakono:
Fix for the first issue discussed in: #202

### DIFF
--- a/tests/regress/Makefile
+++ b/tests/regress/Makefile
@@ -8,6 +8,7 @@ TESTS += ro_mem_test nr_mem_test
 TESTS += timeout_segfault
 TESTS += rep_movsb
 TESTS += mem_unmap
+TESTS += mem_double_unmap
 TESTS += mem_protect
 TESTS += mem_exec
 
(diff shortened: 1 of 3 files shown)


JonathonReinhart:
cc @aquynh 

Going forward, it would be really helpful if these tests were implemented under `tests/unit`, using CMocka.  Yes, it is a regression test, but it is a unit test too - the unit under test here is the memory mapping.

The key is that the tests will be run automatically and charted individually. Also the output is just much better, and the assertions happen immediately.

ranmrdrakono:
Ok, I felt it was not really testing the unit but a very specific behavior (e.g. double unmap after exactly two maps), which kind of only makes sense as a regression test. Should I move this test into a unit test and submit a new pull request?

JonathonReinhart:
There are many testing methodologies, but I believe your unit tests should strive to cover 100% of the code under test. This would include calling `mem_unmap` for a region that has been unmapped.  Honestly, I see no reason why the majority of our tests wouldn't all be under the same CMock-based framework, for consistency.  It keeps you from having to closely analyze the individual regression test output. Also, it cuts down on the amount of repeated code, which is always good.

Let's wait for @aquynh to weigh in here, but personally I would say yes, move this test to a testcase in`tests/unit/test_mem_map.c`

PR #214, opened 2015-10-22
create versioned library name and symlink unversioned name on *nix

### LINKED ISSUE #205: Test programs unable to find libunicorn.so.0
(Just wanted to put this in a proper issue, so we can track it.)

After the SONAME was added in 4860fdb, `make test` no longer works:

```
./test_sanity: error while loading shared libraries: libunicorn.so.0: cannot open shared object file: No such file or directory
```

Becuase libunicorn.so has an SONAME of libunicorn.so.0, the linker uses the SONAME for the DT_NEEDED entry of the tests. But because a library with that name does not exist, they would fail to run.

A solution would be to actually output `libunicorn.so.0` in the top-level directory when building, since that is what the test executables link against.

JonathonReinhart:
This fixes #205.

This change only affects *nix builds, and shouldn't affect `libunicorn.dll` on mingw, for example. However, it'd be nice if someone can build it to make sure. (It would be great if we had CI for this!)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,8 @@
 *.o
 *.a
 *.dSYM
+*.so
+*.so.*
 
 qemu/config-all-devices.mak
 
--- a/Makefile
+++ b/Makefile
@@ -266,7 +266,6 @@ ifeq ($(UNICORN_SHARED),yes)
 	$(INSTALL_LIB) $(LIBRARY) $(LIBDIR)
 ifneq ($(VERSION_EXT),)
 	cd $(LIBDIR) && \
-	mv lib$(LIBNAME).$(EXT) lib$(LIBNAME).$(VERSION_EXT) && \
 	ln -s lib$(LIBNAME).$(VERSION_EXT) lib$(LIBNAME).$(EXT)
 endif
 endif
--- a/Makefile
+++ b/Makefile
@@ -131,7 +131,8 @@ LIBRARY = $(BLDIR)/$(LIBNAME).$(EXT)
 else ifeq ($(IS_CYGWIN),1)
 LIBRARY = $(BLDIR)/$(LIBNAME).$(EXT)
 else	# *nix
-LIBRARY = $(BLDIR)/lib$(LIBNAME).$(EXT)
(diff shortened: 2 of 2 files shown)


aquynh:
have you tried to do "sudo make install" with this fix? see the "install" target of Makefile: it does not look like the proper libs with soft link will be installed into /usr/lib on Linux.

JonathonReinhart:
Fixed.

I noted that we should probably be running `ldconfig` to handle the installed symlink madness (especially when version numbers become more complex, and you need multiple symlinks). Or, we should be using `libtool` to handle this.

aquynh:
on OSX, i have this error for "sudo ./make.sh install"

    mkdir -p /usr/local/lib
    install -m0755 ./libunicorn.0.dylib /usr/local/lib
    cd /usr/local/lib && \
	ln -s libunicorn.0.dylib libunicorn.dylib
    ln: libunicorn.dylib: File exists
    make: *** [install] Error 1

so i changed "ln -s" to "ln -sf" to fix it.

merged, thanks.

aquynh:
it is OK on OSX, but in Ubuntu 14.04, i still cannot do "make test".

run "make; make test" to reproduce error like this:

    ....
    make -C tests/unit test
    make[1]: Entering directory `/home/quynh/projects/unicorn.git/tests/unit'
    gcc -Wall -Werror -Wno-unused-function -g -L ../../ -lcmocka -lunicorn -I ../../include -o test_sanity test_sanity.c
    /tmp/cctVUkGM.o: In function `test_uc_assert_macros_constants':
    /home/quynh/projects/unicorn.git/tests/unit/test_sanity.c:8: undefined reference to `uc_strerror'
    /home/quynh/projects/unicorn.git/tests/unit/test_sanity.c:8: undefined reference to `print_error'
    /home/quynh/projects/unicorn.git/tests/unit/test_sanity.c:8: undefined reference to `_fail'
    ....

JonathonReinhart:
I'm developing on Fedora 22, and I am not seeing that problem. From the output, your linker is not finding symbols from Unicorn or CMocka.  Do you have `libcmocka` installed? When this fails, is `libunicorn.so` in the project root directory?

PR #216, opened 2015-10-23
allow setting x86 segment base to host-sized value

lunixbochs:
this matches the segment management in qemu-user and is required to emulate the FS/GS register behavior


aquynh:
merged, thanks!

Issue #210, opened 2015-10-21
Changing PC from within an UC_HOOK_CODE doesn't work


hugoreinaldo:
Is there a way to change the program counter from within an UC_HOOK_CODE ? None of the following examples seem to work: 

    def hook_code(uc, address, size, userdata):
        uc.reg_write(arm_const.UC_ARM_REG_PC, 0x00)

    def hook_code(uc, address, size, userdata):
        uc.mem_write(address+size, 0x00)

lunixbochs:
Try stopping emulation and doing an emu_start() at the new address.

aquynh:
Can you send a pull request to put your testcase under tests/regress/, so it is possible to test and confirm this issue? Thanks

gaffe23:
There are test cases for this in `tests/regress/callback-pc.py` using both `UC_HOOK_CODE` and `UC_HOOK_BLOCK`.

hugoreinaldo:
@lunixbochs I'm trying your solution as a workaround for the issue.

@gaffe23 The test case tests/regress/callback-pc.py covers this issue, however there is no assertion validating the PC change in this test.

aquynh:
this issue was already discussed at https://github.com/unicorn-engine/unicorn/pull/4

gaffe23:
@hugoreinaldo Yep, I wrote that before we actually had a test suite. I'm about to submit a pull request to add assertions, just to make it clear that we'd like this to work at some point.

farmdve:
@aquynh 

Have you found a way to change EIP from hooked code? I've been studying Unicorn and the QEMU internals(thank god they are mostly stripped), it's difficult, there is lots of code, but I am making progress.

I do have some ideas, but they are incomplete.

aquynh:
yes, a rough idea is to quit the current basic block immediately, flush the JIT translation cache, then tell the engine to change the EIP, so next time JIT will start from the new place.

that is the approach, but i dont have time to work on the implementation yet. this is a bit tricky to code. if you can, just go ahead doing it.

aquynh:
you guys should send in a testcase about changing PC from the callback. i can fix that if i have time.

farmdve:
I am feeling in the mood to try and tackle this.

[...]

PR #218, opened 2015-10-23
add assertion to callback-PC test to make it clear what the correct behavior should be (issue #210)

gaffe23:


aquynh:
merged, thanks!

aquynh:
this testcase passes now.

```
$ ./callback-pc.py
>>> Tracing basic block at 0x10000000, block size = 0xa
>>> Setting PC to 0xffffffff
.>>> Tracing instruction at 0x10000000, instruction size = 2
>>> Setting PC to 0xffffffff
.
Ran 2 tests in 0.007s

OK
```

PR #219, opened 2015-10-24
limit uc_mem_(read|write) size to INT_MAX

### LINKED ISSUE #132: `size_t size` in uc_mem_read
uc_mem_read allow a `size_t size` to be passed to specify the amount to read. `size` is eventually passed to uc->read_mem on line 348 in the form of `len`, after ensuring that the memory region actually offers enough data to read on line 347. Up until that point, we're dealing entirely with 64 bit integers, assuming LP64, for example, a 64 bit Linux. However, uc->read_mem is of type
    bool (*uc_read_mem_t)(AddressSpace *as, hwaddr addr, uint8_t *buf, int len)
so `len` is cast to `int`, which is 32 bit on LP64. This will cause a truncation, and therefore less memory will be read than expected. It might be a good idea to limit `size` to be of type `int` and check that it is larger or equal to zero before doing anything, since we seem to be limited to reading INT_MAX anyway.
What do you think?

JonathonReinhart:
Fixes #132.

### DIFF
--- a/uc.c
+++ b/uc.c
@@ -341,6 +341,10 @@ uc_err uc_mem_read(uc_engine *uc, uint64_t address, void *_bytes, size_t size)
 {
     uint8_t *bytes = _bytes;
 
+    // qemu cpu_physical_memory_rw() size is an int
+    if (size > INT_MAX)
+        return UC_ERR_ARG;
+
     if (!check_mem_area(uc, address, size))
         return UC_ERR_READ_UNMAPPED;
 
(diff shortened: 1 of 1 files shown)


aquynh:
if we limit the read/write size to INT_MAX, should we also limit the memory size of `uc_mem_map()` & `uc_mem_unmap()` to the same number?

JonathonReinhart:
This was added simply because `cpu_physical_memory_rw` takes an `int`. I would have to review the underlying code used to implement `uc_mem_map`. But nothing is stopping one from calling read multiple times to overcome the 4GB limitation, if they have more memory mapped.

wtdcode:
Link to #1217 for further investigation.

wtdcode:
Fixed in df5daed. Thanks!

PR #222, opened 2015-10-26
Test for executing from  MIPS kseg0/kseg1 memory ranges

pkooiman:
As requested, pull request for MIPS kseg0/1 range test case


aquynh:
merged, but we still need to add this new testcase to `tests/regress/Makefile` to compile it.

thanks.

Issue #225, opened 2015-10-28
Incomplete python memory bindings


mrphrazer:
Hi,

it seems as would `mem_unmap` and `mem_map_prot`  be missing in the python bindings.

aquynh:
yes Python's `mem_unmap` & `mem_protect` are still missing. feel free to send a pull request to fix that, thanks.

PR #226, opened 2015-10-28
Python bindings mem_protect and mem_unmap

mrphrazer:
Related to https://github.com/unicorn-engine/unicorn/issues/225.

Hope this works in this way


aquynh:
merged, thanks!

PR #231, opened 2015-10-31
Go bindings: add Close() and set as GC finalizer (fix #230)

### LINKED ISSUE #230: Go binding `.Close()` method
Should the Go bindings expose a `.Close()` method on Unicorn instances? Seems that the Java and Python bindings do call `uc_close` when destroying instances, but I can't find any similar references in the Go bindings.

lunixbochs:


bindings/go/unicorn/unicorn_test.go
```diff
@@ -24,3 +24,16 @@ func TestMemUnmap(t *testing.T) {
...
+
+func TestDoubleClose(t *testing.T) {
+	mu, err := NewUnicorn(ARCH_X86, MODE_32)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if err := mu.Close(); err != nil {
+		t.Fatal(err)
+	}
+	if err := mu.Close(); err != nil {
```
aquynh: i dont know Go, but this does not look right: you verify with the same condition `err != nil` in both times when Close is called? the first time should success, but second time must fail, no?

lunixbochs: No, I'm testing to make sure calling Close() twice does NOT error, because I implicitly call it in a GC hook and a user might also call it.

bindings/go/unicorn/unicorn_test.go
```diff
@@ -24,3 +24,16 @@ func TestMemUnmap(t *testing.T) {
 		t.Fatal(fmt.Errorf("Expected ERR_WRITE_UNMAPPED, got: %v", err))
 	}
 }
+
+func TestDoubleClose(t *testing.T) {
+	mu, err := NewUnicorn(ARCH_X86, MODE_32)
```
aquynh: IMHO this function name `NewUnicorn` looks weird. is it better to be just `Unicorn`?

lunixbochs: No. Unicorn is the type. New<Type>() is the idiomatic Go convention.

aquynh:
merged, thanks!

PR #234, opened 2015-11-01
python README.TXT: clarify working directories

williballenthin:
clarify in which directory to build the core project and install the python bindings. when i went to install the python bindings, i was momentarily confused why the root Makefile didn't have an `install3` target.


aquynh:
merged, thanks!

PR #239, opened 2015-11-02
python bindings: use relative paths for imports to support py3

### LINKED ISSUE #235: python bindings should use package-relative imports
To support Python3 bindings, the Unicorn binding source files should use package-relative imports.

```
[willi@hostname python]$ python
Python 3.5.0 (default, Sep 20 2015, 11:28:25) 
[GCC 5.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import unicorn
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.5/site-packages/unicorn/__init__.py", line 3, in <module>
    from unicorn_const import *
ImportError: No module named 'unicorn_const'
```

Essentially the fix is:

```
-from unicorn_const import *
+from .unicorn_const import *
```

This should also work on Python2, so no change to existing Python2 users' behavior.

I'll submit a PR to fix this issue, probably today. Opened this issue not to request support, but coordinate/share ideas.

williballenthin:
closes #235.

works on test scripts on my system. please double check bindings continue to work for you under py2.

### DIFF
--- a/bindings/python/unicorn/__init__.py
+++ b/bindings/python/unicorn/__init__.py
@@ -1,4 +1,4 @@
 # Unicorn Python bindings, by Nguyen Anh Quynnh <aquynh@gmail.com>
 from . import arm_const, arm64_const, mips_const, sparc_const, m68k_const, x86_const
-from unicorn_const import *
-from unicorn import Uc, uc_version, uc_arch_supported, version_bind, debug, UcError
+from .unicorn_const import *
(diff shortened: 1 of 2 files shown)


aquynh:
merged, thanks!

cforger:
This is still an issue on the 64 bit Python bindings I just downloaded today from the main website. 

unicorn\__init__.py still needs this change:

from .unicorn_const import *

..and there seem to be more. Is the bin on the website stale?

Issue #232, opened 2015-11-01
Unexpected behavior calling `emu.Stop` from `HOOK_CODE`


williballenthin:
In a Go program, I'm encountering unexpected behavior when trying to implement "single-stepping" using `HOOK_CODE` callbacks that stop the emulator. An example program and the expected vs observed output is available [here](https://gist.github.com/williballenthin/e7b1ec10efd2de19c46a). I intend to port the sample to C to determine if the issue is limited to the Go bindings, and I'll update this issue when that is done. If this is *expected* behavior, please simply tell me and close the issue.

The [sample program](https://gist.github.com/williballenthin/e7b1ec10efd2de19c46a) has a `step` function that registers a `HOOK_CODE` callback that unconditionally stops the emulator. The function then runs the emulator, which should emulate a single instruction (at least thats my understanding, and suggested in #228). It then removes the hook.

When invoking the `step` function two times, exactly one instruction is *not* emulated per invocation. During the first `step`, the emulator does two instructions (callback hit twice, RIP changed by two instruction's worth). During the second `step`, its unclear to me what the emulator does. The callback is hit twice, with the addresses of two valid instructions, but RIP is not changed after the emulation.

I expected the callback to be hit once per `step` invocation, and that RIP would change by one instruction's worth. Is this a bug?

lunixbochs:
The Go bindings are a pretty thin wrapper around the C bindings. They shouldn't change any major behavior.

aquynh:
if you can rewrite your code in C or Python, i will look at that.

williballenthin:
I've ported the sample code to Python and updated the gist [here](https://gist.github.com/williballenthin/e7b1ec10efd2de19c46a). Same behavior.

aquynh:
yes this is a bug. can you send a PR to put this Python testcase under `tests/regress/`?

thanks

williballenthin:
Sure, I will do that. 

-----Original Message-----

aquynh:
here is something we should discuss: at address 0x1000, you decide to stop emulation. then after that, the PC is still 0x1000, right? it seems illogical to have PC at the next instruction, like what you are expecting.

(of course currently we have a bug preventing engine to stop immediately, but only do that at end of the basic block. i am testing the fix now)

aquynh:
note that the hook handler is actually executed **before** the hooked instruction is executed, so i think it makes sense to have the PC unchanged after emulation returns.

anybody has different ideas?

[...]

PR #240, opened 2015-11-02
add hook_code_stop_emu.py test for issue #232

williballenthin:
tests (and currently reports a failure) on py2 & py3

```
[willi@hostname regress]$ python3 hook_code_stop_emu.py 
pc: 0x1000
F
FAIL: test_hook_code_stop_emu (__main__.HookCodeStopEmuTest)
  File "hook_code_stop_emu.py", line 76, in test_hook_code_stop_emu
    "Emulator failed to stop after one instruction")
AssertionError: 4103 != 4105 : Emulator failed to stop after one instruction

Ran 1 test in 0.002s

FAILED (failures=1)
```

### DIFF
--- a/tests/regress/hook_code_stop_emu.py
+++ b/tests/regress/hook_code_stop_emu.py
@@ -29,7 +29,6 @@ class SingleStepper:
             self._test.assertEqual(1, self._hit_count, "HOOK_CODE invoked too many times")
             uc.emu_stop()
 
-
     def step(self):
         self._hit_count = 0
         h = self._emu.hook_add(UC_HOOK_CODE, self._stop_hook)
--- a/tests/regress/hook_code_stop_emu.py
+++ b/tests/regress/hook_code_stop_emu.py
@@ -67,15 +67,17 @@ class HookCodeStopEmuTest(regress.RegressTest):
             # 0x1016:  4883c201        add rdx, 1
             
             stepper = SingleStepper(mu, self)
-            self.assertEqual(0x1000, mu.reg_read(UC_X86_REG_RIP), "Unexpected PC")
             showpc(mu)
+            self.assertEqual(0x1000, mu.reg_read(UC_X86_REG_RIP), "Unexpected PC")
+
 
             stepper.step()
+            showpc(mu)            
             self.assertEqual(0x1007, mu.reg_read(UC_X86_REG_RIP),
                              "Emulator failed to stop after one instruction")
-            showpc(mu)
 
             stepper.step()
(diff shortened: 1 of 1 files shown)


aquynh:
merged, thanks!

tests/regress/hook_code_stop_emu.py
```diff
@@ -0,0 +1,89 @@
...
+        self._emu = emu
+        self._hit_count = 0
+        self._test = test
+
+    def _stop_hook(self, uc, address, *args, **kwargs):
+        if self._hit_count == 0:
+            self._hit_count += 1
+        else:
+            self._test.assertEqual(1, self._hit_count, "HOOK_CODE invoked too many times")
+            uc.emu_stop()
```
aquynh: this code is very confused: do you actually want to call uc.emu_stop() when _hit_count is 0 in the other path above, when you get hit the first time??

tests/regress/hook_code_stop_emu.py
```diff
@@ -0,0 +1,89 @@
...
+            # 0x1009:  48c7c700400000  mov rdi, 0x4000
+            # 0x1010:  488907          mov [rdi], rdx
+            # 0x1013:  488b07          mov rdx, [rdi]
+            # 0x1016:  4883c201        add rdx, 1
+            
+            stepper = SingleStepper(mu, self)
+            showpc(mu)
+            self.assertEqual(0x1000, mu.reg_read(UC_X86_REG_RIP), "Unexpected PC")
+
+
```
aquynh: you are supposed to set RIP to new address before single step (2 times in code below) because emu_stop() is supposed not to advance the RIP to the next instruction, as discussed.

tests/regress/hook_code_stop_emu.py
williballenthin: Sorry, I should have documented the code.

The intent of the `.step()` method is to emulate exactly one instruction. Because the `HOOK_CODE` callback is invoked before an instruction is emulated, we need to ignore the first invocation (line 27), and then stop the emulator before any subsequent instructions (line 30).

Looks like I missed a `self._hit_count += 1` after line 30, but that doesn't change the functionality of the method. Just the assertion on line 29 (`_hit_count` will never be greater than 1 right now).

tests/regress/hook_code_stop_emu.py
williballenthin: The intent of the `SingleStepper` is to move RIP forward by one instruction. That is a side effect of the `.step()` method.

PR #242, opened 2015-11-03
Fix segfault on emu_stop()

practicalswift:
Fix for the following segfault:

```python
unicorn.Uc(unicorn.UC_ARCH_X86, unicorn.UC_MODE_64).emu_stop()
...
python[23549]: segfault at ac ip 00007fc5395fe682 sp 00007fffb7093b90 error 6 in libunicorn.so.0[7fc5394cc000+915000]
```


aquynh:
merged, thanks!

PR #246, opened 2015-11-04
X86 EFLAGS not synced properly.

farmdve:
these commits should fix my previous ones.


aquynh:
merged, thanks!

farmdve:
The only thing I have not verified is whether this affects branching e.g JZ/JNZ or just the values of the EFLAGS.

Do you have any clue why the flags are not synced?

aquynh:
this should be fixed now with latest commit, please confirm.

PR #248, opened 2015-11-05
X86 zero-byte opcode causes Unicorn to crash.

farmdve:
The opcode 0x00 translates as `add byte ptr ds:[eax],al`, which leads to
a segfault in Unicorn.

Using a debugger, I believe the crash is located in
`qemu/translate-all.c` in function `page_flush_tb_1`, more specifically
this code `pd[i].first_tb = NULL;`

### DIFF
--- a/tests/regress/00opcode_uc_crash.c
+++ b/tests/regress/00opcode_uc_crash.c
@@ -1,0 +1,69 @@
+#include <stdlib.h>
+#include <stdio.h>
+#include <assert.h>
+
+#include <unicorn/unicorn.h>
+
+#define X86_CODE32 "\x00" // add byte ptr ds:[eax],al
+#define ADDRESS 0x1000000
+
+static void VM_exec()
+{
+    uc_engine *uc;
+    uc_err err;
+    uint32_t tmp;
+    unsigned int r_eax;
+
+    r_eax = 0x1000008;
+
+    // Initialize emulator in X86-32bit mode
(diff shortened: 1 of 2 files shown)


tests/regress/00opcode_uc_crash.c
```diff
@@ -0,0 +1,69 @@
+#include <stdlib.h>
+#include <stdio.h>
+#include <assert.h>
+
+#include <unicorn/unicorn.h>
+
+#define X86_CODE32 "\x00" // add byte ptr ds:[eax],al
```
aquynh: anyway, this should be "\x00\x00" (2 bytes), not one byte

aquynh:
merged, thanks!

farmdve:
Wow, sorry for that. I make sure to double-check everything before I submit, but I guess I didn't. With \x00\x00 there is no crash, however with just the one byte there is.

aquynh:
yes, can you find the real code causing this crash?

farmdve:
Looking at my code with \x00 vs \x00\x00, with the one byte that causes the crash QEMU is calling page_flush_tb_1 with level 0, whereas with \x00\x00 it doesn't.

Further testing shows that the address of the PageDesc structure is mangled. In one call of the page_flush_tb_1 function it is 0x02948b00 in another it's 0x0002948b right shifted by two bits.

Some kind of stack corruption?

PR #249, opened 2015-11-06
Fix crash on 00 opcode

farmdve:


farmdve:
Darn it, I still can't get around all this git nonsense.

PR #250, opened 2015-11-06
Fix crash on 00 opcode.

farmdve:
l1_map should be an array of pointers, thus void **


aquynh:
good catch, merged, thanks!

PR #254, opened 2015-11-09
Attempting to set some bits in the Unicorn EFLAGS doesn't work.

farmdve:
When attempting to set all flags except trap flag, the EFLAGS value
should be 0x00244ED7, I've tested this on Windows and here
https://ideone.com/WQAvk1 which is presumably Linux.

Unicorn however has the value 0x00000ED7, bits 11-21 are not set. Bit 21
is the ID bit indicating whether or not CPUID is available.

### DIFF
--- a/tests/regress/Makefile
+++ b/tests/regress/Makefile
@@ -14,6 +14,7 @@ TESTS += mem_exec
 TESTS += mips_kseg0_1
 TESTS += eflags_nosync
 TESTS += 00opcode_uc_crash
+TESTS += eflags_noset
 
 all: $(TESTS)
 
--- a/tests/regress/eflags_noset.c
+++ b/tests/regress/eflags_noset.c
@@ -1,0 +1,106 @@
+#include <stdlib.h>
+#include <stdio.h>
+#include <stdint.h>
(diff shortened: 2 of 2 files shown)


aquynh:
merged, thanks!

aquynh:
compile this on Ubuntu with GCC 4.8.4 resulted in error:

    cc -I../include eflags_noset.c ../../libunicorn.a -lglib-2.0   -lpthread -lm -o eflags_noset
    eflags_noset.c: Assembler messages:
    eflags_noset.c:23: Error: operand type mismatch for `push'
    eflags_noset.c:26: Error: operand type mismatch for `pop'
    make: *** [eflags_noset] Error 1

farmdve:
Ah...you need to compile as 32-bit I think.

2015-11-10 5:23 GMT+02:00 Nguyen Anh Quynh <notifications@github.com>:

aquynh:
can you fix this to compile on x64?

farmdve:
Working on it, sorry for the trouble :/

2015-11-10 12:18 GMT+02:00 Nguyen Anh Quynh <notifications@github.com>:

Issue #209, opened 2015-10-21
Failed to compile on a Raspberry Pi 2


DeveloppSoft:
Hello guys,
So, I have discovered your project because someone suggested it to me on freenode.net.
I downloaded the code and do what COMPILE.txt explain, but I fail to compile using ./make.sh with the folowing error :
```shell
root@kali:~/unicorn-0.9# ./make.sh
if [ "x86 m68k arm aarch64 mips sparc" != "`cat config.log`" ]; then make clean; fi
rm -rf libunicorn* unicorn*.lib unicorn*.dll && cd qemu && make -j 8
make[1]: Entering directory '/root/unicorn-0.9/qemu'
  CC    m68k-softmmu/tcg/tcg.o
  CC    arm-softmmu/tcg/tcg.o
  CC    mipsel-softmmu/tcg/tcg.o
  CC    x86_64-softmmu/tcg/tcg.o
  CC    mips-softmmu/tcg/tcg.o
  CC    mips64el-softmmu/tcg/tcg.o
  CC    aarch64-softmmu/tcg/tcg.o
  CC    mips64-softmmu/tcg/tcg.o
In file included from /root/unicorn-0.9/qemu/tcg/tcg-op.h:24:0,
                 from /root/unicorn-0.9/qemu/tcg/tcg.c:49:
/root/unicorn-0.9/qemu/tcg/arm/tcg-target.c: In function ‘tcg_target_init_m68k’:
/root/unicorn-0.9/qemu/tcg/arm/tcg-target.c:1997:22: error: ‘tcg_target_available_regs’ undeclared (first use in this function)
     tcg_regset_set32(tcg_target_available_regs[TCG_TYPE_I32], 0, 0xffff);
                      ^
/root/unicorn-0.9/qemu/tcg/tcg.h:140:42: note: in definition of macro ‘tcg_regset_set32’
 #define tcg_regset_set32(d, reg, val32) (d) |= (val32) << (reg)
                                          ^
/root/unicorn-0.9/qemu/tcg/arm/tcg-target.c:1997:22: note: each undeclared identifier is reported only once for each function it appears in
     tcg_regset_set32(tcg_target_available_regs[TCG_TYPE_I32], 0, 0xffff);
                      ^
/root/unicorn-0.9/qemu/tcg/tcg.h:140:42: note: in definition of macro ‘tcg_regset_set32’
 #define tcg_regset_set32(d, reg, val32) (d) |= (val32) << (reg)
                                          ^
[...]

PR #259, opened 2015-11-12
improve multiarch support

lunixbochs:
related to #209

doesn't entirely fix non-x86, but this is the first step

### DIFF
--- a/qemu/tcg/aarch64/tcg-target.c
+++ b/qemu/tcg/aarch64/tcg-target.c
@@ -1718,10 +1718,10 @@ static const TCGTargetOpDef aarch64_op_defs[] = {
 
 static void tcg_target_init(TCGContext *s)
 {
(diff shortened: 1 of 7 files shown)


aquynh:
great, please keep working on this issue!

lunixbochs:
Quick question, some tcg ops rely on variables filled by platform detection done during tcg_target_init()
qemu/tcg/arm/tcg-target.c

Like `DEF(div_i32, 1, 2, 0, IMPL(TCG_TARGET_HAS_div_i32))` where `TCG_TARGET_HAS_div_i32` on ARM is provided by the extern symbol `use_idiv_instructions` from `tcg.o`, but `tcg.o` is getting linked multiple times, so there's a duplicate symbol warning.

I can't move this platform detection to the context, as the condition variables being accessed need to go into a top-level array (which is filled pre-init). Any thoughts on where to put these variables?

A harder example is tcg_target_deposit_valid as it's a forward declared function.

Issue #223, opened 2015-10-27
absurdly large mem_map() aborts unicorn


lunixbochs:
from unicorn import *
    u = Uc(UC_ARCH_X86, UC_MODE_32)
    u.mem_map(0x60802000, 0x28bd211200004000)

aquynh:
can you put this script under `tests/regress/`?

lunixbochs:
I haven't been paying attention but it looks like regression tests are now some kind of managed test suite?

This isn't a very nice unit test and will break the regression suite from running (because it aborts Unicorn). Do you still want it in the same format?

aquynh:
On Sat, Oct 31, 2015 at 1:42 PM, Ryan Hileman <notifications@github.com>
wrote:

> I haven't been paying attention but it looks like regression tests are now
> some kind of managed test suite?
>
> This isn't a very nice unit test and will break the regression suite from
> running (because it aborts Unicorn). Do you still want it in the same
> format?
>

do you mean one test fails will abort all the following tests?

lunixbochs:
I didn't see how the regressions run, but yes it will crash the enclosing Python process.

aquynh:
can you send a PR to put your script there? i will look at that.

JonathonReinhart:
CMOCKA is supposed to handle that. It will catch e.g. SIGSEGV, abort the test and go on.

lunixbochs:
Added a C regression test for this.

aquynh:
fixed, thanks!

PR #260, opened 2015-11-12
add mem_map_large regression (for #223)

lunixbochs:


aquynh:
merged, thanks!

aquynh:
did you check with latest code? this testcase gives this output:

    $ ./mem_map_large
    Trying large map.
    uc_mem_map() failed: No memory available or memory not present (UC_ERR_NOMEM)

lunixbochs:
Yep, looks good to me now.

PR #262, opened 2015-11-13
Add test for emulation within and across BBs

williballenthin:
Expected behavior: the test passes.

Observed behavior: test fails:
```
[willi@hostname regress]$ python2 run_across_bb.py 
pc: 0x1000
>>> Tracing instruction at 0x1000, instruction size = 5
>>> Tracing instruction at 0x1005, instruction size = 1
>>> Tracing instruction at 0x1006, instruction size = 1
>>> Tracing instruction at 0x1007, instruction size = 5
pc: 0x100c
>>> Tracing instruction at 0x100c, instruction size = 1
!!! about to bail due to bad fetch... here's the data at PC:
b800000000404000
F
FAIL: test_run_across_bb (__main__.RunAcrossBBTest)
  File "run_across_bb.py", line 133, in test_run_across_bb
    self.assertFalse(True, "ERROR: %s @ 0x%x" % (e, mu.reg_read(UC_X86_REG_EIP)))
AssertionError: ERROR: Invalid memory fetch (UC_ERR_FETCH_UNMAPPED) @ 0x1010

Ran 1 test in 0.002s

FAILED (failures=1)
```

Unless my understanding of the assembly is incorrect (possible, its a bit late here, though I've re-read a few times), it seems that the emulator bails on a bad fetch, which it should not. The final `print` statement shows that instructions exist at PC. My original intent for the test was to show possibly weird behavior on emulating within or across basic blocks (`emu_start` from the middle of one BB to the middle of another BB).

See incoming issue.

### DIFF
--- a/tests/regress/run_across_bb.py
+++ b/tests/regress/run_across_bb.py
@@ -1,0 +1,137 @@
+#!/usr/bin/python
+#
+# This test demonstrates emulation behavior within and across
+#  basic blocks.
+
+from __future__ import print_function
+import binascii
+import regress
+import struct
+
+from unicorn import *
+from unicorn.x86_const import *
+
+
+CODE = binascii.unhexlify(b"".join([
+  b"b800000000", # 1000: b8 00 00 00 00          mov    eax,0x0
+  b"40",         # 1005: 40                      inc    eax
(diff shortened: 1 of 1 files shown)


aquynh:
merged, thanks!

Issue #269, opened 2015-11-16
Invalid write of size 4 in cpu_tb_exec_x86_64


practicalswift:
Offending code:

```#define BINARY "\xc4\xe2\xdb\xf6\xe2"```

Segfault:

```segfault at 480 ip 00007ffadbff5a8d sp 00007ffad88498c0 error 6 in libunicorn.so.0[7ffadbff5000+1000]```

valgrind:

```
==4198== Invalid write of size 4
==4198==    at 0x59B2A8B: ??? (in /usr/lib/libunicorn.so.0)
==4198==    by 0x1D1E18FF: ???
==4198==    by 0xDFFFFFFFF: ???
==4198==    by 0xE8FFFFFFFF: ???
==4198==    by 0x1A7F30EF: ???
==4198==    by 0x59B2AAA: ??? (in /usr/lib/libunicorn.so.0)
==4198==    by 0x7F2174681A3FEAFF: ???
==4198==    by 0x1D1E192F: ???
==4198==    by 0x4EB7D73: tcg_out_opc (tcg-target.c:450)
==4198==    by 0x1D1E192F: ???
==4198==  Address 0x480 is not stack'd, malloc'd or (recently) free'd
```

**Comment by farmdve** [comment]
I looked into this issue as well.

It's again a problem with the jitted code. 

mov rax,qword ptr ds:[r14+10] // moves the value of RDX to rax
mul rax // multiplies
mov dword ptr ds:[rax+280],eax // uh oh, it tries to store to rax+280, thats definitely not right.
mov dword ptr ds:[rsi+80],test64.1000005 // update pc counter I believe.

**Comment by aquynh** [comment]
then you should compare the output of "mulx rsp,rsp,rdx" and "mulx rsp,rax,rdx" to see the difference, then to find out the issue.

**Comment by aquynh** [comment]
How did you test this code with Qemu? Can you post your test here?

**Comment by farmdve** [comment]
I did not. Somebody else did and said it crashed for him on original qemu. I reported this to qemu, they wrote this test case https://gist.github.com/XanClic/b8d87473e5f324312a94 and it doesn't crash for them, however they have found that the order of updating is not correct. So while it doesn't crash for them, there is a bug still.

So I got even further on the segfault. In tcg_reg_alloc_op there is the following code.
```c
        if (NEED_SYNC_ARG(i)) {
            tcg_reg_sync(s, reg);
        }
```

It updates the registers, but remember, they are wrong. 

[...]

PR #270, opened 2015-11-16
Crash case: Invalid write of size 4 in cpu_tb_exec_x86_64 (issue #269)

practicalswift:
Test case for issue #269


aquynh:
merged, thanks!

PR #272, opened 2015-11-17
Fix issue #269

farmdve:
Patch from here
http://lists.nongnu.org/archive/html/qemu-devel/2015-11/msg03848.html

Also fix another potential issue with constants from
https://github.com/qemu/qemu/commit/bbeb82395eaca0e3c38b433b2d2a5bad4a5fbd81#diff-9e0011b4d4a5890b309421630e6d86c3


farmdve:
Please test before merging, thanks.

aquynh:
it is awesome to see that we are starting to contribute to QEMU :-)

merged, thanks!

Issue #261, opened 2015-11-12
file mmap() support


lunixbochs:
I'd like to be able to directly mmap() files into target memory. I think this is facilitated in qemu-user by doing a "guest-to-host" address translation and just calling mmap() directly. Do host memory addresses backing guest memory ever change position? Something to guarantee a guest address range is backed by a fixed contiguous host memory region would be sufficient for this.

Maybe an API like `guest_fixed_map(host_addr, guest_addr, size)` would work, which points a region of guest memory at an existing region of host memory. This way it's not tied to anything overly platform specific like mmap() APIs.

JonathonReinhart:
I think this is huge. The vast majority of data copied into guest memory is probably coming from a file anyway.

aquynh:
yes this will be nice to have. Qemu has something called memory-backend-file, so you might want to look at that to see if we can use it for this API.

lunixbochs:
qemu-user seems to think calling mmap directly is fine: https://github.com/hackndev/qemu/blob/master/darwin-user/mmap.c#L225

I just think a portable version of file mapping doesn't really need to be Unicorn's job. All I really want is the ability to make a fixed mapping from guest to host memory so I can do whatever I want with it.

aquynh:
yes, feel free to propose new API and implement it.

lunixbochs:
Proposed API usage (as implemented in the pull request):

    uc_mem_map_ptr(handle, addr, size, void *ptr);

Permissions are not specified because Unicorn does not control them.

aquynh:
this is already solved with the new API `uc_mem_map_ptr()`, so closed. thanks

PR #277, opened 2015-11-28
implement host-controlled memory mapping for #261

lunixbochs:
This PR adds an exported `uc_mem_map_ptr(uc, addr, size, ptr)` function, which allows you to map guest memory regions backed by a host-controlled pointer. This allows the user to fully control the backing memory, including any platform-specific functions like memory mapping files.

There's a test at [tests/unit/test_mem_map_ptr.c](https://github.com/lunixbochs/unicorn/blob/d497a917ef9edf8e9095d958e65afe3a236e3a1a/tests/unit/test_mem_map_ptr.c)


uc.c
```diff
@@ -570,11 +570,8 @@ static uc_err _hook_mem_access(uc_engine *uc, uc_hook_type type,
     return UC_ERR_OK;
 }
 
-UNICORN_EXPORT
-uc_err uc_mem_map(uc_engine *uc, uint64_t address, size_t size, uint32_t perms)
-{
-    MemoryRegion **regions;
-
+// common setup/error checking shared between uc_mem_map and uc_mem_map_ptr
+static uc_err mem_map_start(uc_engine *uc, uint64_t address, size_t size, uint32_t perms) {
```
aquynh: code style: please put the open bracket in a new line

uc.c
```diff
@@ -591,6 +588,7 @@ uc_err uc_mem_map(uc_engine *uc, uint64_t address, size_t size, uint32_t perms)
     if ((perms & ~UC_PROT_ALL) != 0)
         return UC_ERR_ARG;
 
+    MemoryRegion **regions;
```
aquynh: please declare variables at the begin of the function

uc.c
```diff
@@ -600,7 +598,12 @@ uc_err uc_mem_map(uc_engine *uc, uint64_t address, size_t size, uint32_t perms)
         uc->mapped_blocks = regions;
     }
 
-    uc->mapped_blocks[uc->mapped_block_count] = uc->memory_map(uc, address, size, perms);
+    return UC_ERR_OK;
+}
+
+// common final step shared by uc_mem_map and uc_mem_map_ptr
+static uc_err mem_map_finish(uc_engine *uc, MemoryRegion *block) {
```
aquynh: please fix the open bracket here, too.

lunixbochs:
New commits pushed addressing code style comments.

aquynh:
i merged, and pushed to this to a new branch "map-ptr". i also added a sample code in `samples/sample_x86.c`, but it segfaults now on Ubuntu 14.04 64bit, as follows.

```
$ ./samples/sample_x86 -32
Emulate i386 code
>>> Tracing basic block at 0x1000000, block size = 0x2
>>> Tracing instruction at 0x1000000, instruction size = 0x1
>>> --- EFLAGS is 0x0
>>> Tracing instruction at 0x1000001, instruction size = 0x1
>>> --- EFLAGS is 0x4
>>> ECX = 0x1235
>>> EDX = 0x788f
>>> Read 4 bytes from [0x1000000] = 0x4a41
===================================
Emulate i386 code - use uc_mem_map_ptr()
```

lunixbochs:
at /home/vagrant/build/unicorn/qemu/translate-all.c:787
    783 static void page_flush_tb_1(int level, void **lp)
    ...
    787     if (*lp == NULL) {

aquynh:
fixed with the latest commit.

lunixbochs:
Thinking about it a bit, QEMU reinforces the memory permissions internally and expects the target to be mprotect READ+WRITE? In that case, we actually need to expose QEMU's memory permissions during map.

aquynh:
there are some major issues with this API:

- the permission of memory we give to uc_mem_map_ptr() is not controlled by Unicorn. what if this memory is allocated by wrong permission? (we assign permission as UC_PROT_ALL by default now). or what happen if the permission is changed at run-time?

- the memory is allocated outside of Unicorn, thus can be manipulated without its awareness. example: if we change the code inside the memory, we need to tell Unicorn about that. this is not considered at the moment.

lunixbochs:
Yeah, I think we should expose Unicorn's permissions and document that you'll likely run into problems with Unicorn's API if you pass in memory that doesn't have read or write perms.

For (2), I don't know of any real-world cases where you jump into a shared memory segment that's changing... that sounds terrible to do properly even without an emulator in the middle. The big case of executing code from memory mapped files is loading some number of unchanging binaries during the dynamic linking stage. I'm more interested in mmap for data.

lunixbochs:
New commit adds permissions argument.

aquynh:
yes, but i think a good API would do should prevent as much as possible mistakes from users.

can you please rebase your latest commit against the "map-ptr" branch, and send a new PR?

PR #278, opened 2015-11-28
add permissions to map_ptr api

lunixbochs:
continuation of #277


aquynh:
merged, thanks!

how can Python use this new API?

Issue #284, opened 2015-12-01
[x86_64] Invalid read of size 8 in tb_flush_x86_64


practicalswift:
Crash case: https://github.com/unicorn-engine/unicorn/pull/285

Unicorn segfaults on `uc_emu_start(…)` after doing a `uc_mem_map(…, …, …, UC_PROT_READ)`.

valgrind:

```
==26434== Invalid read of size 8
==26434==    at 0x4F97E40: tb_flush_x86_64 (translate-all.c:811)
==26434==    by 0x4ED02A2: cpu_x86_exec (cpu-exec.c:277)
==26434==    by 0x4FE1503: qemu_tcg_cpu_thread_fn_x86_64 (cpus.c:214)
==26434==    by 0x1AB98181: start_thread (pthread_create.c:312)
==26434==    by 0x1A68747C: clone (clone.S:111)
```

farmdve:
Going to hunt this first thing tomorrow.

aquynh:
this is fixed, please confirm.

practicalswift:
Confirmed! Thanks for fixing! :-)

PR #285, opened 2015-12-01
Crash case: Invalid read of size 8 in tb_flush_x86_64 (#284)

practicalswift:


aquynh:
merged, thanks!

can you put this into Makefile to compile it, and update .gitignore as well?

PR #294, opened 2015-12-05
Added MSVC++ support for unicorn

xorstream:
This lets you import the pre-built unicorn.dll files with Microsoft
Visual C++ projects.

There is support for static and dynamic linking of dlls. This has been
tested as working for both 32bit and 64bit versions.

The dynamic linking code should also work in Linux, though I have not
tested it.


aquynh:
merged, thanks!

aquynh:
should we move `include/unicorn/unicorn_dynload.h` to `bindings/msvc`, since this is only for MSVC binding?

xorstream:
Yeh if you like. I initially had unicorn_dynload,c in the root dir with uc.c too, but then moved it to the bindings dir. I wasn't sure where they best fitted.

I was thinking about the prebuilt binaries for windows. It would be best with the prebuilt packs to have the unicorn_dynload.h file in the main "include/unicorn" directory so that it is ready to use with the least fuss by windows users. It would probably be best to have the prebuilt unicorn_staload.lib files in the root of the binary package and so maybe also have a prebuilt version of unicorn_dynload.lib in the root too.

I held off making a lib version of unicorn_dynload.c since it also had linux support in it, but perhaps it would be best especially for the prebuilt binaries since it might reduce confusion to the end users.

xorstream:
Once this is sorted I can modify the existing sample sources to support MSVC++ if you like.
This basically just involves adding the following lines to the top of each sample source:

#ifdef _MSC_VER
#include <io.h>
#define PRIx64	"I64X"
#else
#include <unistd.h>
#include <inttypes.h>
#endif

#ifdef DYNLOAD
#include <unicorn/unicorn_dynload.h>
#else
#include <unicorn/unicorn.h>
#endif

I could also create MSVC++ solution/project files to easily build samples in MSVC++

aquynh:
when distributing the Windows binaries, we can have a copy of `bindings/msvc/` inside, so it is fine to move `unicorn_dynload.h` there.

yes, please go ahead adding support for MSVC to the samples.

robhshu:
:+1: I'm also keen to get some MSVC bindings in Unicorn

xorstream:
The MSVC bindings should now be available in the latest code on github.

PR #295, opened 2015-12-05
Added MIPS delay slot code hook test

xorstream:
Tests that the code hook gets called for instructions in the branch
delay slot for MIPS cpu.


aquynh:
can you add this to tests/regress/Makefile to compile it?

aquynh:
this PR has conflicts, perhaps due to the reverted commit. would you mind sending this PR again - based on the latest code, and without the revert?

thanks

xorstream:
will do

PR #297, opened 2015-12-08
Added MSVC sample projects

xorstream:
Added sample projects for MSVC binding.
Supports 32bit and 64bit projects using static or dynamic linking.

Added MSVC support for existing unicorn samples.
Supports 32bit and 64bit projects using static or dynamic linking.


robhshu:
Thanks for this contribution. I'm wondering if it's worth renaming the solution files to match the Visual Studio edition? Thoughts?

xorstream:
Yeh that might be a good idea, then people could provide solution files from other versions of Visual Studio (and their own project files too if required).

What do you think about renaming:
samples solution file from msvc.sln to msvc_vs2012.sln (or just vs2012.sln?)
bindings solution file from samples.sln to samples_vs2012.sln (or just vs2012.sln?)

Should i rename the project files too?
like from staload.vcxproj to staload_vs2012.vcxproj
etc.

Thoughts ppl?

aquynh:
i think there are two solutions:

- have a single .sln for VS2010. all the newer versions of VS can handle this (so we do not support older VS)

- use CMake to compile (so we need to create `CMakeLists.txt`). since CMake can generate SLN for all versions of VS, we can solve this problem.

thanks.

aquynh:
@xorstream, which VS version do you support with this PR?

xorstream:
This supports VS 2012.
From the readme:

These solution and project files were created using Visual Studio 2012.
They should be able to be opened in newer versions of Visual Studio.
For older versions of Visual Studio you could try a little hack of changing
the line in the solution file "samples.sln" from:
	Microsoft Visual Studio Solution File, Format Version 12.00
to
	Microsoft Visual Studio Solution File, Format Version 11.00

Or whatever version number your Visual Studio version uses.
(Hint: Check an existing solution file you have created with your version
 of Visual Studio to know what this value is expected to be.)

aquynh:
i think VS2010 is still common now, so we should handle that. 
the easiest solution now is to change the supported version to VS2010. can you do that?

xorstream:
I only have VS2012 sorry.
In the past I have altered the solution file in a text editor like I mention in the readme, but I think it would be better for someone who has VS2010 to make the changes and test.

xorstream:
http://stackoverflow.com/questions/13042356/how-to-convert-vs2012-sln-project-files-to-vs2010
This page has some info on how you might convert a 2012 project to 2010. However it really should be tested by someone who has vs2010 before doing so.

Perhaps the 2012 versions could be put in the repo for now until someone is found to try the 2010 conversion test, then they could submit the 2010 conversion changes if it works.

aquynh:
merged with a simple fix on code style, thanks!

so you dont move `unicorn_dynload.h` to `bindings/msvc` yet?

xorstream:
I thought you were going to move unicorn_dynload.h.
I arranged all the MSVC samples projects to have include paths for the binding dir and the normal includes dir. So moving unicorn_dynload.h to the bindings dir will not require any changes to the MSVC samples.

robhshu:
Related; `unicorn_dynload.c` is missing hooks from the `uc_hook_type` enum (any `_UNMAPPED` types).

PR #302, opened 2015-12-09
Added MIPS delay slot code hook test.

xorstream:
Tests that the code hook gets called for instructions in the branch delay slot for MIPS cpu.


aquynh:
merged, thanks!

PR #304, opened 2015-12-09
Add ARM mode for instruction byteswapping.

jantore:
This commit adds a new mode (UC_MODE_BYTESWAP) that can be used with UC_ARCH_ARM to enable QEMU's instruction byteswapping. This is useful when trying to emulate big-endian code that is interpreted by QEMU as little-endian.

The changes aren't major, but I haven't really spent much time looking at the Unicorn or QEMU codebase. I'm not entirely sure about UC_MODE_BYTESWAP and UC_MODE_BIG_ENDIAN sharing the same value, for instance, but it makes sense in my head.


radare:
Imho byteswap is conceptually wrong because it doesnt makes any reference to the host or target endianness

lunixbochs:
is ARM byte-swapping any different from an armeb big-endian mode?

jantore:
Yes, byteswapping only affects how QEMU loads instructions. The endianness on memory loads/stores is not changed. As such this makes it practical to emulate mixed-endian images. Admittedly a corner case, but it made my life a little easier. :-)

aquynh:
agreed with @radare that this does not look right.

aquynh:
we now support ARM big-endian mode, please check the latest Github code.

PR #305, opened 2015-12-09
Unmapped memory is not freed.

farmdve:
While uc_mem_unmap does unmap memory regions from Unicorn, it does not
free the memory. It accumulates over time when reusing a single Unicorn instance.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -118,6 +118,7 @@ x86_16_segfault
 mips_invalid_read_of_size_4_when_tracing
 invalid_read_in_tb_flush_x86_64
 sparc_jump_to_zero
+mem_nofree
 
 
 #################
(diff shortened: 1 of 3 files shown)


aquynh:
merged, thanks!

farmdve:
So apparently the problem is that in 

```c
void object_unref(struct uc_struct *uc, Object *obj)
{
    if (!obj) {
        return;
    }
    g_assert(obj->ref > 0);

    /* parent always holds a reference to its children */
    if (atomic_fetch_dec(&obj->ref) == 1) {
        object_finalize(uc, obj);
    }
}
```

`obj->ref` contains a total of around 8 references, thus `object_finalize` is never called which in turns does not call `object_deinit` where the memory free functions are called.

lunixbochs:
Can you make sure your code will work with the map-ptr branch, where a memory region can be allocated outside QEMU?

After quick look it should be fine as long as you're calling the memory region's own destructor, as map_ptr has a special destructor: https://github.com/unicorn-engine/unicorn/blob/6d21ebabea3f4ad2cae3da63249af24155215738/qemu/memory.c#L1223

farmdve:
@lunixbochs 

Your observation is correct. If we call the memory region's own destructor it should be fine, shouldn't it?

lunixbochs:
Sounds correct to me, thanks.

PR #306, opened 2015-12-10
Moved unicorn_dynload.h into bindings dir.

xorstream:
Moved unicorn_dynload.h into bindings dir.

All MSVC sample files should still work without requiring any changes.


aquynh:
merged, thanks!

PR #307, opened 2015-12-10
Added mnemonic printing to disasm function.

xorstream:
The disasm function was not printing the instruction mnemonic it was only printing instruction operands. This was changed to now print the mnemonic as well.


aquynh:
merged, thanks!

PR #308, opened 2015-12-10
Added testcase for int instruction tracing

ranmrdrakono:


JonathonReinhart:
Thanks for the unit test!

I think this might make more sense as part of `test_x86`. When I started setting up `test/unit`, I imagined there being tests for high-level unicorn features like memory mapping (hence 'test_mem_map'), and then per-architecture test programs (as there may be shared helper functions in them that apply to an architecture).

aquynh:
merged, thanks!

PR #310, opened 2015-12-11
Fix uc_mem_unmap memory leak and in uc_close.

farmdve:
It appears the problem is that we are not calling the memory region
destructor. After modifying memory_unmap to include the destructor call
for the memory region, the memory is freed.

Furthermore in uc_close we must explicitly free any blocks that were not
unmapped by the user to prevent leaks.

This should fix issue 305.


aquynh:
merged with a little fix: g_free() can handle NULL pointer, so you dont need to check the pointer value.

can you look at the remaining memory leak issue?

thanks.

PR #311, opened 2015-12-11
Fix segfault introduced in my previous commits.

farmdve:
Although I am not sure why mr->enabled needs to be set to false when we are uninitializing everything.


aquynh:
merged, thanks!

PR #312, opened 2015-12-11
Test for stack hooking.

univm:
If address < 0x0030000 the hook isn't called.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -122,6 +122,7 @@ mem_nofree
 mips_delay_slot_code_hook
 
 test_mem_high
+rw_hookstack
 
 
 #################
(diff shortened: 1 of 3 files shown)


aquynh:
merged, thanks!

univm:
Ah, it appears the problem is if you don't specify begin > end the values are not initialized and seems to make it not work.

I believe if these values are not specified, they should be initialized to default begin > end.

aquynh:
you can send a pull request to do this.

aquynh:
it is better if we can enforce these parameters so they must be always present.

lunixbochs:
You can't enforce missing arguments in C.

Issue #290, opened 2015-12-03
[mips32] code hook not executed for delay slot instructions


xorstream:
When using a code hook with mips it is not calling the hook for instructions in a delay slot.

The instructions in a delay slot do get executed by the emulator, as I can see their effect on values in registers. Just the hook doesn't get called for them.

My test instructions are:

	unsigned char loop_test_code[] = {
		0xFF,0x00,0x04,0x24,	// li      $a0, 0xF
		// loop1
		0xC0,0x10,0x04,0x00,	// sll     $v0, $a0, 3
		0x00,0x00,0x44,0xAC,	// sw      $a0, 0($v0)
		0xFD,0xFF,0x80,0x14,	// bnez    $a0, loop1
		0xFF,0xFF,0x84,0x24,	// addiu   $a0, -1
	};

I have previously set an instruction hook by doing:

	// hook all instructions by having @begin > @end
	uc_hook hhc;
	uc_hook_add(uc, &hhc, UC_HOOK_CODE, mips_codehook, NULL, (uint64_t)1, (uint64_t)0);

The hook gets called for every instruction except "addiu   $a0, -1"

aquynh:
could you please send in a pull request on this issue to put a test under either tests/regress/ or tests/unit/? it makes it easier to confirm the problem & test the fix this way.

thanks.

lunixbochs:
Huh, in #133 it was calling the hook twice.

aquynh:
@xorstream, do you still see your problem with the latest Unicorn (from Github)?

xorstream:
I was using the v0.9 stable build not the latest build from git. So perhaps things have changed. I am reading up on how to use git/github atm. I will download the latest version later today and test it. If the issue still exists i'll create test code for it and submit a pull request.

aquynh:
yes, the latest code on Github fixes quite a lot of issues since v0.9.

just in case, you can always get the latest code directly from this link: https://github.com/unicorn-engine/unicorn/archive/master.zip

xorstream:
Ok I can confirm that the issue still exists with latest code from github. I'll make up a test for it.

Btw is it standard to build the source inside your local repo clone?
When I did so it detected a change to "qemu\config-all-disas.mak"
It looks like perhaps that needs to be added to .gitignore after the "qemu/config-all-devices.mak" entry.

aquynh:
it is recommended to build & install Unicorn like this (on Mac, Linux)

    $ ./make.sh
[...]

Issue #133, opened 2015-09-08
how does UC_HOOK_CODE interact with branch delay slots?


lunixbochs:
I saw this trace output on MIPS:

```
0x4007e8: 2d009510 beq $a0, $s5, 0x4008a0
0x4007ec: 6a008228 slti $v0, $a0, 0x6a
0x4007ec: 6a008228 slti $v0, $a0, 0x6a
```

This is weird because I got the same instruction address twice in a row (which was confusing the deadlock detection I added to detect bugs in Unicorn). Based on the addresses here, `slti` is in a branch delay slot after the `beq` instruction.

xorstream:
This test case actually jumps to the slti instruction at 400004.
So it executes the slti @ 400004 once due to being in the delay slot, then it executes it a second time during normal execution since it is the target of the branch. So for this test case I would expect to see the slti executed two times.

0x400000: 0000a412 beq    $s5, $a0, 0x400004
0x400004: 6a008228 slti   $v0, $a0, 0x6a
0x400008: 00000000 nop

If instead the test case is changed to the following, then it will only execute the slti once.

0x400000: 0100a412 beq    $s5, $a0, 0x400008
0x400004: 6a008228 slti   $v0, $a0, 0x6a
0x400008: 00000000 nop


So the "fix" that @aquynh did for this should probably be reverted in order to fix issue #290.
HOWEVER reverting that fix leaves the code with a separate issue where the size of the delay slot instruction is 0xf8f8f8f8 instead of 4 (or whatever the real instruction size is, maybe 2 for mips16). Perhaps this is related to the code at "qemu/target-mips/translate.c:19229":
            // save block address to see if we need to patch block size later
            env->uc->block_addr = pc_start;
            gen_uc_tracecode(tcg_ctx, 0xf8f8f8f8, trace->callback, env->uc, pc_start, trace->user_data);

aquynh:
the size of Mips instruction is fixed now, please confirm.

xorstream:
The mips instruiction size fix works.

It still requires a revert or change of the previously altered hook behavior.
I will submit a pull request to fix it.

PR #316, opened 2015-12-15
Fix codehook for MIPS instructions in delay slot

xorstream:
This fixes issue #290.
This reverts code changes that were added to fix issue #133, but that I believe to be incorrect.

Tests "mips_branch_delay.py" and "mips_delay_slot_code_hook.c" should now operate correctly.


aquynh:
merged, thanks!

PR #320, opened 2015-12-16
Added test for uc_open() and uc_emu_start() being called by different threads.

xorstream:
Currently "uc->qemu_global_mutex" is locked by uc_open() and unlocked
by uc_emu_start(). This is a problem because the mutex implementation
must be locked and unlocked by the same thread. This means that uc_open()
and uc_emu_start() must be executed in the same thread. This is an unnecessary
limitation since it is beneficial to be able to run the emulator in its own
thread without having to also do the initial setup for the emulator in the
same thread.

A quick fix for this would be to change the mutex lock to be a semaphore lock,
thereby allowing lock/unlock by different threads. A potential downside is
that sempahores are slower than mutexes, however from what I could see
this probably isn't an issue as "uc->qemu_global_mutex" doesn't appear to
be used when the emulator is running, just during setup.
(I put debug prints in the mutex lock/unlock functions to see when they were
used and there weren't any prints while the emulator was running.)

An alternate fix would be to change the logic of how the
"uc->qemu_global_mutex" lock is used.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 *.dSYM
 *.so
 *.so.*
+*.exe
 
 qemu/config-all-devices.mak
 
--- a/tests/regress/Makefile
+++ b/tests/regress/Makefile
@@ -1,5 +1,11 @@
-CFLAGS += -I../include
+
+CFLAGS += -I../../include
+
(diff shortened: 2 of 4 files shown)


aquynh:
merged, thanks!

PR #323, opened 2015-12-18
Build for Cygwin

iroiro123:
Unicorn can be built for Cygwin.

    $ apt-cyg install make gcc-core pkg-config libpcre-devel zlib-devel libglib2.0-devel
    $ ./make.sh install

--
Cygwin64

    $ uname -srvmo
    CYGWIN_NT-6.1 2.3.1(0.291/5/3) 2015-11-14 12:44 x86_64 Cygwin
    $ gcc -dumpversion; gcc -dumpmachine
    4.9.3
    x86_64-pc-cygwin

--
Cygwin32

    $ uname -srvmo
    CYGWIN_NT-6.1-WOW 2.3.1(0.291/5/3) 2015-11-14 12:42 i686 Cygwin
    $ gcc -dumpversion; gcc -dumpmachine
    4.9.3
    i686-pc-cygwin

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,8 +105,10 @@ else
 IS_CYGWIN := $(shell $(CC) -dumpmachine | grep -i cygwin | wc -l)
 ifeq ($(IS_CYGWIN),1)
 EXT = dll
-AR_EXT = lib
+AR_EXT = a
 BIN_EXT = .exe
+UNICORN_CFLAGS := $(UNICORN_CFLAGS:-fPIC=)
+#UNICORN_QEMU_FLAGS += --disable-stack-protector
 else
 # mingw?
 IS_MINGW := $(shell $(CC) --version | grep -i mingw | wc -l)
@@ -129,7 +131,10 @@ ifeq ($(UNICORN_SHARED),yes)
 ifeq ($(IS_MINGW),1)
 LIBRARY = $(BLDIR)/$(LIBNAME).$(EXT)
 else ifeq ($(IS_CYGWIN),1)
-LIBRARY = $(BLDIR)/$(LIBNAME).$(EXT)
+LIBRARY = $(BLDIR)/cyg$(LIBNAME).$(EXT)
+LIBRARY_DLLA = $(BLDIR)/lib$(LIBNAME).$(EXT).$(AR_EXT)
+$(LIBNAME)_LDFLAGS += -Wl,--out-implib=$(LIBRARY_DLLA)
+$(LIBNAME)_LDFLAGS += -lssp
 else	# *nix
 LIBRARY = $(BLDIR)/lib$(LIBNAME).$(VERSION_EXT)
 LIBRARY_SYMLINK = $(BLDIR)/lib$(LIBNAME).$(EXT)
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -105,8 +105,10 @@ else
 IS_CYGWIN := $(shell $(CC) -dumpmachine | grep -i cygwin | wc -l)
 ifeq ($(IS_CYGWIN),1)
 EXT = dll
-AR_EXT = lib
+AR_EXT = a
```
aquynh: why dont you keep .lib extension? 
the idea was to generate the standard library for Windows, hence .lib.

Makefile
```diff
@@ -114,6 +116,7 @@ ifeq ($(IS_MINGW),1)
 EXT = dll
 AR_EXT = lib
 BIN_EXT = .exe
+UNICORN_CFLAGS := $(UNICORN_CFLAGS:-fPIC=)
```
aquynh: so this fixes MINGW compilation as well, not only CYGWIN?

can you please separate the pull request so it is easier to track down the fix?

aquynh:
besides, if you think it is necessary to extend https://github.com/unicorn-engine/unicorn/blob/master/COMPILE.TXT to include instructions for Cygwin compilation, please do so.

farmdve:
Tbh, I've noticed that the Makefile is sort of broken, not from these commits, from before. Specifying extra eflags is not possible, e.g if I wanted to build Unicorn with -ggdb3 I had to hack the configure file.

aquynh:
feel free to fix what you saw with pull requests. this is the community project, after all :-)

Makefile
iroiro123: cyg*.dll and lib*.a are used for Cygwin but not for native Windows.
Cygwin's gcc can link these, and also Cygwin's python can load them.
If you want to generate a library which can be used for native Windows, you should compile for MinGW.

iroiro123:
I added some instructions to COMPILE.TXT. If my English is broken, please point out the mistake.

Makefile
```diff
@@ -129,7 +131,10 @@ ifeq ($(UNICORN_SHARED),yes)
 ifeq ($(IS_MINGW),1)
 LIBRARY = $(BLDIR)/$(LIBNAME).$(EXT)
 else ifeq ($(IS_CYGWIN),1)
-LIBRARY = $(BLDIR)/$(LIBNAME).$(EXT)
+LIBRARY = $(BLDIR)/cyg$(LIBNAME).$(EXT)
```
aquynh: why do you want `cyglibunicorn.dll`, but not `libunicorn.dll`?

aquynh: see the new branch "cygwin" at https://github.com/unicorn-engine/unicorn/tree/cygwin, in which i merged this pull request. once all the remaining questions answered, i will merge it to "master" branch. thanks

iroiro123: Cygwin's shared libraries are named `cyg*.dll` instead of `lib*.dll`.
https://cygwin.com/cygwin-ug-net/dll.html
> The name of your library is ${module}, prefixed with cyg for the DLL and lib for the import library. Cygwin DLLs use the cyg prefix to differentiate them from native-Windows MinGW DLLs, ...

And, Cygwin's dlopen(3) treats the file name `lib*.so` as `cyg*.dll`.
https://cygwin.com/git/gitweb.cgi?p=newlib-cygwin.git;a=blob;f=winsup/cygwin/dlfcn.cc
get_full_path_of_dll() converts the file name.

aquynh:
merged into the "master" branch, thanks!

PR #329, opened 2015-12-20
Emu stop in hook overrun

xorstream:
Added test for uc_emu_stop() in code hook not always stopping the emu at the current instruction.

In the added code it should stop when executing the instruction at the address 0x100008 but you can see it goes on to execute the instrucion at 0x10000C before then stopping with a PC value being 0x10000C.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -122,6 +122,7 @@ sparc_jump_to_zero
 mem_nofree
 mips_delay_slot_code_hook
 threaded_emu_start
+emu_stop_in_hook_overrun
 
 test_mem_map_ptr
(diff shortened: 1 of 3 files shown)


aquynh:
can you please add `emu_stop_in_hook_overrun` to `.gitignore` as well?

aquynh:
merged, thanks!

aquynh:
fixed, please confirm.

still need to fix this issue for other archs (x86 & mips already fixed).

xorstream:
Fix works for me in the test code and my more complex external project. (Both are MIPS)

PR #330, opened 2015-12-20
Added test for MIPS branch likely instructions.

xorstream:
Test for MIPS branch likely instructions only executing their delay slot instruction when the branch is taken.
Currently it seems to always execute the delay slot instruction like a normal non-"likely" style branch.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -123,6 +123,7 @@ mem_nofree
 mips_delay_slot_code_hook
 threaded_emu_start
 emu_stop_in_hook_overrun
+mips_branch_likely_issue
 
 test_mem_map_ptr
 test_mem_high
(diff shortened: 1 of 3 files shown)


xorstream:
I wonder with this whether the emulator is actually executing the instruction when it should not?
Or perhaps it is not executing it when it should not, but is still calling the code hook for it.

aquynh:
merged, thanks!

aquynh:
this is a bug due to the callback should be executed only when the delay branch is executed. the delay slot concept of Mips is really tricky.

xorstream:
I recall reading a quote by the guy who invented the MIPS instruction set saying if he could go back and start over he wouldn't have included delay slots due to their hassle. :)

PR #331, opened 2015-12-20
Updated mips_branch_likely_issue test to check if executed and hooked.

xorstream:
I updated the mips_branch_likely_issue test to separately check whether the branch likely delay slot is executed by the emulator and whether it has the code hook called for it.

Results show that currently the emulation is correct, ie branch likely delay slot is only executed when the branch is taken, but the codehook is wrong.

The output from the tests is:

Test 1 SHOULD execute the delay slot instruction:
  Emulator did execute the delay slot:  CORRECT
  Emulator did hook the delay slot:  CORRECT

Test 2 SHOULD NOT execute the delay slot instruction:
  Emulator did not execute the delay slot:  CORRECT
  Emulator did hook the delay slot:  WRONG

TEST 1 PASSED!
TEST 2 FAILED!


aquynh:
this is confused:

    Test 2 SHOULD NOT execute the delay slot instruction:
    Emulator did not execute the delay slot: CORRECT

the delay slot must be executed whether the branch is taken or not, so why "did not execute" is CORRECT here? do i miss something?

tests/regress/mips_branch_likely_issue.c
```diff
@@ -33,38 +33,44 @@ Currently it seems to always execute the delay slot instruction like a normal no
...
+// This code SHOULD execute the instruction at 0x100010.
 const unsigned char test_code_1[] = {
-	0x01,0x00,0x02,0x24,	// 100000: li      $v0, 1
-	0x02,0x00,0x03,0x24,	// 100004: li      $v1, 2
-	0x01,0x00,0x62,0x54,	// 100008: bnel    $v1, $v0, 0x100010
-	0x00,0x00,0x00,0x00,	// 10000C: nop
+	0x00,0x00,0x04,0x24,	// 100000: li      $a0, 0
+	0x01,0x00,0x02,0x24,	// 100004: li      $v0, 1
+	0x02,0x00,0x03,0x24,	// 100008: li      $v1, 2
+	0x01,0x00,0x62,0x54,	// 10000C: bnel    $v1, $v0, 0x100010
```
aquynh: the branch address should be 0x100014, not 0x100010 here. typo?

tests/regress/mips_branch_likely_issue.c
```diff
@@ -33,38 +33,44 @@ Currently it seems to always execute the delay slot instruction like a normal no
...
+// This code SHOULD NOT execute the instruction at 0x100010.
 const unsigned char test_code_2[] = {
-	0x01,0x00,0x02,0x24,	// 100000: li      $v0, 1
-	0x01,0x00,0x03,0x24,	// 100004: li      $v1, 1
-	0x01,0x00,0x62,0x54,	// 100008: bnel    $v1, $v0, 0x100010
-	0x00,0x00,0x00,0x00,	// 10000C: nop
+	0x00,0x00,0x04,0x24,	// 100000: li      $a0, 0
+	0x01,0x00,0x02,0x24,	// 100004: li      $v0, 1
+	0x01,0x00,0x03,0x24,	// 100008: li      $v1, 1
+	0x01,0x00,0x62,0x54,	// 10000C: bnel    $v1, $v0, 0x100010
```
aquynh: same here: the branch address should be 0x100014, not 0x100010. typo?

xorstream:
For your first point:
There are two ways that branches handle delay slots.
One is a normal "branch" which always executes the delay slot instruction.
The other is a "branch likely" where the delay slot instruction is only executed if the branch is taken.
This one why I added the two test cases.
Test 1 tests if the branch is taken (and so the delay slot should be executed).
Test 2 tests if the branch is not taken (and so the delay slot should not be executed).

For your second point:
Oops yes, I added the "li      $a0, 0" in at the last minute. The code hex values are still correct, but the comment attached the the "bnel    $v1, $v0, 0x100010" instructions shows the wrong address. So just the comment needs changing.

xorstream:
Btw you can easily tell the difference between a "branch" and a "branch likely" instruction since the "branch likely" instructions usually end in "l".
So "bne" is the "branch" version of "branch if not equal".
And "bnel" is the "branch likely" version of "branch if not equal".

aquynh:
fixed now, please confirm.

xorstream:
Yes the "branch likely" handling of the delay slot is now correct.

Issue #287, opened 2015-12-02
Weird bug in ARM emulation(investigating but need some help about code behaviour)


JCYang:
I emulated some ARMv7(thumb and ARM mixed) code and tried to exam the memory status at some point.
In order to emulate a known imported function, I hooked the entry of the lazy link PLT block(uc_hook_add + UC_HOOK_CODE) of the function and patched it to BX LR(I write r0 in my emulation codes), and it worked.
But a bug found, emulator seems to halt at a certain point later, yet the CPU usage indicate it is still live and running. In order to investigate what's wrong, I tried adding global instruction hook which simply print the instruction ran, in the callback of my patched hook. And weird thing happen, this very bug got "fixed", the engine continue to run to the "until" instruction I specified. 

Some code snippet might better tell the story:

def GlobalHook(uc_emu, address, code_size, usr_data):
      pass   # even place holder helps!!!

def EmulateFunc(uc_emu, address, code_size, usr_data):
      r0 = uc_emu.reg_read(UC_ARM_REG_R0)
      # some calculation here
      uc_emu.reg_write(UC_ARM_REG_R0, result)
      uc_emu.hook_add(UC_HOOK_CODE, GlobalHook) # without this hook, the engine will halt at a certain point later

I've tried to read and modify Unicorn to help diagnose this bug, so I know it halt at a certain point.
And I guess, the hook callback codes make a different to the internal states of the engine, but I'm not familiar with Unicorn codebase, and the code generation engine it used, when I try to look for gen_helper_uc_tracecode from the source, only one invoke location and the definition is nowhere.

I'm continue investigating this bug and would like to fix it, but I really need some help from you, thank you.

aquynh:
Can you please put together a sample code exposing this issue, and send in
a pull request to put it under tests/regress/ or tests/unit/ ?

Thanks.



On Dec 2, 2015 6:17 PM, "JCYang" <notifications@github.com> wrote:
>
> I emulated some ARMv7(thumb and ARM mixed) code and tried to exam the
memory status at some point.
> In order to emulate a known imported function, I hooked the entry of the
lazy link PLT block(uc_hook_add + UC_HOOK_CODE) of the function and patched
it to BX LR(I write r0 in my emulation codes), and it worked.
> But a bug found, emulator seems to halt at a certain point later, yet the
CPU usage indicate it is still live and running. In order to investigate
what's wrong, I tried adding global instruction hook which simply print the
[...]

PR #333, opened 2015-12-21
Fix possible wrong conditional branch in generated host code by fixing the tcg_liveness_analysis().

JCYang:
Refer to https://github.com/unicorn-engine/unicorn/issues/287 for further info.


aquynh:
Can you send in a testcase exposing this problem?

aquynh:
merged, thanks!

Issue #334, opened 2015-12-21
hook_add, hook_del UC_HOOK_CODE does not work properly


michalmalik:
Hello,

I am using the git version from today. Please, don't mind that incorrect hook_code function when hooking UC_HOOK_INTR.

**1) Double hook_add raises no exception with UC_HOOK_CODE**

I've got this code:

```python
from unicorn import *
from unicorn.x86_const import *

ADDRESS = 0x8048000
STACK_ADDRESS = 0xffff000
STACK_SIZE = 4096
CODE = "1\xDBSCSj\x02jfX\x89\xE1"
EP = ADDRESS + 0x54

emu = Uc(UC_ARCH_X86, UC_MODE_32)
emu.mem_map(ADDRESS, 0x1000)
emu.mem_write(EP, CODE)

emu.mem_map(STACK_ADDRESS, STACK_SIZE)
emu.reg_write(UC_X86_REG_ESP, STACK_ADDRESS + STACK_SIZE)

def hook_code(mu, address, size, user_data):
        print(">>> Tracing instruction at 0x%x, instruction size = %u" %(address, size))
        
#####
       
emu.hook_add(UC_HOOK_INTR, hook_code, None)
emu.hook_add(UC_HOOK_INTR, hook_code, None)
       
#####
       
emu.emu_start(EP, EP + len(CODE), count = 3)
print("EIP: 0x%x" % emu.reg_read(UC_X86_REG_EIP))

emu.emu_start(EP, EP + len(CODE), count = 3)
print("EIP: 0x%x" % emu.reg_read(UC_X86_REG_EIP))
```

When executed, this raises an exception, which is *correct*:
```
unicorn.unicorn.UcError: Hook for this type event already exists (UC_ERR_HOOK_EXIST)
```

However, if I change *UC_HOOK_INTR* to *UC_HOOK_CODE*, no exception is raised:
```python
#####
       
emu.hook_add(UC_HOOK_CODE, hook_code, None)
emu.hook_add(UC_HOOK_CODE, hook_code, None)
       
#####
```

```
>>> Tracing instruction at 0x8048054, instruction size = 2
>>> Tracing instruction at 0x8048056, instruction size = 1
>>> Tracing instruction at 0x8048057, instruction size = 1
EIP: 0x8048058
>>> Tracing instruction at 0x8048054, instruction size = 2
>>> Tracing instruction at 0x8048056, instruction size = 1
>>> Tracing instruction at 0x8048057, instruction size = 1
[...]

PR #335, opened 2015-12-22
Add regress test for issue #334

michalmalik:
https://github.com/unicorn-engine/unicorn/issues/334

The code is a bit different than in the issue, but it should demonstrate the main problem just fine.


aquynh:
merged, thanks!

michalmalik:
Is the fact that you can use hook_add(UC_HOOK_CODE, ..) more than once documented somewhere?

```python
for i in range(0, 10):
        print( emu.hook_add(UC_HOOK_CODE, hook_code, None) )
```

```
1
2
3
4
5
6
7
8
9
10
```

PR #336, opened 2015-12-22
Add test for pre-instruction hook being invoked more times.

farmdve:
As I described before, the issue is caused by the particular way some instructions are jitted. The translation block is exited to store the data via the `helper_le_stl_mmu` helper and subsequently re-entered, in this test's case, it's re-entered to check if ECX > 0, invoking the hook one additional time.

This means that counting instructions can be incorrect with some instructions.

Currently, I've added the `rep stosd dword ptr es:[edi], eax` instruction, but likely all instructions prefixed with REP and/or just STOS are going to cause this, and any other instructions which need to exit the TB and re-enter it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -128,6 +128,7 @@ mips_branch_likely_issue
 test_mem_map_ptr
 test_mem_high
 rw_hookstack
+hook_extrainvoke
 
 
 #################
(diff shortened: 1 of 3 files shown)


aquynh:
merged, thanks!

PR #339, opened 2015-12-22
Fix memory read callback to return a value instead of 0

farmdve:
This changes the order of the callback depending on some situations.

@aquynh 
What do you think?

### DIFF
--- a/qemu/softmmu_template.h
+++ b/qemu/softmmu_template.h
@@ -176,7 +176,7 @@ WORD_TYPE helper_le_ld_name(CPUArchState *env, target_ulong addr, int mmu_idx,
     int index = (addr >> TARGET_PAGE_BITS) & (CPU_TLB_SIZE - 1);
     target_ulong tlb_addr = env->tlb_table[mmu_idx][index].ADDR_READ;
     uintptr_t haddr;
-    DATA_TYPE res;
+    DATA_TYPE res = 0;
     int error_code;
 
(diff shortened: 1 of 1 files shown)


farmdve:
@aquynh Are the changes ok?

aquynh:
this sounds ok. will look closer this weekend when i have more time, thanks!

aquynh:
this change modifies the semantics of current UC_HOOK_MEM_READ, in which we no longer get the notification on read event, but only when read is successful. originally, the memory value passed to the callback is *irrelevant*, and should not be referenced during the callback (actually i should not include this parameter for this callback when i designed this API). now with this change, we get the value (since read is successful)

any comments?

lunixbochs:
I like the change. I was previously manually reading the address from unicorn which has way more overhead.

aquynh:
how about this solution: rather than changing the semantics of UC_HOOK_MEM_READ, we add a new hooking type UC_HOOK_MEM_READ_AFTER, which is triggered **after** memory read successes? this requires a new memory event UC_MEM_READ_AFTER, too.

aquynh:
i implemented this as discussed above by adding a new hook type UC_HOOK_MEM_READ_AFTER, and also introduced a new memory access event UC_MEM_READ_AFTER. the commit is https://github.com/unicorn-engine/unicorn/commit/4083b870324a6256242ce266844ff6e7e5bbd7f7. if you see any issues, please report.

PR #342, opened 2015-12-24
(Fix #341) SYSENTER instruction is not properly hooked with uc_hook_add in x86 emulation

### LINKED ISSUE #341: SYSENTER instruction is not properly hooked in x86 emulation.
[helper_sysenter](https://github.com/unicorn-engine/unicorn/blob/ed319bda0b7f7b6bb4a946d24682897e7f465c23/qemu/target-i386/seg_helper.c#L2304) don't check properly if a call interrupt callback is registred with uc_hook_add.

Spl3en:
helper_sysenter in qemu/target-i386/seg_helper.c didn't check properly if a call interrupt callback was registred.
It has been fixed by copying the helper_syscall behavior.

### DIFF
--- a/qemu/target-i386/translate.c
+++ b/qemu/target-i386/translate.c
@@ -7426,10 +7426,14 @@ static target_ulong disas_insn(CPUX86State *env, DisasContext *s,
         if (CODE64(s) && env->cpuid_vendor1 != CPUID_VENDOR_INTEL_1)
             goto illegal_op;
 
-        gen_update_cc_op(s);
-        gen_jmp_im(s, pc_start - s->cs_base);
-        gen_helper_sysenter(tcg_ctx, cpu_env, tcg_const_i32(tcg_ctx, s->pc - pc_start));
-        gen_eob(s);
+        if (!s->pe) {
+            gen_exception(s, EXCP0D_GPF, pc_start - s->cs_base);
+        } else {
+            gen_update_cc_op(s);
+            gen_jmp_im(s, pc_start - s->cs_base);
+            gen_helper_sysenter(tcg_ctx, cpu_env, tcg_const_i32(tcg_ctx, s->pc - pc_start));
+            gen_eob(s);
+        }
         break;
     case 0x135: /* sysexit */
         /* For Intel SYSEXIT is valid on 64-bit */
--- a/qemu/target-i386/translate.c
+++ b/qemu/target-i386/translate.c
@@ -7425,14 +7425,11 @@ static target_ulong disas_insn(CPUX86State *env, DisasContext *s,
         /* For Intel SYSENTER is valid on 64-bit */
         if (CODE64(s) && env->cpuid_vendor1 != CPUID_VENDOR_INTEL_1)
             goto illegal_op;
-        if (!s->pe) {
-            gen_exception(s, EXCP0D_GPF, pc_start - s->cs_base);
-        } else {
-            gen_update_cc_op(s);
-            gen_jmp_im(s, pc_start - s->cs_base);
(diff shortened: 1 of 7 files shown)


qemu/target-i386/translate.c
```diff
@@ -7425,14 +7425,11 @@ static target_ulong disas_insn(CPUX86State *env, DisasContext *s,
         /* For Intel SYSENTER is valid on 64-bit */
         if (CODE64(s) && env->cpuid_vendor1 != CPUID_VENDOR_INTEL_1)
             goto illegal_op;
-        if (!s->pe) {
-            gen_exception(s, EXCP0D_GPF, pc_start - s->cs_base);
-        } else {
```
aquynh: why did you remove this check on `!s->pe`?

Spl3en: I admit it is purely based on my observation on how SYSCALL is handled. Also, during my tests, this condition was true, so there might be another issue explaining why sysenter_helper is not called.

aquynh: if you recover this check (that you removed), is your testcase still working?

aquynh:
also, please dont forget to add `sysenter_hook_x86` to `tests/regress/Makefile` to compile it, and add `sysenter_hook_x86` to `.gitignore`.

qemu/target-i386/translate.c
Spl3en: If I recover this check it doesn't work, the sysenter hook is not called anymore.

Spl3en:
Done.

aquynh:
that branch should not be removed. instead, we need to initialize the engine properly in protected mode. currently the CPU is not in protected mode, hence your observation.

Spl3en:
Alright, I was sure what `pe` meant, I understand better now. I've restored the check.

aquynh:
will be great if you can find how to set engine in protected mode when the engine is created. (hint: the flag HF_PE_SHIFT)

Spl3en:
I think I found the reason why it failed to be in protected mode.

CR0 was not correctly reset in [unicorn.c : x86_reg_reset](https://github.com/unicorn-engine/unicorn/blob/51323c9c17a32f204b169bebc10d1c79553e966f/qemu/target-i386/unicorn.c#L135).
We need to call [cpu_x86_update_cr0](https://github.com/Spl3en/unicorn/commit/bb375e4fa9d68bf2878048d77ef07db7e209f1b1#diff-92173f1ce0d5319bbb03193432e9205dR135) instead of manually setting CR0, so the hflags is also updated correctly.

Now the test passes correctly.

farmdve:
@aquynh 

Do you think we should design another API, for instance called emu_opt_set e.g emu options set where we can specify processor-specific things?

E.g protected mode, long mode, real mode etc. Maybe even other interesting flags.

lunixbochs:
I second this, but I don't think it needs the "emu" prefix. Also kinda related to https://github.com/unicorn-engine/unicorn/tree/cpu_model

farmdve:
@lunixbochs 

I'm just trying to stick to the standards. Not having the emu part would make it look weird compared to the rest of the API.

lunixbochs:
You already have a valid prefix (opt). uc_emu_opt_set is just long-winded.

farmdve:
Ah, brain fart on my part then. I honestly forgot about the uc prefix. uc_opt_set then.

aquynh:
yes i proposed a new API for customize all kind of runtime options, like `uc_option()`. this will be similar to `cs_option()` API of Capstone: https://github.com/aquynh/capstone/blob/master/include/capstone.h#L383

aquynh:
merged, thanks!

Issue #343, opened 2015-12-24
Starting MIPS emulation at 0x80000000 or higher results in fetch from 0 (and PC of 0)


metrzero:
This issue can be reproduced by changing the MIPS sample to set the ADDRESS value to 0x80000000 instead of 0x10000.

metrzero:
It seems as though this is because of the implicit user/kernel mapping. However, Unicorn does not seem to provide an API for setting the MIPS coprocessor registers prior to emulation.

aquynh:
sounds like some issues with the MMU of Mips. would you mind sending in a pull request to put a testcase under `tests/` directory, so we can confirm & test the bugfix?

metrzero:
Looks good on current head.

PR #347, opened 2015-12-28
Regression test for MIPS kernel mode execution

metrzero:
Created a regression test as requested in Issue #343. This test passes when the base address is less than KSEG0 (0x80000000) and currently fails when greater than or equal to KSEG0.


aquynh:
merged, thanks!

metrzero:
Not a problem. Thanks for the quick merge.

aquynh:
fixed, please confirm.

PR #349, opened 2015-12-28
added test for unmap of doubly mapped region

ranmrdrakono:
Crashes if a page is unmapped after another page was mapped twice. Its unclear if mapping twice should fail or unmap should succeed.

### DIFF
--- a/tests/unit/Makefile
+++ b/tests/unit/Makefile
@@ -29,4 +29,4 @@ test_mem_map_ptr: test_mem_map_ptr.c
 test_mem_high: test_mem_high.c
 
 ${ALL_TESTS}:
-	gcc ${CFLAGS} -o $@ $^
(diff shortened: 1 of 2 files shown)


tests/unit/Makefile
```diff
@@ -29,4 +29,4 @@ test_mem_map_ptr: test_mem_map_ptr.c
 test_mem_high: test_mem_high.c
 
 ${ALL_TESTS}:
-	gcc ${CFLAGS} -o $@ $^
+	clang ${CFLAGS} -o $@ $^
```
aquynh: why do you change gcc to clang?

btw, should we change "gcc" to $CC so this works with all compilers?

aquynh:
merged, with modification to reflect the recent change on the semantics of `uc_mem_map()` that it cannot map a new memory range that overlaps existing memory regions. thanks!

PR #357, opened 2016-01-04
Updated dotnet binding

enkomio:
- Refactored solution with a cleaner design
- Added more tests
- Implemented IDisposable interface (and fixed memory leaks)
- Used only CLS datatypes in Unicorn managed library
- Bug fixing


aquynh:
merged, thanks!

PR #360, opened 2016-01-05
Add some fixes to mem_unmap and mem_nofree tests.

farmdve:
My mem_nofree test fails on the latest code as apparently my address and
sizes overlap each other.

The mem_unmap test failed as the invalid memory hook case was incorrect.

Additionally, if mem_map fails for some reason, the new allocated memory region is not stashed, and this creates a problem with my memory leak fixes.


aquynh:
merged, thanks!

PR #361, opened 2016-01-07
Remove more instances of tcg_register_jit.

farmdve:


JonathonReinhart:
Just curious... what does this solve / accomplish? It'd be helpful to include that in the commit message or PR.

farmdve:
Memory leak fix from unused code. And then we removed more unused code in case the code fails to compile on Linux for ARM and the rest.

aquynh:
merged, thanks!

PR #363, opened 2016-01-08
Fix memory leaks as reported by DrMemory and Valgrind.

farmdve:
@aquynh 
I've also fixed that crash on sample_x86 -0. As of now, there are no leaks on Windows, except from InitializeCriticalSection(but this is by design on Windows) on the 32-bit build. And errors reported from sample_x86 -32 on Linux with Valgrind, specifically "Invalid read of size 8" are due to the way QEMU adds pointers as arguments, it stores it as an index and then later replaces it with the actual pointer in tcg_constant_folding, additionally these errors appear even on unicorn master. Works fine under gdb.
It is however weird that it only appears if you call multiple tests e.g test_i386(), test_x86_64, but not when only one test is run. My hypothesis is because some of the pointers lead to same address from the previous instances and Valgrind thinks it was freed and later used again.

ARM and probably the rest of the arches have significant memory leaks as
they have no release interface.

Additionally, DrMemory does not have 64-bit support and thus I can't
test the 64-bit version under Windows. Under Linux valgrind supports
both 32-bit and 64-bit but there are different macros and code for Linux
and Windows.


qemu/unicorn_common.h
```diff
@@ -55,9 +56,48 @@ static void release_common(void *t)
...
+    if(uc->qemu_thread_data)
+        free(uc->qemu_thread_data);
+
+#if TCG_TARGET_REG_BITS == 32
+    for(int i = 0; i < s->nb_globals; i++)
+    {
+        TCGTemp *ts = &s->temps[i];
+        if(ts->base_type == TCG_TYPE_I64)
+        {
+            if(ts->name && ((strcmp(ts->name+(strlen(ts->name)-2), "_0") == 0) || (strcmp(ts->name+(strlen(ts->name)-2), "_1") == 0)))
```
aquynh: can you explain a bit about this code? why did you free only temp variables ending with _0 and _1?

aquynh:
thanks, this is a nice important step towards removing this release blocker!

i merged this into a new branch "memleak". please send next PRs against this branch. 

regrading the release code for non-x86 archs, you can easily add them. you can find why we release the variables in the range https://github.com/unicorn-engine/unicorn/blob/memleak/qemu/target-i386/unicorn.c#L34-L49: that is because we allocated these variables in https://github.com/unicorn-engine/unicorn/blob/memleak/qemu/target-i386/translate.c#L8334-L8362 (in `gen_intermediate_code_internal()`). the same thing should be done for all the other archs to free the memory allocated for these TCG variables in corresponding function gen_intermediate_code_internal() of each arch.

aquynh:
i added a bunch of code to test memleak into `tests/regress/memleak_xxx.c`. compile them with "make" inside `tests/regress` and run them to watch the memory quickly leaked for all the arches - except x86.

qemu/unicorn_common.h
aquynh: can you answer this question?

farmdve: These global temps are created in tcg_global_mem_new_internal and are suffixed with _0 and _1 respectively, when the host is 32-bits and a 64-bit type is needed.

aquynh:
merged the "memleak" branch into the "master" branch, thanks!

Issue #366, opened 2016-01-11
ARM pc-relative VLDR UC_ERR_INSN_INVALID


lunixbochs:
This throws invalid instruction:

    vldr s16, [pc, #244]

This works fine:

    vldr	s15, [r6]

aquynh:
how does Qemu handle this code?
if Qemu 2.2.1 cannot handle it, how about Qemu 2.5.0?

farmdve:
Latest qemu supports the instruction, however their code base has been significantly altered for the ARM arch.

If we are to fix this we must backport all changes as they have disentangled VFP instruction decoding from disas_coproc_insn

aquynh:
@lunixbochs, can you add more code after the call to `emu_start()` to assert new CPU context (registers + memory value), so it is easier to confirm the fix?

thanks

farmdve:
I am attempting to update to 2.3.0 and then to 2.4.0, but this will be a very slow and tedious process with lots of manual patching.

aquynh:
no worry, i will sync straightly to 2.5.0.

farmdve:
How? Most of the commits introduced are breaking.

aquynh:
Manually, of course.

farmdve:
Like, diffing 2.2.1 and current unicorn master, creating a patch, then applying it to 2.5.0 and manually fixing the hunks?

aquynh:
Yes, that is only way.

aquynh:
Ah no, a better way is to diff 2.2.1 and 2.5, then manually patch the diff to current code.

farmdve:
As far as I understood, 2.2.1 is a fork of 2.2.0 with some select patches of master, in essence it's it's own beast so to say. To do it you'd have to diff with 2.2.0.

aquynh:
what is in 2.2.1, but not in 2.5.0?

farmdve:
The qemu developers told me that 2.2.1 is a fork of 2.2.0 with select fixes backported. It means that 2.2.1 doesn't match up with master at any point in time or other releases. Only 2.X.0 e.g 2.2.0, 2.3.0, 2.4.0, 2.5.0 match up with each other as far as i understood it.

aquynh:
actually Unicorn was based on 2.2.0, then i ported it to 2.2.1. and i dont remember anything special in this version at all.

i am not worried anyway. lets see how it works ...

aquynh:
[...]

PR #367, opened 2016-01-11
add regress for #366

lunixbochs:


aquynh:
probably Qemu does not support this encoding yet?

merged, thanks!

lunixbochs:
GCC 4.3.3 generates it, so I'd be a bit surprised. It showed up in a binary I was trying to debug, and I found many instances of it in another binary I had sitting around.

Issue #351, opened 2015-12-29
Reuse of Unicorn Instances


eqv:
Can Unicorn instances be reused after a Start call returned with an error? I am currently Trying to reuse the instances (e.g. running again from a different address), but this seems to lead to inconsistent behavior (e.g. UC_ERR_FETCH_UNMAPPED in the second run at an address that is guaranteed to be mapped, and that is fetched correctly in a third run), even though I reset all mappings / registers. Is this to be expected?

lunixbochs:
Sounds like a bug.

farmdve:
I have been re-using uc_emu_start many times. Although I've hooked the most important stuff like invalid memory and invalid fetch.

Works for me.

aquynh:
if you can reliably reproduce this issue, please send in a testcase.

ranmrdrakono:
Unfortunately I can only reproduce this in very long runs. Fixed this by recreating a new unicorn instance after every stop. If I find the time, I will try to construct a minimal example. Would you accept testcases written in Go instead of C?

aquynh:
no, please keep testcases in C or Python only.

if your testcase is in Go, just post it, and hopefully somebody can help to rewrite it in C or Python.

williballenthin:
I *think* this is also affecting me. Since I haven't yet constructed a minimal test case, you should take the following with the grain of salt, since its just conjecture at this point.

In part of my code, I brute force the emulation of some instructions (to see where jumps and calls go). Occasionally, I end up in nowhere-land. I register a `HOOK_MEM_UNMAPPED` handler to log the events and stop the emulator. When I've detected this case, I reset PC and keep emulating at the next instruction.

*Sometimes* (but not always), the subsequent invocation of `.emu_start` immediately returns with the unmapped memory error code of the previous run. Currently, I handle the bug by retrying the emulation a second time, which seems to (usually?) work.

Again, I could be wrong (it could even be a bug in my code!), but the symptoms sound similar to what OP described.

williballenthin:
Added test case demonstrating issue.

farmdve:
Please re-write your test in C, so that it is easier for me to trace the code with gdb.

williballenthin:
I think you are misreading the test case. The comment `mapped: 0x1000` is intended to show that the `mov` instruction is found at address 0x1000. I can see how you might have understood this to mean that the memory access was to 0x1000, but as you showed, that is not correct.

[...]

Issue #369, opened 2016-01-11
qemu throws assertion on mem_map with size 0


lunixbochs:
from unicorn import *
    uc = Uc(UC_ARCH_X86, UC_MODE_32)
    uc.mem_map(0x2000, 0)

Sorry, in a rush and don't have time to submit a regress PR.

farmdve:
Impossible. 

In uc.c in uc_mem_map

```c
    if (size == 0)
        // invalid memory mapping
        return UC_ERR_ARG;
```

aquynh:
this is now fixed thanks to a PR of @farmdve, thanks.

lunixbochs:
Confirmed.

PR #370, opened 2016-01-11
Reset env->invalid_error before executing a translation block.

farmdve:
Should remove the INVALID memory error when re-using an instance. Should fix issue #351 and a fix for issue #369


aquynh:
the sanity check should be done for uc_mem_map_ptr() as well. i fixed that after merging this.

thanks.

PR #373, opened 2016-01-15
add new api uc_list_mapped_mem

geek-li:
function define is :
 UNICORN_EXPORT  void uc_list_mapped_mem(uc_engine* uc);
input:

execute result  like this:
===========================mapped memory status ===========================

id:0     [range]:0x0000000001000000--0x0000000001200000  [attributes]:RWX
id:1     [range]:0x0000000002002000--0x0000000002202000  [attributes]:---
id:2     [range]:0x0000000003000000--0x0000000003200000  [attributes]:-W-
id:3     [range]:0x0000000004000000--0x0000000004200000  [attributes]:RW-
id:4     [range]:0x0000000005000000--0x0000000005200000  [attributes]:--X

### DIFF
--- a/uc.c
+++ b/uc.c
@@ -1044,3 +1044,50 @@ uc_err uc_hook_del(uc_engine *uc, uc_hook hh)
 {
     return hook_del(uc, hh);
 }
+
+UNICORN_EXPORT
+void uc_list_mapped_mem(uc_engine* uc)
+{
+	unsigned char type[4] = { 'R', 'W', 'X', 0 };
+	uint64_t start;
+	uint64_t end;
+
+	if (uc->mapped_block_count != 0)
+	{
+		printf("\n===========================mapped memory status ===========================\n\n");
+		for (int i = 0; i < uc->mapped_block_count; i++)
+		{
+			unsigned char perms[4] = { '-', '-', '-', 0 };
+			switch (uc->mapped_blocks[i]->perms)
+			{
+			case UC_PROT_NONE:
+				break;
+			case UC_PROT_READ:
+				perms[0] = type[0];
+				break;
+			case UC_PROT_WRITE:
+				perms[1] = type[1];
+				break;
+			case UC_PROT_READ | UC_PROT_WRITE:
+				perms[0] = type[0];
+				perms[1] = type[1];
+				break;
+			case UC_PROT_EXEC:
+				perms[2] = type[2];
+				break;
+			case UC_PROT_ALL:
+				perms[0] = type[0];
+				perms[1] = type[1];
+				perms[2] = type[2];
+				break;
+			default:
+				break;
+			}
+			start = uc->mapped_blocks[i]->addr;
+			end = uc->mapped_blocks[i]->addr + uc->mapped_blocks[i]->size.lo;
+			printf("id:%-2d    [range]:0x%016"PRIx64"--0x%016"PRIx64"  [attributes]:%s  \n", i, start, end, perms);
+		}
+
+	}
+	else printf("can't list mem\n");
+}


lunixbochs:
This functionality is definitely missing right now (I worked around it by tracking mappings in my own code), but I think it's far more fitting and has more utility for Unicorn to return the mappings and expect the user to print them (or do whatever else).

I also think the name should have a uc_mem prefix, like `uc_mem_get_mappings()`

geek-li:
Thx,I think the api name is not a important issue. U  r right and Unicorn  need  more untilty  code.

aquynh:
this is nice, but this API should return a list of memory regions, rather than printing out to the console like this. it is then up to the program calling this API to decide what to do with the list - like printing out, for example.

geek-li:
yes, the  code  need add  this  feature

aquynh:
what you need to do now is to define a new structure (in `unicorn.h`) for the memory region, which contains start_address, end_address, privilege), then make this API return an array of them.

geek-li:
because I am writing  a windows emulator  .so I need these features.

aquynh:
@geek-li thanks a lot for the pull request, but this is not properly implemented in some aspects.

i added a new API named `uc_mem_regions()` doing this at the commit https://github.com/unicorn-engine/unicorn/commit/3d8f413ce8fc836c6b99f3954600c6ecb55b85bb. your program now just needs to call it to retrieve the list of memory regions, then does whatever you want (such as print the list out)

thanks!

geek-li:
OK，Thank  you  for  your work，I  think It’s  a useful  utility    api .

 

发件人: bounces+848413-1bfb-ffgyyihan=126.com@sgmail.github.com [mailto:bounces+848413-1bfb-ffgyyihan=126.com@sgmail.github.com] 代表 Nguyen Anh Quynh
发送时间: 2016年1月16日 17:02
收件人: unicorn-engine/unicorn
抄送: g0tmi1k
主题: Re: [unicorn] add new api uc_list_mapped_mem (#373)

 

@geek-li <https://github.com/geek-li>  thanks a lot for the pull request, but this is not properly implemented in some aspects.

i added a new API named uc_mem_regions() doing this at the commit  <https://github.com/unicorn-engine/unicorn/commit/3d8f413ce8fc836c6b99f3954600c6ecb55b85bb> 3d8f413. your program now just needs to call it to retrieve the list of memory regions, then does whatever you want (such as print the list out)

thanks!

lunixbochs:
Do you see an easy way to refactor to avoid allocations? I call the Usercorn equivalent a ton on my end, so something zero-copy would be preferable.

aquynh:
the point is how can you predict how much memory is needed to contain the list of regions?
i cannot find a simpler API (for user) than dynamically doing malloc from inside the core.

dzzie:
using vs2008 when I call free on the alloced regions  its triggering a breakpoint in RtlValidateHeap. 

Maybe because it was alloced with a different runtime/heap manager? 
I am using the prebuilt binaries 1.0rc2
might need a uc_free_alloc function?

uc_mem_region *regions;
uint32_t count;
uc_err err = uc_mem_regions(uc, &regions, &count);
free(regions);

aquynh:
yes, i think so. can you (temporarily) use `uc_context_free((uc_context *)regions)` to see if it solves your problem.

if this works, then we should rename the API `uc_context_free()` to `uc_free()` or something, to free all kind of memory internally allocated.

dzzie:
yup that did the trick, just had to add it to the def file and regenerate lib. 
stepped through it a bit and it just did a call to msvcrt.free

aquynh:
ok, i will rename `uc_context_free()` to `uc_mem_free()`, and we use this API to free memory allocated by both `uc_context_alloc` & `uc_mem_regions`

aquynh:
i just committed a change to rename uc_context_free() to uc_mem_free(). now you can use this API to free memory allocated by uc_mem_regions(). if you see any issues, please report.

dzzie:
the uc_context* and new uc_mem_free api have not yet made it into unicorn_dynload.c or the msvc def file. It might be worth keeping uc_context_free around for a while since its so small and its removal is a breaking change for any previous code written.

lunixbochs:
I think it's okay to issue breaking API changes on master as there's also a stable version system. The context APIs haven't yet made it into a stable version so they're not being removed from one, hence compat isn't as big of a deal.

In the future, function attrs like `deprecated` and `error` can be used in the header to warn if we break anything.

aquynh:
agreed, lets just update MSVC binding to the latest code.

Issue #364, opened 2016-01-09
IMUL mem/Ib DWORD; OPC_IMUL_GvEvIb; "imul eax, [ecx+0x41], 0x10"


egberts:
In Intel IMUL opcode described in one method using different descriptions:
- 6B / r ib IMUL r32, imm8 doubleword register = doubleword register \* sign-extended immediate byte 
- imul Gv, Ev, I
- `OPC_IMUL_GvEvIb`
- `imul eax, [ecx+0x41], 0x10`
- `6b414110 imul eax,DWORD PTR [ecx+0x41],0x10`

When encountering an IMUL opcode of i386 architecture, specifically when using with an immediate 8-bit multiplier, the emulator engine does not properly multiply the number 0x5151494a by 0x10 and get the expected 0x151494a0 result.  

Interesting thing, is that a standalone IMUL instruction works, but this stack-based AlphaMixed code snippet (derived from Metasploit) failed.

Will submit working proof-of-failure code soon.

aquynh:
yes, please put the testcases under `tests/`. if your code is in C, preferred place is under `tests/unit`.

thanks.

egberts:
Unable to procure a working snippet of code, but I have a clearly documented unit test C file for this troubleshooting of IMUL aex, mem, Ib" and securing a pass/fail scenario.

[test_i386_imul_eax_r_ib.c.zip](https://github.com/unicorn-engine/unicorn/files/85268/test_i386_imul_eax_r_ib.c.zip)

$ unzip test_i386_imul_eax_r_ib.c.zip
$ gcc test_i386_imul_eax_r_ib.c -lunicorn
$ a.out

egberts:
I'm staring at the `tcg_gen_muls2_i32` function... in GDB,  perhaps this is farmdve's cup of tea because this signed IMUL operation got messed before it stores that corrupted value into the EAX register?
```C
        case MO_32:
            tcg_gen_trunc_tl_i32(tcg_ctx, cpu_tmp2_i32, *cpu_T[0]);
            tcg_gen_trunc_tl_i32(tcg_ctx, cpu_tmp3_i32, *cpu_T[1]);
            tcg_gen_muls2_i32(tcg_ctx, cpu_tmp2_i32, cpu_tmp3_i32,
                              cpu_tmp2_i32, cpu_tmp3_i32);
            tcg_gen_extu_i32_tl(tcg_ctx, *cpu_regs[reg], cpu_tmp2_i32);
[...]

PR #378, opened 2016-01-16
refactor to allow multiple hooks for one type

### LINKED ISSUE #375: overlapping hooks aren't very robust
Different projects using Unicorn (or different components of the same project) can't overlap very well right now because many of the hook forms expect to be the only hook installed.

The only exception I know is hooks with range parameters, of which you can have multiple, as long as you don't install a hook with no range parameters.

This isn't very powerful, and doesn't lend itself to making components that extend Unicorn nicely. If I made a code library that implements stack tracing internally, it's much nicer for it to maintain its own Unicorn hook than to centralize everything in my code around a single code hook, a single memory hook, etc.

I'm actually already running into this in Usercorn, where I have a main instance that manages a binary, runs it, handles syscalls, etc, but I also have a networked debugger that wants to attach additional hooks. It'd be nice if I could just pile more hooks on top of Unicorn, but that's not really an option right now.

----

TLDR: I think the hook system needs some love. The success of `hook_add()` should never depend on existing hooks.

lunixbochs:
This fixes #375

Some details in https://github.com/unicorn-engine/unicorn/issues/375#issuecomment-172161759


list.c
```diff
@@ -0,0 +1,57 @@
...
+#include <stdint.h>
+#include "list.h"
+
+// simple linked list implementation
+
+struct list *list_new(void) {
+    return calloc(1, sizeof(struct list));
+}
+
+void list_clear(struct list *list) {
```
aquynh: we have a common coding style, in which the open bracket is put on a new line, alone. can you please fix this?

qemu/softmmu_template.h
```diff
@@ -184,17 +184,22 @@ WORD_TYPE helper_le_ld_name(CPUArchState *env, target_ulong addr, int mmu_idx,
...
     if (mr == NULL) {
+        struct hook *hook;
+        int handled = 0;
 #if defined(SOFTMMU_CODE_ACCESS)
         error_code = UC_ERR_FETCH_UNMAPPED;
-        if (uc->hook_mem_fetch_idx != 0 && ((uc_cb_eventmem_t)uc->hook_callbacks[uc->hook_mem_fetch_idx].callback)(
-                                       uc, UC_MEM_FETCH_UNMAPPED, addr, DATA_SIZE, 0,
-                                       uc->hook_callbacks[uc->hook_mem_fetch_idx].user_data)) {
+        uc_hook_loop_range(uc, hook, UC_HOOK_MEM_FETCH_UNMAPPED, addr) {
+            if ((handled = ((uc_cb_eventmem_t)hook->callback)(uc, UC_MEM_FETCH_UNMAPPED, addr, DATA_SIZE, 0, hook->user_data)))
```
aquynh: we still have improper indentation here and there ...

aquynh: looks like you put this indentation here because it belongs to the inner block?

qemu/target-arm/translate.c
```diff
@@ -10408,15 +10407,11 @@ static void disas_thumb_insn(CPUARMState *env, DisasContext *s) // qq
...
-    if (env->uc->hook_insn) {
-        struct hook_struct *trace = hook_find(s->uc, UC_HOOK_CODE, s->pc);
-        if (trace)
-            gen_uc_tracecode(tcg_ctx, 2, trace->callback, env->uc, s->pc, trace->user_data);
-        // if requested to emulate only some instructions, check to see
-        // if we need to exit immediately
-        if (env->uc->emu_count > 0) {
-            check_exit_request(tcg_ctx);
-        }
+    struct hook *hook;
```
aquynh: we are following C89, so please declare all the variables at the beginning of block

qemu/softmmu_template.h
lunixbochs: indentation here looks correct to me. this is a loop macro to make looping over the hooks not a nightmare. I originally had a visible `for ()` statement involved but I couldn't get it to work with c89.

lunixbochs:
New commit with c89 and code style fixes.

list.c
```diff
@@ -0,0 +1,62 @@
...
+    }
+    item->prev = list->tail;
+    item->next = NULL;
+    item->data = data;
+    if (list->head == NULL) {
+        list->head = item;
+    }
+    if (list->tail != NULL) {
+        list->tail->next = item;
+    }
```
aquynh: do you miss the "else" path here, in which you need to set `list->tail = item`?

list.c
```diff
@@ -0,0 +1,62 @@
...
+                cur->prev->next = cur->next;
+            }
+            if (cur->next != NULL) {
+                cur->next->prev = cur->prev;
+            }
+            free(cur);
+            break;
+        }
+        cur = next;
+    }
```
aquynh: dont you consider that "head" and "tail" can be modified in this function, when you remove a list item?

include/uc_priv.h
```diff
@@ -60,16 +60,53 @@ typedef bool (*uc_args_int_t)(int intno);
...
+    struct list_item *cur;                                                 \
+    for (                                                                  \
+    cur = (uc)->hook[idx##_IDX].head;                                      \
+    cur != NULL && ((hh) = (struct hook *)cur->data)                       \
+        && !uc->stop_request /* stop excuting callbacks on stop request */ \
+        extra;               /* extra is used below for address ranges */  \
+    cur = cur->next)
+
+// loop without address check
+#define uc_hook_loop(uc, hh, idx) uc_hook_loop_common(uc, hh, idx, )
```
aquynh: for consistency, what do you think about renaming these macro to use uppercase instead? i was a bit confused when i first saw this.

for the loop macro, should we use more traditional name, such as HOOK_FOREACH?
(Qemu is using macros such as CPU_FOREACH, QTAILQ_FOREACH etc)

uc.c
```diff
@@ -935,206 +918,72 @@ MemoryRegion *memory_mapping(struct uc_struct* uc, uint64_t address)
...
-
-    if (type & UC_HOOK_MEM_READ_PROT)
-        ret = _hook_mem_invalid(uc, UC_HOOK_MEM_READ_PROT, callback, user_data, hh);
-
-    if (type & UC_HOOK_MEM_WRITE_PROT)
-        ret = _hook_mem_invalid(uc, UC_HOOK_MEM_WRITE_PROT, callback, user_data, hh);
-
-    if (type & UC_HOOK_MEM_FETCH_PROT)
-        ret = _hook_mem_invalid(uc, UC_HOOK_MEM_FETCH_PROT, callback, user_data, hh);
+    struct hook *hook = calloc(1, sizeof(struct hook));
```
aquynh: i dont see anywhere this "hook" is freed. list_remove() and list_clear() dont do this, so we are having memory leaking issue now?

uc.c
```diff
@@ -284,7 +278,9 @@ uc_err uc_close(uc_engine *uc)
     // TODO: remove uc->root    (created with object_new())
     uc->root->free(uc->root);
 
-    free(uc->hook_callbacks);
+    for (i = 0; i < UC_HOOK_MAX; i++) {
+        list_clear(&uc->hook[i]);
```
aquynh: list_clear() does not free the hooks we did malloc() before

uc.c
```diff
@@ -935,206 +918,72 @@ MemoryRegion *memory_mapping(struct uc_struct* uc, uint64_t address)
...
     return ret;
 }
 
 UNICORN_EXPORT
 uc_err uc_hook_del(uc_engine *uc, uc_hook hh)
 {
-    return hook_del(uc, hh);
+    int i;
+    for (i = 0; i < UC_HOOK_MAX; i++) {
+        list_remove(&uc->hook[i], (void *)hh);
```
aquynh: this does not look good, because this loop always go thru all the entries even list_remove() already removed the hook at some point.

also, the hook is not freed anywhere, so memory is leaking.

uc.c
```diff
@@ -935,206 +918,72 @@ MemoryRegion *memory_mapping(struct uc_struct* uc, uint64_t address)
...
-            begin = va_arg(valist, uint64_t);
-            end = va_arg(valist, uint64_t);
-            ret = _hook_mem_access(uc, UC_HOOK_MEM_WRITE, begin, end, callback, user_data, hh);
-            break;
-        case UC_HOOK_MEM_READ | UC_HOOK_MEM_WRITE:
-            begin = va_arg(valist, uint64_t);
-            end = va_arg(valist, uint64_t);
-            ret = _hook_mem_access(uc, UC_HOOK_MEM_READ | UC_HOOK_MEM_WRITE, begin, end, callback, user_data, hh);
-            break;
+    // everybody but HOOK_INSN gets begin/end, so exit early here.
```
aquynh: why dont we also give begin/end options to HOOK_INSN? i dont see why we should limit this option in this case. just give the users a choice, and they can decide when to use it?

aquynh:
overall, a very nice work! look forward to all the remaining issues cleared so we can merge this PR, thanks.

uc.c
lunixbochs: This removes the hook from each *type*, like if you have a MEM_READ and MEM_WRITE with the same hook struct. I've updated to use reference counting to track how many lists a hook is in and free appropriately.

uc.c
lunixbochs: The current ABI does not allow for this. This patch doesn't change the ABI in any way.

include/uc_priv.h
lunixbochs: switching to HOOK_FOREACH and HOOK_FOREACH_RANGE (for address range checking)

list.c
lunixbochs: fixed

list.c
lunixbochs: fixed

lunixbochs:
New commit(s) are up. I pushed the changes on their own to make them obvious, then rebased to make this one commit.

lunixbochs:
I'd also like to point out that this linked list implementation is not thread safe, so Unicorn needs to lock in `uc_hook_add()` if it wants thread-safe hook manipulation.

Also, calling hooks would need to read-lock in case a hook is deleted, but I don't like the performance implications.

Instead we could disallow hook deletion while the emulator is running (I think adding hooks is safe as the write for `list->tail->next = item` will hopefully be generated as aligned/atomic)

aquynh:
yes thread-safe is important to have.

aquynh:
please go ahead breaking compatibility by adding begin/end range support to all kind of hooks. we will possibly break API on `uc_open()` with the branch https://github.com/unicorn-engine/unicorn/tree/cpu_model anyway.

aquynh:
i will bump API version to 1.0 to reflect the API change for the next version.

lunixbochs:
Updating the bindings for begin/end on hook_insn is going to be a bit of work. I'd rather move that to a new PR and get this merged sooner so stuff like @farmdve's patch can be updated.

lunixbochs:
I pushed again with 4e7300e added to fix a bug chaining bounds-checked hooks (if any hook failed to match the address range it would stop looking for hooks).

list.c
```diff
@@ -0,0 +1,72 @@
...
+        list->head = item;
+    } else {
+        list->tail->next = item;
+    }
+    list->tail = item;
+    return item;
+}
+
+// returns 1 if entry was removed, 0 otherwise
+int list_remove(struct list *list, void *data)
```
aquynh: should we make this to return `bool` instead of `int`?

qemu/softmmu_template.h
```diff
@@ -178,23 +178,31 @@ WORD_TYPE helper_le_ld_name(CPUArchState *env, target_ulong addr, int mmu_idx,
     uintptr_t haddr;
     DATA_TYPE res;
     int error_code;
+    struct hook *hook;
+    int handled;
```
aquynh: "handled" should have type "bool", not "int", because "uc_cb_eventmem_t" return "bool" value

include/list.h
```diff
@@ -0,0 +1,18 @@
+#ifndef UC_LLIST_H
+#define UC_LLIST_H
+
+struct list_item {
+    struct list_item *prev, *next;
```
aquynh: do we really need double-linked list? i think one-way linked list is enough (?)

qemu/target-i386/translate.c
```diff
@@ -4768,16 +4767,16 @@ static target_ulong disas_insn(CPUX86State *env, DisasContext *s,
...
     // Unicorn: trace this instruction on request
-    if (env->uc->hook_insn) {
-        trace = hook_find(env->uc, UC_HOOK_CODE, pc_start);
-        if (trace) {
-            if (s->last_cc_op != s->cc_op) {
-                sync_eflags(s, tcg_ctx);
-                s->last_cc_op = s->cc_op;
-                changed_cc_op = true;
-            }
+    if (HOOK_EXISTS(env->uc, UC_HOOK_CODE)) {
```
aquynh: does this change the logic of original code? originally, we have this "if" block after the range check (after `hook_find()`), but now this "if" block skips the check totally.

lunixbochs: yeah, this is actually a bug. HOOK_EXISTS should also have a bounds check variant.

lunixbochs:
615cb0e switches to bool and singly-linked list

lunixbochs:
acdc1ef adds a bounded HOOK_EXISTS check

list.c
```diff
@@ -0,0 +1,66 @@
...
+    struct list_item *item = malloc(sizeof(struct list_item));
+    if (item == NULL) {
+        return NULL;
+    }
+    item->next = NULL;
+    item->data = data;
+    if (list->head == NULL) {
+        list->head = item;
+    } else {
+        list->tail->next = item;
```
aquynh: this should be removed because list no longer has tail

aquynh: oh no we still have `tail`, but is it still necessary given the list is single now?

qemu/softmmu_template.h
aquynh: i guess you missed this in the latest fix?

qemu/softmmu_template.h
```diff
@@ -595,24 +639,30 @@ void helper_le_st_name(CPUArchState *env, target_ulong addr, DATA_TYPE val,
     int index = (addr >> TARGET_PAGE_BITS) & (CPU_TLB_SIZE - 1);
     target_ulong tlb_addr = env->tlb_table[mmu_idx][index].addr_write;
     uintptr_t haddr;
+    struct hook *hook;
+    int handled;
```
aquynh: "handled" here should be "bool", too

qemu/softmmu_template.h
```diff
@@ -742,24 +797,30 @@ void helper_be_st_name(CPUArchState *env, target_ulong addr, DATA_TYPE val,
...
```
aquynh: "handled" here should be "bool", too

qemu/target-arm/translate-a64.c
```diff
@@ -10984,10 +10985,10 @@ static void disas_a64_insn(CPUARMState *env, DisasContext *s)
...
 
     // Unicorn: trace this instruction on request
-    if (env->uc->hook_insn) {
-        struct hook_struct *trace = hook_find(s->uc, UC_HOOK_CODE, s->pc - 4);
-        if (trace)
-            gen_uc_tracecode(tcg_ctx, 4, trace->callback, env->uc, s->pc - 4, trace->user_data);
+    HOOK_FOREACH(env->uc, hook, UC_HOOK_CODE) {
+        if (! HOOK_BOUND_CHECK(hook, s->pc - 4))
+            continue;
+        gen_uc_tracecode(tcg_ctx, 4, hook->callback, env->uc, s->pc - 4, hook->user_data);
```
aquynh: i have a concern here: we insert multiple calsl to TCG code here (in a loop), but is it better to just have a single callback that multiplex to all the registered hooks?

we can merge this for now, but keep this in mind for further consideration in the future.

aquynh: actually we have another place down there that patches the block size & code size, and this gets all wrong since it only patch the first callback's parameter. this problem only shows if we register more than 1 callback.

so have a single callback that multiplex to all the registered hooks as i mentioned above makes sense now.

uc.c
```diff
@@ -935,200 +931,63 @@ MemoryRegion *memory_mapping(struct uc_struct* uc, uint64_t address)
...
+    if (type & UC_HOOK_INSN) {
+        hook->insn = va_arg(valist, int);
+        hook->begin = 1;
+        hook->end = 0;
+        if (list_append(&uc->hook[UC_HOOK_INSN_IDX], hook) == NULL) {
+            free(hook);
+            return UC_ERR_NOMEM;
+        }
+        hook->refs++;
```
aquynh: is this "refs" properly implemented? i fail to see where we can bump "refs" up after the first time (?)

aquynh:
can you please write a sample that registers more than 1 callback of the same hook type to test this new feature? attention is on the most tricky hook types UC_HOOK_BLOCK or UC_HOOK_CODE, where we have to patch TCG call's parameter as in my comments above.

we can either put this sample under `tests/unit/` or `samples/`. thanks

qemu/target-arm/translate-a64.c
farmdve: Do not place the hook where we patch the size arguments, I've tried it and it won't work in certain cases.

uc.c
```diff
@@ -935,200 +931,63 @@ MemoryRegion *memory_mapping(struct uc_struct* uc, uint64_t address)
...
+        if ((type >> i) & 1) {
+            // TODO: invalid hook error?
+            if (i < UC_HOOK_MAX) {
+                if (list_append(&uc->hook[i], hook) == NULL) {
+                    if (hook->refs == 0) {
+                        free(hook);
+                    }
+                    return UC_ERR_NOMEM;
+                }
+                hook->refs++;
```
lunixbochs: refs is bumped here

list.c
lunixbochs: tail is a shortcut to the last item so we don't need to iterate over the list to append to it

aquynh: oh right i forgot this, so please ignore above comments

qemu/target-arm/translate-a64.c
aquynh: what do you mean? is there any samples where callback doesnt work?

lunixbochs: A single hook will be slower because the address bounds check won't happen at JIT time. I think we should just patch the size arguments for all of them.

uc.c
aquynh: ok i see. i missed this part

qemu/target-arm/translate-a64.c
aquynh: what do you mean by "address bounds check work happen at JIT time"? we just need a callback managed inside the core, and this callback works exactly like before this PR, but inside it just call all the user callbacks one by one.

farmdve: @aquynh 

When I was implementing the post instruction hook I had initially placed it where we patched the size argument. But because of how certain instruction got jitted, there were instances where the hook was never called(direct jumps, ret, and rep prefixed instructions).

lunixbochs: Does this mean there are instances where the size won't be patched?

aquynh: @farmdve if you are talking about bugs in current pre-hook type, that should be fixed. please send in testcases for what you found.

aquynh: ah i see what you mean, but the difference is minimal that i doubt it makes much difference.

ok .can do what you want but a caveat: the distance from the beginning of JIT buffer to the size parameter between 32bit & 64bit host machine can be difference, so consider this when you patch

lunixbochs:
I added a test for two basic block hooks, which fails to patch size as expected.

aquynh:
if you still want to insert multiple calls to TCG buffer, there is one thing that you overlooked: the TCG buffer size is limited, and when the buffer is full, then the frontend translation exits to create another buffer. so there is the limitation on the number of hooks you can insert in one buffer. you have to take care this when you still have to insert more calls into the next buffer.

farmdve:
Can you guys remind me of the purpose of multiple hooks? Sounds like more overhead to me. Each hook incurs about 6-7 more JITTED instructions per single host instruction.

lunixbochs:
Example of simultaneous HOOK_CODE uses in Usercorn, which are at entirely different places in my codebase and basically need a central dispatch (whether it's in Unicorn or Usercorn):

- ARM kernel trap page (code at 0xffff0000).
- Non-interactive debugging (for register diffs, will be used for stuff like taint analysis).
- Single stepping, watchpoints, etc in interactive debugger.

farmdve:
@lunixbochs 

Any ETA for your refactor?
Also, not sure if it's relevant or not, but shouldn't we attempt to port to 2.5.0 qemu code base?

aquynh:
It will take a good time to port to 2.5.0, and this work has no conflict
with new port, so go ahead doing whatever you planned, dont wait for the
port.

Thanks

lunixbochs:
cbd1a3c1f3dde4cea72fddf35230bda512c8739b fixes the patching issue of HOOK_CODE and HOOK_BLOCK by moving the loop to helper_uc_tracecode

lunixbochs:
I also rebased against master so this should cleanly merge again. After the code review of [rework code/block tracing[(0886ae8), are we missing anything else?

aquynh:
if you take at the output of the master branch before and after this PR is merged, you will see some differences.

    $ samples/sample_all.sh > sample_output.txt

lunixbochs:
It's probably all differences in the bounds checking. Isn't this wrong? From sample_x86.c:

    uc_hook_add(uc, &trace1, UC_HOOK_BLOCK, hook_block, NULL, (uint64_t)ADDRESS, (uint64_t)ADDRESS);

aquynh:
no i dont think the samples are wrong.

so after your fix, the remaining difference is the number of instructions executed? why is this difference?

lunixbochs:
I'm guessing something changed with stop requests

lunixbochs:
I think the memory callbacks are broken because I added begin/end range to the invalid memory hooks, and the samples don't set begin/end.

aquynh:
Yes all related samples in C need to be fixed.

On the bindings, we just need to make a change to have begin >  end as
default, so no change in binding samples is needed.

lunixbochs:
New diff:

    513c513
    < not ok - Fetch from non-executable memory at 0x10101d
    ---
    > not ok - Fetch from non-executable memory at 0x10101b
    516c516
    < Executed 4122 instructions
    ---
    > Executed 4120 instructions

aquynh:
do you have any ideas why we have these 2 differences?

lunixbochs:
not yet.

lunixbochs:
I think this is related to TB caching. Shouldn't `uc_mem_protect()` flush the TB to fault immediately?

master branch: https://bochs.info/p/n6k7mq
hook-refactor branch: https://bochs.info/p/7gvtn

aquynh:
i dont mean your code is buggy, but we still need to understand why these differences.

i pushed a new branch "hook" with this PR merged: https://github.com/unicorn-engine/unicorn/tree/hook
will merge this branch back to "master" when it is ready.

thanks.

lunixbochs:
You're currently allowed to execute some number of instructions in an NX page without faulting if you mem_protect() during HOOK_CODE (likely the remainder of your TCG buffer).

This PR decreased the number of instructions emitted for HOOK_CODE, as it went from a 5-argument to a 4-argument helper. This slightly increases the number of instructions that fit in a TCG buffer, allowing two more guest instructions to execute before the fault.

If I add another TCG parameter inside `gen_uc_tracecode()` the fault happens two instructions earlier as before.

aquynh:
yes, if UC_PROT_EXEC is removed from a page's permission, the related page should be flushed. in other cases it seems we can ignore.

farmdve:
So can anyone bring me up to speed? The hook is being called, because QEMU is executing instructions from a protected area, despite the fact that it should exit before it does so?

aquynh:
yes, code execution must exit immediately after EXE permission is removed, but this is quite tricky to implement.

perhaps this issue is related to https://github.com/unicorn-engine/unicorn/pull/380 ?

aquynh:
Ryan, this is a very nice work! merged, thanks!

now please go ahead adding begin/end support to all the rest of hook types.

aquynh:
i just pushed a commit to fix the issue of `mem_apis.c`, in which execution fails to quit immediately when EXEC permission is removed: https://github.com/unicorn-engine/unicorn/commit/e750a4e97c4db33aa00866eb4875098b871ce4a8

see the difference made by commit in for `mem_apis.c`. all the numbers makes sense now.

```
 ===================================
 # Example of marking memory NX (faulting)
 BEGINNING EXECUTION
-not ok - Fetch from non-executable memory at 0x10101b
+not ok - Fetch from non-executable memory at 0x101000
 not ok - Failure on uc_emu_start() with error 14: Fetch from non-executable memory (UC_ERR_FETCH_PROT)
 FAILED EXECUTION
-Executed 4120 instructions
+Executed 4094 instructions
```

egberts:
I believe I can account for that discrepancies in the executed instruction count (is that still a problem for you as it is for me now as I'm now using today's GIT master (75e5fb4).?)

In the same [`test_tb_x86.c`](https://github.com/unicorn-engine/unicorn/blob/master/tests/unit/test_tb_x86.c) (in tests/unit), this discrepancy is **NOW** found *EARLIER* in the test code whose opcode sequence being UC_HOOK_CODE'd twice in Unicorn.  (Tweaked `test_tb_x86.c` RIP/PC to start earlier at 0x60000021).

```asm
; ECX=0x5ffffff8
; memory being modified at 0x60000028 (self-modify scenario)
60000021 304130    xor    [ecx+0x30], al
60000024 41        inc    ecx
```

The TCG buffer had two UC_HOOK_CODE_IDX entries for helper to process on before emulating the `xor` opcode.

It APPEARS to be specific to indexed-operand. This is well-documented in #364 (and is also happening before the  problematic #364 issue and GDT-related issues). This is now my primary focus.

My plan is to determine how many UC_HOOK_CODE_IDX were initially created at start_thread() and then how many were there at TCG regeneration.  Earlier recollection shows one at initial start_thread(), and two at regeneration due to self-modify of its own resident code block condition.

PR #379, opened 2016-01-16
Add unit test case for Issue #364 "Self-Modifying x86 code".

egberts:
A pretty catty unit test case with intent on demonstrating that the instruction translation cache gets to be flushed properly after an x86-based self-modifying program modifies the immediate byte (Ib(of the **IMUL Gv,Ev,Ib" opcode sequence.

Failure is reported with an exit code of 255 (byte -1).
Success of properly flushed instruction translation cache buffer is reported with an exit code of 0.


egberts:
Withdrawing pull request as it was supposed to be located under tests/unit (not tests/regress).

PR #380, opened 2016-01-16
Added unit test for TB flushing by x86_32 Self-modify of IMUL Gv, Ev, Ib instruction sequence

egberts:
for exercising proper flushing of the instruction translation cache.


aquynh:
merged with some minor fixes, thanks!

PR #388, opened 2016-01-22
Altered configure to generate config-all-disas.mak with no new line

xorstream:
The newline at the end of config-all-disas.mak was causing issues with "\r\n" vs "\n" and so it would appear to be modified when it was not.

If you don't like this fix then perhaps config-all-disas.mak could instead be added to the ignore list so that git doesn't keep trying to commit the changes to it due to the different newline.


aquynh:
actually we dont need `config-all-disas.mak`, so i removed it altogether with the commit https://github.com/unicorn-engine/unicorn/commit/309abbe45370b72ffef3596d9337b2970d88cf38. now you dont need to bother anymore, thanks!

xorstream:
problem solved :)

PR #392, opened 2016-01-22
Uc mode sort

xorstream:
Sorted uc_modes by CPU and added masks


aquynh:
do we really need to expose the masks in unicorn.h? 

if users do not need this, we better move them to include/uc_priv.h or somewhere else.

also, you are putting these masks inside "enum uc_mode", but the right place for them is outside.

thanks.

xorstream:
Putting them here shows the user what modes are currently supported. Also keeping them close to the mode values themselves helps ensure that changes to the mode values will also get propagated into the mask values.

aquynh:
but what we put inside the uc_mode enum is all the hardware modes. the masks are not hardware mode, so they do not belong there.

xorstream:
Should I make them as a #define them in uc_priv.h?
I'll do it tomorrow if so.

aquynh:
i think users do not need these masks, so moving them to uc_priv.h sounds good to me.

thanks.

xorstream:
Changes masks to #defines in uc_priv.h

xorstream:
I have been going through and checking all mode flag usage and I realised that while the UC_MODE_LITTLE_ENDIAN mode flag is the default mode, little endian is not the default endian for all processor types. In fact it is only the default for x86 and ARM. Big endian is the default for MIPS and PPC, as well as M68K and SPARC which only support big endian.

The issue is how to handle specifying flags for the big endian processors?

For MIPS and PPC the UC_MODE_BIG_ENDIAN flag should be required to differentiate from UC_MODE_LITTLE_ENDIAN which they can support as well.

Should M68K and SPARC, which are big endian only, be required to pass in the UC_MODE_BIG_ENDIAN flag to uc_open()? If they don't pass in the flag then it is the same as passing in the UC_MODE_LITTLE_ENDIAN flag. So it seems to me that the flag *should* be passed in to check against trying to use little endian for them.

aquynh:
i think for those with only 1 mode (either big or little endian), we can just skip the sanity verification because even if the wrong mode is passed in, it does no harm. i dont see much value to enforce tight check here.

xorstream:
It might do no harm internally, but the user might think they have successfully set a big endian processor to be little endian since the little endian flag (UC_MODE_LITTLE_ENDIAN) is 0.

xorstream:
I'll close this since the other uc_mode_fix PR that was already merged contains the same fixes this first added.

PR #395, opened 2016-01-23
Fix uc_mode usage in source code

xorstream:
Fixed usage of uc_modes in source code including:
* doing bitflag style '&' checks instead of == or !=.
* labelling currently unsupported modes in the header file
* doing checks on supported flags for each arch type in uc_open()
* doing checks on supported endians for each arch type in uc_open()
* fixes to mask values for big endian processors


aquynh:
this looks good, but it doesnt compile due to some missing `)`. can you fix that?

thanks.

xorstream:
heh yeh I just noticed that, compile fix commited

aquynh:
now compilation is OK, but if you run `samples/sample_arm` you will see error:

    Failed on uc_emu_start() with error returned: 8

xorstream:
All samples seem ok now.

Though I don't really think the checks need to be in unicorn_arm.c for arm_reg_read() and arm_reg_write(). Checks are already done to ensure the mode is valid in uc_open() and won't proceed if not.

So what do you think about removing the check in unicorn_arm.c on lines 45 and 81:
    if (!(uc->mode & ~UC_MODE_ARM_MASK)) {

aquynh:
yes you should remove those unnecessary lines.

aquynh:
because your code fails to compile now, again due to some missing ')'

xorstream:
ehehe sorry I must have not committed correctly since I did test here and it compiled and ran all samples ok.

aquynh:
are you sure the mode we are using for Sparc is SPARC64?

xorstream:
I assumed that you set the UC_MODE_SPARC64 flag if you want 64bit and leave it unset if you want 32bit. I made this assumption based on this code from uc_open():

                if (mode & UC_MODE_SPARC64)
                    uc->init_arch = sparc64_uc_init;
                else
                    uc->init_arch = sparc_uc_init;

xorstream:
We could add explicit 32 bit flags for each processor, basically an alias to:
    UC_MODE_32 = 1 << 2
And then check that it is set in uc_open()

Or we could add a "default" flag like:
    UC_MODE_SPARC32 = 0,
If you wanted to make it clearer that it is 32bit by default.

I think we should choose one style and do the same for all CPUs though so that the usage is clear. So either always have to specify 32 bit, or assume that no flags (ie 0) means 32 bit.
If we go with the second options then we should remove the existing flags:
    UC_MODE_32
    UC_MODE_MIPS32

aquynh:
merged, thanks!

to be consistent, i addes UC_MODE_SPARC32 (= UC_MODE_32). so passing 0 as Sparc mode is considered wrong now.

on the idea of removing some modes, we should avoid that for compatibility reason. lets keep them, but we can add some alias modes if needed.

PR #399, opened 2016-01-24
Added ppc 32bit mode and added sparc mode checks

xorstream:
Added ppc 32bit mode and added sparc mode checks to bring it in line with other archs


aquynh:
merged, thanks!

PR #401, opened 2016-01-25
Fixed mode flags for sparc tests

xorstream:
Added mode endian flags for sparc tests and fixed mode in sparc_jump_to_zero.c


aquynh:
merged, thanks!

you might want to fix the hardware modes for Sparc in all the code under `samples/` and `bindings/*`, too?

Issue #292, opened 2015-12-04
Can I call uc_emu_start() and uc_emu_stop() from different threads?


xorstream:
Should I be able to start the emu in one thread by doing uc_emu_start(), and then stop it from another thread by doing uc_emu_stop()? My aim is to have the emulation running in the background and be able to stop it from another thread.

When I attempt to do so in windows I get an error when I call uc_emu_start() which I think is due to it being called from a different thread to the one that I called uc_emu_open() from.

Error msg:
Assertion failed!

Program: C:\unicorn_msvc\dyn\Debug\dyn.exe
util/qemu-thread-win32.c, Line 70

Expression: mutex->owner == GetCurrentThreadId()

This application has requested the Runtime to terminate it in an unusual way.
Please contact the application's support team for more information.

This points to the following assertion:
    void qemu_mutex_unlock(QemuMutex *mutex)
    {
        assert(mutex->owner == GetCurrentThreadId());
        mutex->owner = 0;
        LeaveCriticalSection(&mutex->lock);
    }

So I am wondering if this behaviour is by design? And if so what is the expected way to run the emu in the background?

If a thread is busy running uc_emu_start() then it cannot do its own calls to uc_emu_stop(), except for in the callback hooks or if uc_emu_start() had a count or timeout limit. And so if other threads also cannot call uc_emu_stop() then I guess I have to do something like check a flag after a timeout or in code hooks for if uc_emu_stop() should be called.

farmdve:
I was just about to come here to ask about thread-safety myself.

I basically wish to have a GUI thread that displays registers, maybe even memory, but I am not too sure how thread-safe Unicorn is, can I start/stop from a different thread? Is reading or writing to registers thread-safe?

aquynh:
Unicorn is thread-safe by design so yes you can stop emulation from other thread. if you find something wrong in implementation, please open an issue with details on how to reproduce it. it is best to do this with a pull request to put the sample code under tests/ directory.

xorstream:
The problem appears to be due to the mutex implementation in "unicorn\qemu\util\qemu-thread-win32.c". It uses critical sections for it's mutexes, however critical sections can only be used within a single thread. So when you try to do stuff from other (multiple) threads it fails.

The fix is to change it from using critical sections to actual mutexes/semaphores. I can make this change, but atm I am battling against git hehe. Once I work out how to bend git to my will I will make some fixes for it.

univm:
[...]

PR #405, opened 2016-01-25
Fix for calling uc_open() and uc_emu_start() from different threads

xorstream:
This is a fix attempt for calling uc_open() and uc_emu_start() from different threads as explained in issue #292 and regression test added in #320.

I have done various debug prints while trying fixes for this and it appears like this will be ok but perhaps someone with more experience with qemu internals can double check. Feel free to suggest other ways of doing this.

### DIFF
--- a/include/uc_priv.h
+++ b/include/uc_priv.h
@@ -88,6 +88,7 @@ struct hook_struct {
 struct uc_struct {
     uc_arch arch;
     uc_mode mode;
+    bool lock_at_vm_start;
     QemuMutex qemu_global_mutex; // qemu/cpus.c
     QemuCond qemu_cpu_cond; // qemu/cpus.c
     QemuThread *tcg_cpu_thread; // qemu/cpus.c
--- a/qemu/cpus.c
+++ b/qemu/cpus.c
@@ -42,6 +42,11 @@ static void *qemu_tcg_cpu_thread_fn(void *arg);
 
 int vm_start(struct uc_struct* uc)
 {
+    if (uc->lock_at_vm_start) {
+        uc->lock_at_vm_start = false;
+        qemu_mutex_lock_iothread(uc);
+    }
+    
     if (resume_all_vcpus(uc)) {
         return -1;
     }
--- a/qemu/vl.c
+++ b/qemu/vl.c
@@ -93,6 +93,7 @@ int machine_initialize(struct uc_struct *uc)
 {
     MachineClass *machine_class;
     MachineState *current_machine;
+    int ret;
 
     module_call_init(uc, MODULE_INIT_QOM);
     register_types_object(uc);
@@ -127,7 +128,11 @@ int machine_initialize(struct uc_struct *uc)
 
     current_machine->cpu_model = NULL;
 
-    return machine_class->init(uc, current_machine);
+    ret = machine_class->init(uc, current_machine);
+    
+    qemu_mutex_unlock_iothread(uc);
+    uc->lock_at_vm_start = true;
+    return ret;
 }
 
 void qemu_system_reset_request(struct uc_struct* uc)


aquynh:
a quick look, but this is strange to me: i suppose you have to remove some code, but you just added some new code. how does this solve the problem, if the same troubled code is still there?

xorstream:
The problem stems from the fact that uc_open() locks the mutex "qemu_global_mutex" and uc_emu_start() then unlocks the same mutex "qemu_global_mutex". These mutexes must be locked an unlocked by the same thread. So if you call uc_open() from a different thread to uc_emu_start(), then the lock will occur ok, but then you will get an assert error in uc_emu_start() when you try to unlock it.

The callstack for the lock is:
uc_open()
	machine_initialize()	// vl.c
		qemu_mutex_lock_iothread(uc);
			qemu_mutex_lock(&uc->qemu_global_mutex);

The callstack for the unlock is:
uc_emu_start()				// uc.c
	vm_start()				// cpus.c
		qemu_mutex_unlock_iothread()	// main-loop.c
			qemu_mutex_unlock(&uc->qemu_global_mutex);

When I added logs to various functions to see when they are called I didn't notice any "qemu_global_mutex" action between the uc_open() and uc_emu_start() calls. So as a fix I unlock the mutex again at the end of machine_initialize() so that it is not locked when uc_open() returns. I then add an extra lock at the start of vm_start() so that it will be then locked again when it does processing that appears to make use of it. That is the reason for the code additions.

xorstream:
Here is a debug print log I have, I'll try and explain it a bit to help make sense of it.
The important parts are mutex usage in and around the calls to uc_open() and uc_emu_start().
The first number in most prints of the address of the mutex. So 176A0058 is "qemu_global_mutex".
"thid" is the thread id of the caller.

Also mutex lock and unlock functions have a "start" and "end" postfix showing when the function is started and ended since if a lock is not ready it will stop executing until it is ready. So a "lock start" doesn't mean the lock occured, you need to find a "lock end" to know that the lock occurred.

Where you see function names with a number postfix, this means it was getting through parts of the function. So you can see by "machine_initialize 4" that no more qemu_global_mutex action occurred after machine_initialize() finished. And it is not until uc_emu_start() that qemu_global_mutex is then used again.


uc_open()
### uc->qemu_global_mutex
qemu_mutex_init:   176A0058 (thid=1178)
machine_initialize 1

-- main thread : machine_initialize() calls qemu_mutex_lock_iothread()
qemu_mutex_trylock:176A0058 (thid=1178)  start
qemu_mutex_trylock:176A0058 (thid=1178)  end
machine_initialize 2
machine_initialize 3
create thread CPU 0/TCG
qemu_cond_wait 0   176A0058 176A0074
qemu_mutex_unlock: 176A0058 (thid=1178)  start
qemu_mutex_unlock: 176A0058 (thid=1178)  end
qemu_cond_wait 1   176A0058 176A0074
qemu_tcg_cpu_thread_fn 0
qemu_tcg_cpu_thread_fn 1

-- cpu thread : qemu_tcg_cpu_thread_fn() calls qemu_mutex_lock()
qemu_mutex_lock:   176A0058 (thid=2CA8)  start
qemu_mutex_lock:   176A0058 (thid=2CA8)  end
qemu_tcg_cpu_thread_fn 2
qemu_tcg_cpu_thread_fn 3
qemu_cond_wait 2   176A0058 176A0074
qemu_cond_wait 3   176A0058 176A0074
qemu_mutex_lock:   176A0058 (thid=1178)  start		qemu_cond_wait() lock req
qemu_tcg_cpu_thread_fn 4
qemu_tcg_cpu_thread_fn 4.1
qemu_cond_wait 0   176A0058 17725B78
qemu_mutex_unlock: 176A0058 (thid=2CA8)  start
qemu_mutex_unlock: 176A0058 (thid=2CA8)  end
qemu_cond_wait 1   176A0058 17725B78

qemu_mutex_lock:   176A0058 (thid=1178)  end		qemu_cond_wait() lock get
qemu_cond_wait 4   176A0058 176A0074
machine_initialize 4

uc_mem_map()
uc_mem_write()
uc_hook_add()
---- Thread Starting ----
uc_emu_start()
qemu_cond_broadcast         17725B78
qemu_cond_wait 2   176A0058 17725B78
qemu_cond_wait 3   176A0058 17725B78
qemu_mutex_lock:   176A0058 (thid=2CA8)  start
qemu_mutex_unlock: 176A0058 (thid=2338)  start		-- my external thread (uc_emu_start() caller)
Assertion failed!

Program: F:\GitHub\unicorn\tests\regress\threaded_emu_start.exe
util/qemu-thread-win32.c, Line 76

lunixbochs:
discuss in #292 - I'm inclined to think this PR shouldn't be necessary anymore

aquynh:
Solved, thanks

PR #411, opened 2016-01-31
Fix sample_network_auditing.py so it works with Python3

cforger:
- Added detect for Python 2/3 so the correct iteritems()/iter() is called, but without needing to install six
- Switch to 'in' which works in 2/3 instead of legacy '.has_key()'
- Renamed 'id' variable use (which is a built-in) to my_id.
- Small formatting changes to make it more PEP compliant.

### DIFF
--- a/bindings/python/sample_network_auditing.py
+++ b/bindings/python/sample_network_auditing.py
@@ -7,7 +7,14 @@ from unicorn import *
 from unicorn.x86_const import *
 import struct
 import uuid
-import random
+
+# Python 2/3 Compat without installing six
+DEAD_PYTHON = False
+import sys
+
+if sys.version_info[0] < 3:
+    DEAD_PYTHON = True
+    print("Python 2.x is dead. Start thinking about migrating to 3.x. https://wiki.python.org/moin/Python2orPython3")
 
 SIZE_REG = 4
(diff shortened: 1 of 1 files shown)


bindings/python/sample_network_auditing.py
```diff
@@ -7,7 +7,14 @@
...
 import uuid
-import random
+
+# Python 2/3 Compat without installing six
+DEAD_PYTHON = False
+import sys
+
+if sys.version_info[0] < 3:
+    DEAD_PYTHON = True
+    print("Python 2.x is dead. Start thinking about migrating to 3.x. https://wiki.python.org/moin/Python2orPython3")
```
aquynh: would you mind removing the above line, or commenting it out?

lunixbochs: I've spent more time reading this PR than any computer will ever waste in `iteritems()` when running this sample. Just switch to `items()` for Python 2.x and drop the DEAD_PYTHON checks.

cforger:
Yes, I can take out the comment. :)

I can also kill the check entirely, and just use items() but I wasn't sure how far back into Python history you wanted to go.

lunixbochs:
I'm pretty sure items long predates iteritems.

aquynh:
merged, thanks!

PR #413, opened 2016-02-02
added testcase for zero length bb events

eqv:


tests/unit/test_x86.c
```diff
@@ -729,6 +729,56 @@ static void test_x86_16(void **state)
 
 /******************************************************************************/
 
+
+/******************************************************************************/
+
+// callback for SYSCALL instruction (X86).
```
aquynh: is this a correct comment? i dont see anything about SYSCALL here.

eqv: No, sorry this is a copy and paste mistake.

tests/unit/test_x86.c
```diff
@@ -729,6 +729,56 @@ static void test_x86_16(void **state)
...
+    // map 2MB memory for this emulation
+    err = uc_mem_map(uc, address, 2 * 1024 * 1024, UC_PROT_ALL);
+    uc_assert_success(err);
+
+    // write machine code to be emulated to memory
+    err = uc_mem_write(uc, address, code, sizeof(code));
+    uc_assert_success(err);
+
+    // hook interrupts for syscall
+    err = uc_hook_add(uc, &trace1, UC_HOOK_BLOCK, hook_empty_bb, NULL);
```
aquynh: the recent change on API requires begin & end arguments. if you want to hook every block, then set begin > end, for example:

err = uc_hook_add(uc, &trace1, UC_HOOK_BLOCK, hook_empty_bb, NULL, (uint64_t)1, (uint64_t)0);

PR #415, opened 2016-02-02
Create arm_init_input_crash.py

McLovi9:
Maybe this is already fixed cause on Windows I get a crash but on Linux is OK.
"Emulate ARM code
ERROR: No memory available or memory not present (UC_ERR_NOMEM)"


aquynh:
merged, thanks!

on Mac OS X, here is the output of this testcase:

```
$ ./arm_init_input_crash.py 
Emulate ARM code
>>> Tracing basic block at 0xf0000000, block size = 0x8
>>> Tracing instruction at 0xf0000000, instruction size = 4
>>> Tracing instruction at 0xf0000004, instruction size = 4
>>> R0 = 0x37
>>> R1 = 0x3456
====================
Emulate THUMB code
>>> Tracing basic block at 0xf0000000, block size = 0x2
>>> Tracing instruction at 0xf0000000, instruction size = 2
>>> SP = 0x1228
```

PR #416, opened 2016-02-03
add support for setting gdtr, idtr, ldtr, and tr programatically

cseagle:


aquynh:
this adds to `uc_x86_reg` some new enums, such as `UC_X86_REG_GDTR_LIMIT`, `UC_X86_REG_GDTR_BASE`. however these enums are not really *registers*, so i think they should not belong to `uc_x86_reg`.

what do you think about this: we define `UC_X86_REG_GDTR`, then pass the *pointer* of GDTR structure to `uc_reg_read()` and `uc_reg_write()`? if so, we will need new struct for `unicorn/x86.h`, like this:

```
typedef struct uc_x86_gdt {
    uint32_t base;
    uint16_t limit;
} uc_x86_gdt;
```

does this look better?

thanks

aquynh:
if using struct like this, we can add other info like `attr` & `ss` to just one struct, without having to define 4 new enums for the cases like `TR`

cseagle:
qemu/target-i386/cpu.h already contains:

typedef struct SegmentCache {
    uint32_t selector;
    target_ulong base;
    uint32_t limit;
    uint32_t flags;
} SegmentCache;

which is the data type of idtr and gdtr within the cpu env.

I'd suggest exposing this, via a typedef if necessary, in unicorn/x86.h. I'm not sure what your preferred way of doing that is since x86.h doesn't currently include any other headers.

Like you my preference is to have only one new enum per register, but the link between unicorn/x86.h and qemu/target-i386/cpu.h does not seem to exist, at least not obviously.

aquynh:
Yes, we can just copy structs we need over to x86.h, without including that
Qemu header.

cseagle:
I feel like there is a problem here. The definition for SegmentCache from qemu/target-i386/cpu.h:

typedef struct SegmentCache {
    uint32_t selector;
    target_ulong base;
    uint32_t limit;
    uint32_t flags;
} SegmentCache;

The sizeof target_ulong depends on the macro TARGET_LONG_SIZE (from include/exec/cpu-defs.h) which in turn depends on TARGET_LONG_BITS which depends on TARGET_X86_64. Where is this defined? If TARGET_X86_64 is set based on the build platform then this seems like an error to me.

ie if I am building unicorn to run on x86_64 then is TARGET_X86_64 64? And if I build to run on x86_32 is it 32?

This is a problem because the base field is specific to the architecture being emulated, NOT the architecture the emulation is running on.

base needs to be 64 bits when emulating x86_64 and 32 bits when emulating x86_32.

aquynh:
yes, lets set `base` to `uint64_t` so it works in any cases.

cseagle:
base set to uint64_t and register constants updated to one constant per register, parameter to reg_write/reg_read needs to be an x86_mmr pointer

qemu/target-i386/cpu.h
```diff
@@ -699,7 +699,7 @@ typedef enum {
 
 typedef struct SegmentCache {
     uint32_t selector;
-    target_ulong base;
+    uint64_t base;   /* handle 32 or 64 bit CPUs */
```
aquynh: oh, why do we need to fix this code from Qemu? it is better to leave this unchanged, while we still expose `base` to our API as `uint64_t` for our convenience?

include/unicorn/x86.h
```diff
@@ -8,6 +8,15 @@
...
 
+//Memory-Management Register fields (idtr, gdtr, ldtr, tr)
+//borrow from SegmentCache in qemu/target-i386/cpu.h
+typedef struct x86_mmr {
+    uint32_t selector;
+    uint64_t base;   /* handle 32 or 64 bit CPUs */
+    uint32_t limit;
+    uint32_t flags;
+} x86_mmr;
+
```
aquynh: i can see that you are trying to use the same struct for all related instructions. however, IDTR & GDTR do not have `limit` and `flags`, so this is confused. should we have separate structs for them?

aquynh: these are all called "segment" registers in Intel document, so should we rename this struct from "x86_mmr" to something like "uc_x86_segment"?

aquynh: in case we just want 1 struct for everything, then we need to document that IDTR & GDTR do not use "selector" & "flags" fields.

qemu/target-i386/cpu.h
cseagle: Is it guaranteed that target_ulong is always 64 bits? If it is then why not use uint64_t? If it isn't, then we can't cast x86_mmr* to SegmentCache*.

Since the pointer that passed into the qemu code is to a struct defined in x86.h (x86_mmr) we can't cast to x86_mmr* because that type is unknown to qemu. If we cast to SegmentCache* (as is done now) then there needs to be a guarantee that the base field is the same size as the corresponding base field provided by the caller.

include/unicorn/x86.h
cseagle: No they are not segment registers in intel documents. They are Memory Management Registers (see volume 3, section 2.4)

CS, DS, SS, ES, FS and GS are segment registers.

cseagle: Internal to qemu, cpu->gdt and cpu->ldt are declared as SegmentCache (see qemu/target-i386/cpu.h line 861). Unless you want to introduce a new data type in the qemu code as well, then within the qemu code we are obligated to cast the caller's pointer to SegmentCache, so our data type needs to be identical to SegmentCache.

Using these registers is something for advanced users. The caller will need to understand what is and is not valid.

In the supplied change to qemu/target-i386/unicorn.c flags and selector are ignored when reading or writing GDTR and IDTR.

qemu/target-i386/cpu.h
aquynh: no, target_long is 32bit for 32bit machine, and 64bit for 64bit machine. this is how Qemu handles long data type for different kind of guests.

we can't cast SegmentCache to our struct, but we can just copy each field like what you are doing now, can't we?

cseagle: 1. When I built unicorn on a 32 bit machine today, target_ulong ended up being 64 bits. I will restate my question from earlier, what determines the value of TARGET_LONG_BITS during the build process.

2. When you say "no, target_long is 32bit for 32bit machine, and 64bit for 64bit machine" do you mean host or guest? If target_ulong is 32 bits on a 32 bit host, then how do I emulate x86_64 on a 32 bit host since 64 bit register values will not fit into target_ulong.

3. We can't copy any fields from a void* argument to uc_reg_write unless we can cast that pointer to a known data structure type. This is why in my original pull request I had the registers broken out into UC_X86_REG_GDTR_LIMIT, UC_X86_REG_GDTR_BASE, etc

aquynh: on (3), i mean our struct in `x86.h` can just set "base" to *uint64_t*, but we leave the internal the struct "SegmentCache" of Qemu *unchanged*. then we can still copy the fields regardless how *target_ulong* is interpreted

cseagle: That will work if we copy the definition of x86_mmr from x86.h into cpu.h or unicorn.c (your choice) so that we can cast appropriately when we receive the void pointer in uc_reg_read/write

I am fine with that if that is what you want to do.

aquynh: why do we need to copy `x86_mmr` from `include/unicorn/x86.h` to somewhere else? we can just make sure `unicorn.c` can reference this struct in `x86.h`, can't we?

btw, for naming consistency, please stick "uc_" in front of `x86_mmr` struct.

cseagle: Yes of course we can include that, but it would be the only instance (though I did not look too hard) I can find of a file under qemu including a file from include/unicorn.

I assumed, apparently wrongly, that one of your design principles was to keep things under qemu independent of things outside of qemu. If that is not the case, then this is not a problem.

aquynh: yes, some parts of code under qemu/ need access to Unicorn struct. but in general, we try to keep the change to Qemu code minimal, and only do if we cannot avoid that.

aquynh:
merged, thanks!

now please add the new struct `uc_x86_mmr` to all the bindings.

PR #419, opened 2016-02-04
added memory fuzzer and 2 resulting testcases

eqv:
Added a simple memory fuzzer and 2 resulting minized testcases that are currently failing.


aquynh:
nice work, merged, thanks!

PR #422, opened 2016-02-05
X86 seg regs

cseagle:
read/write of x86 segment registers should modify selector field not base field. selectors are always 16 bits.


qemu/target-i386/unicorn.c
```diff
@@ -260,22 +260,22 @@ int x86_reg_read(struct uc_struct *uc, unsigned int regid, void *value)
                     *(int16_t *)value = READ_WORD(X86_CPU(uc, mycpu)->env.eip);
                     break;
                 case UC_X86_REG_CS:
-                    *(int32_t *)value = X86_CPU(uc, mycpu)->env.segs[R_CS].base;
+                    *(int32_t *)value = (uint16_t)X86_CPU(uc, mycpu)->env.segs[R_CS].selector;
```
aquynh: on the left side, don't you want to cast to pointer of **uint16_t** here, rather than int32_t?

cseagle: yes

aquynh:
merged, thanks!

ohjeongwook:
Then we should have new interface that actually can change base address for each segments. I don't see that the selector is actually used in translate.c for at least FS registers. You change FS register value to whatever value, it would end up with reference error with segment base of 0. If it is the cache problem is there any way to programmatically flush it out?

ohjeongwook:
OK, I just read other threads. So currently we don't have any interface out that can directly manipulate FS. I'm working on a code that only needs some fake data from FS registers. I thought just faking base and length and it would be just enough for me. I see some public example code are broken - especially shellcode emulators, etc.

lunixbochs:
If you're on 64-bit, you currently need to WRMSR (opcode `0x0F, 0x30`) with FSMSR (`0xC0000100`) and GSMSR (`0xC0000101`) to write a 64-bit base.

https://github.com/lunixbochs/usercorn/blob/master/go/arch/x86_64/linux.go#L24
https://github.com/lunixbochs/usercorn/blob/master/go/arch/x86_64/arch.go#L47

ohjeongwook:
Exactly, to just support FS, fiddling with low level descriptors don't sound good. The original wrong implementation of FS register setting was actually working well, except it was semantically wrong. We should support FS_BASE, GS_BASE for simple emulation code that just relies on fake memory layout. Or put any descriptor magic inside simple interface.

cseagle:
There is an interface to set the FS register. It is a selector, you may set it. it is nothing more than a selector. What you want to do when you are talking about things like base and limit is manipulate the associated segment descriptor (x86) or the associated MSR (x64).

For 32-bit code, this requires a GDT and descriptor cache that is honored by translate. The descriptor cache is there. Emulations do not get a GDT by default because it is fundamentally just a block of data in the memory space.

Code that manipulates specifically the FS base or FS limit is not desired or useful for 32 bit code. Instead, use the existing code to setup a GDT and the descriptors that you want, then set FS to point at your descriptor.

I need to break down and do a GDT example/how to so that this stops coming up.

lunixbochs:
My point is that fiddling with an MSR for this is painful enough (as you need to trampoline the emulator and map+run shellcode) it at least makes sense to allow writing this MSR with reg_read/write for the 64-bit case.

ohjeongwook:
@cseagle Is that your assumption that setting GDT and setting selector to FS will work, or do you have a test code somewhere? I'm trying to make it work, but no luck yet - maybe I'm misunderstanding something here. I'm more of making my code (which is shellcode analyzer) and PyANA working for x86 environment.

cseagle:
That is the way that it "should" work. If it doesn't work, then the underlying problems in translate should be fixed so that it accounts for segment descriptors.

aquynh:
recently @sashs post a good example on setting up GDT and FS register. there are some minor issues there, however. i am waiting for him to send a PR to put that example under tests/regress/, then i will make a commit to fix this issue in the core.

@cseagle is absolutely right that that the way to go is to emulate the real hardware as accurately as possible.

sashs:
Tomorrow I will send the PR. :)

Am 21.03.16 um 23:12 schrieb Nguyen Anh Quynh:

cseagle:
I think there should be an x86_write_msr interface that lets you write to any MSR you want. That is how the processor works. I have no idea whether qemu does anything with MSRs or not. If not, then special logic should be added to trap writes to the associated MSRs ans update the segment descriptor cache accordingly.

aquynh:
yes, Qemu supports MSR.

somebody should have a testcase for MSR first, then we will proceed from there.

ohjeongwook:
@aquynh I agree that we stick to the way more close to what HW are actually doing. But, usability is the issue here. Many people are simply confused. Even the example code from PyANA is confused, it works well with October release, but breaks with current git. I expect we fix this issue once and for all and clear way. We should not see people confused anymore. So, that is why I suggested an API or virtual reg value *_BASE exposed. If you guys are uncomfortable with fake register value, at least we should expose and api that can set FS base using whatever method dependent on the platform.

ohjeongwook:
@sashs Is this the code we are talking here? https://gist.github.com/sashs/8dad80b0fbe20b4e0e98 or there is another. It looks similar to what I tested, but I can test again if you can point it out.

aquynh:
That sample code of @sashs is good, but will not work until i push my local
commit.

cseagle:
I'm not trying to be a jerk here, but I think many people are confused because they simply don't understand the lower level details of the x86 architecture. If the goal of the Unicorn project is to provide a faithful emulation of the x86 hardware, then that is what we should do. Along the way we can also do our best to help people that wish to use the emulator to understand the tool that they are working with. 

There is a big difference between a user space emulator, that might hide some of the ugly details of kernel space, including interrupt handling, descriptor tables and page tables, and a full blown CPU emulator that allows emulation of the deepest details of the architecture being emulated.

it might be nice if there was a user space setup shim that created a sane basic setup for those that simply want to emulate user space code. Such a shim might give you a standard GDT setup to mimic Linux or Windows environments for example.

ohjeongwook:
The reason why many people use unicorn is it's usability and accessibility. Making it harder to use will not help it. Look at how many people are interested in writing shellcode analyzer here. That is one of the hot use case of unicorn. At least I don't expect it to emulate full OS stack. I want to use it for relatively short code to get quick idea on what it is doing.

Also confusion about FS register simply comes from lack of documents and sample code. The sample code I could find was treating FS register as FS base not selector. People just assume that is what it is supposed to be, and it was like it until this Feb. 

The unicorn engine is already placing a lot of hooks and shortcut methods for manipulating state of cpu and memory out of context. I still don't get it why adding new method for playing with register base is so much of a problem. Nobody objects on internal implementation of managing descriptors or MSR, just open up a simple api so we can just call it from our scripts. Just pass the register and base, and you get what you want. Is that too much?

aquynh:
The target is to emulate accurately, but still keep it very easy to use. We
can add some special "API" for OS-specific cases like this to help setup
environment.

ohjeongwook:
Thanks @aquynh. Just let me know if I can help with anything.

aquynh:
we can add a new option for `uc_option()`, for example `UC_OPT_WINDOWS_TIB`, so you can just make a single call like this to setup Windows TIB:

    uc_option(uc, UC_OPT_WINDOWS_TIB, base_addr_of_TIB);

does this make you happy?

ohjeongwook:
It looks good.

lunixbochs:
Naming it after Windows feels weird, as Linux and friends also want easy functions to set TLS registers (and using the exact same architecture tools as a Windows option would use).

aquynh:
i think the case of Windows is special in which it just needs to setup FS segment for TIB, but Linux also needs to setup other segments.

if you want more flexibility, then we have to pass through register id (FS, GS, etc) besides the base address, making it a bit more complicated.

aquynh:
i just created a new branch "tib" for this problem. see the commit https://github.com/unicorn-engine/unicorn/commit/66526d775b3d407fecee804d4e6487d13f13037a:

- add new API `uc_option()` (which will also be used for many other options in the future)
- add new option `UC_OPT_WINDOWS_TIB` to setup TIB.
- update Python binding to support above additions. so you can setup TIB with just a single call:

        uc.windows_tib = base_address

(be sure to map memory at `base_address` for this to work)

let me know if there is anything wrong. if not, i will merge this to the master branch.

thanks.

ohjeongwook:
I tested and got some runtime errors.

Comments here: https://github.com/unicorn-engine/unicorn/commit/66526d775b3d407fecee804d4e6487d13f13037a

After that, everything works perfect for me.

aquynh:
fixed, thanks.

aquynh:
i pushed this commit to support TIB for Windows 64: https://github.com/unicorn-engine/unicorn/commit/03aed19145d1e9966871db2c274879e86f050d0c

currently we only set the `base` for TIB segment, but ignore its `limit`.
what should be the value of `limit`?

JonathonReinhart:
> - add new API uc_option() (which will also be used for many other options in the future)
> - add new option UC_OPT_WINDOWS_TIB to setup TIB.

I thoroughly oppose this.

Putting x86-specific things in the core Unicorn API is a bad idea.

Putting x86, Windows-specific things in the core Unicorn API is a horrible idea.

Shoving x86, Windows-specific things into a generic `uc_option` interface is one of the worst design decisions I could imagine for this project. This is exactly the kind of API bastardization I warned of when I pushed back against the whole `uc_option` API during the beta.

> I'm not trying to be a jerk here, but I think many people are confused because they simply don't understand the lower level details of the x86 architecture. If the goal of the Unicorn project is to provide a faithful emulation of the x86 hardware, then that is what we should do. Along the way we can also do our best to help people that wish to use the emulator to understand the tool that they are working with.

 :+1: This. I feel like Unicorn is falling prey to the "try to implement whatever the users ask for immediately" problem that comes along when many open source projects start to mature. Trying to quickly satisfy every request that comes in without taking the time to thoroughly design the APIs creates an un-maintainable mess that no one wants to work on.

> it might be nice if there was a user space setup shim that created a sane basic setup for those that simply want to emulate user space code. Such a shim might give you a standard GDT setup to mimic Linux or Windows environments for example.

I agree with @cseagle that a set of user-mode setup helper functions, separate from the Unicorn core, which help to set up userspace for a Unicorn emulation session would be a much better design.

ohjeongwook:
"I agree with @cseagle that a set of user-mode setup helper functions, separate from the Unicorn core, which help to set up userspace for a Unicorn emulation session would be a much better design." 

Does this mean new code base or separate folder with x86/64 utilities? From the user's point of view, if something they can pick up easily is provided, it doesn't really matter what design it is using. I was just elaborating making unicorn users repeat same thing over again over again with their code is not so convenient.

cseagle:
A utilities hierarchy might useful, or better, a contrib hierarchy. It might contain things like windows or linux memory setups, file loaders to map executables into emulated memory space, or collections of stub functions for emulating common system calls and library functions

sashs:
I am developing something like that in python. So if this is wanted,
tell me and I write it in C as well.

aquynh:
@JonathonReinhart, i am not falling into the trap of "implement whatever users ask". the reason i did this is this is really a huge issue, and should be a priority to be solved for now.

can you proposal your API solution so users can easily setup OS/Arch-specific context, then we will discuss over that?

thanks

ohjeongwook:
@sashs Can you share when you have some progress? Are you trying to put them in unicorn project or separate one?

zachriggle:
While we're waiting for agreement on a "simple" interface to perform these actions, can anybody point to a PoC / example which sets the FS base and works with Unicorn today?  I don't mind going the long way to set up LDT/GDT entries.

sashs:
@ohjeongwook I am writing a reverse engineering framework which uses unicorn. More or less I am implementing there what is requested here.
But if it is planned to integrate that in unicorn, I will stop implemting that separatly and help implementing that for unicorn.

@zachriggle We are waiting that the PR #478 is merged into unicorn. Then my example should work (https://gist.github.com/sashs/8dad80b0fbe20b4e0e98).

aquynh:
I am attending a conference right now. Will sit down and look at these PR
as soon as i can, thanks.

PR #424, opened 2016-02-05
Added query func python bind

McLovi9:


bindings/python/unicorn/unicorn.py
```diff
@@ -262,6 +263,14 @@ def mem_protect(self, address, size, perms=UC_PROT_ALL):
         if status != UC_ERR_OK:
             raise UcError(status)
 
+    # return CPU mode at runtime
+    def query(self, query_mode=UC_QUERY_MODE):
```
aquynh: please dont set default value for query_mode here, because this is supposed to be used for other type of queries in the future.

aquynh:
there is no need to close this PR to open a new one. next time you just need to push new commits to your branch, then they will show up here automatically. thanks.

McLovi9:
OK thanks, I will remember next time, I'm new to git.

aquynh:
no worry, now you already know everything you need to know about Git. it is that simple :-)

egberts:
Of course, when aquynh said 

> next time you just need to push new commits to your branch, then they will show up here automatically

, don't forget to insert the "#424" into the commit notes during commit

McLovi9:
OK thank you very much!

PR #428, opened 2016-02-07
Bindings

cseagle:
Add support for read/write of x86 mmr. Deprecate original reg_read/write API, introduce new reg_read/write API that supports receiving/returning Objects as values


aquynh:
merged, thanks!

PR #431, opened 2016-02-11
two more testcases

eqv:


aquynh:
merged, thanks!

aquynh:
on OSX, these 2 testcases never have any problems. what is your environment?

eqv:
Ubuntu, the problem seems to be a memcpy with overlapping ranges which is undefined behviour. It's quite likely that your implementation of memcpy does something different from the one on my system. However if you run the testcase in valgrind you should be able to see the overlapping memcpy no matter how it is implemented.

lunixbochs:
Ooh, I had some problems with inconsistent crashes on Linux I haven't looked into yet. This might be related.

PR #434, opened 2016-02-17
Testcase to set ZF and modify eflags in x86/x86_64

emdel:
This testcase should set ZF and make RDX contain 'babe'.
For some reason, the emulator ignores my modifications.
We have the same behavior if we set the global variable zf to 0.
By playing with this global variable we should be able to jump in 
the two different branches. 
Surprisingly, the eflags register is always consistent with the emulation 
and thus RDX at the end always contains 'c0ca'
I'm not familiar with Qemu code, but I noticed the difference between 
env->eflags and env->eflags0.

I hope I'm not missing something trivial, and that my code makes sense :)

Any hint is really appreciated.

### DIFF
--- a/tests/regress/jumping.py
+++ b/tests/regress/jumping.py
@@ -1,0 +1,167 @@
+#!/usr/bin/env python
+# Mariano Graziano
+
+from unicorn import *
+from unicorn.x86_const import *
+
+import regress
+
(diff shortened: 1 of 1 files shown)


aquynh:
merged, thanks!

PR #440, opened 2016-02-22
test case: x86 guest paging

Ramirez57:
Test case for x86 paging using virtual addresses mapped by Unicorn, as well as unmapped.

Attempting to read/write from virtual address ranges unmapped by Unicorn wrongly causes protection faults, even when the virtual address points to read/write regions of Unicorn memory.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,356 +1,41 @@
-# Unicorn Emulator Engine
-# By Dang Hoang Vu <dang.hvu -at- gmail.com>, 2015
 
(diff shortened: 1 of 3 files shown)


aquynh:
why is the whole content of Makefile changed? if this is due to CR/LF, can you please fix that?

thanks

JonathonReinhart:
@Ramirez57 FYI: You don't need to close a PR and open a new one. You can just force-push to your branch, and this PR will pick up the new changes.

PR #441, opened 2016-02-22
test case: x86 guest paging

Ramirez57:
Test case for x86 paging using virtual addresses mapped by Unicorn, as well as unmapped.

Attempting to read/write from virtual address ranges unmapped by Unicorn wrongly causes protection faults, even when the virtual address points to read/write regions of Unicorn memory.

p.s. sorry for double pull


aquynh:
merged with a minor fix, thanks!

PR #447, opened 2016-02-29
Unicord fails to emulate VMOV on ARM (actually Thumb)

mkravchik:


aquynh:
great, but there are still some issues:

- do "chmod +x vmov.py" to make your script executable.

- this still needs assert to verify the correctness of this instruction. you may want to check for the new values of related registers. see other Python testcases under `tests/regress` for reference.

please fix these issues so i can merge it, thanks.

petevine:
Thanks @mkravchik, I'm going to use that name for my next project ;)

mkravchik:
What exactly you are going to use and for what project?

petevine:
Not sure yet, probably something about heart surgery or maybe musical instruments... Does that strike a cord? :)

PR #448, opened 2016-02-29
Fixed restoring of eflags after helper call

zaddach:
If the value of the eflags register is changed inside the uc_tracecode helper, the previously computed value from helpers is used nevertheless, and the change is not taken into account. This patch forces a reload of the condition temporary from the eflags register value after each helper call.


aquynh:
merged, thanks!

however, this is expensive to restore eflags after every instruction. a better approach is to let user to notify the core about this special change, and the engine checks for this after every instruction.

lunixbochs:
you "notify the core" by doing `uc_reg_write(UC_X86_REG_EFLAGS)`

aquynh:
that is smart :-)

PR #450, opened 2016-03-04
[query] add UC_QUERY_PAGE_SIZE uc_query helper

nplanel:
Return the current page size used by the current arch.
Useful to call uc_mem_map() with memory/size aligned.


aquynh:
merged, thanks!

PR #454, opened 2016-03-08
Fix FS and GS segments size

Spl3en:
The segments FS and GS sizes have been changed recently and shouldn't have been modified. Those segments are supposed to be able to contain far pointers, thus must be 64 bits sized so it can support UC_MODE_64.

Faulty commits : 
https://github.com/unicorn-engine/unicorn/commit/f3dc2522a090e1f8467f5a67571dc3e1bef49786
https://github.com/unicorn-engine/unicorn/commit/4cb43be5bf5a7dee090a6c0c337e1ae0573a679f

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -139,6 +139,7 @@ test_multihook
 test_pc_change
 mem_fuzz
 test_x86_soft_paging
+fs_gs_size
 
 
 #################
--- a/qemu/target-i386/unicorn.c
+++ b/qemu/target-i386/unicorn.c
@@ -156,10 +156,10 @@ int x86_reg_read(struct uc_struct *uc, unsigned int regid, void *value)
                     *(int16_t *)value = X86_CPU(uc, mycpu)->env.segs[R_DS].selector;
                     return 0;
                 case UC_X86_REG_FS:
-                    *(int16_t *)value = X86_CPU(uc, mycpu)->env.segs[R_FS].selector;
+                    *(int16_t *)value = X86_CPU(uc, mycpu)->env.segs[R_FS].base;
                     return 0;
                 case UC_X86_REG_GS:
-                    *(int16_t *)value = X86_CPU(uc, mycpu)->env.segs[R_GS].selector;
+                    *(int16_t *)value = X86_CPU(uc, mycpu)->env.segs[R_GS].base;
                     return 0;
             }
             // fall-thru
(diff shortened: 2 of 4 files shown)


JonathonReinhart:
This is wrong. You're the third person to suggest this. `FS` and `GS` are 16-bit **selectors** into the GDT.

Spl3en:
Okay, sorry for the misunderstanding and thanks for the explaination.

JonathonReinhart:
Since this seems to be a common misconception, feel free to contribute some documentation, and/or some "helper" code that sets up a GDT with given bases/lengths. I'm not sure that Unicorn has any code to do so right now.

Nukem9:
@JonathonReinhart http://www.felixcloutier.com/x86/WRFSBASE:WRGSBASE.html http://www.felixcloutier.com/x86/RDFSBASE:RDGSBASE.html

I don't know how these instructions work internally. Should just be aware of them though (since nothing makes sense on 64bit...)

aquynh:
no we have no example/sample code for this yet. i already asked in another issue if anybody has something to show, but no answer so far.

wondering if we can do something from the core, so this would works, by default, without much setup for users.

Spl3en:
@Nukem9 It should apply to unicorn : https://github.com/unicorn-engine/unicorn/blob/6986fa39475313e3b55294e99ca9439e397f8e40/include/unicorn/x86.h#L686

I just tried to execute those instructions from userland and had no problem reading FS/GS base.

cseagle:
I keep meaning to contribute a gdt setup example, but haven't had time to do so. The easy approach at emulation start would be to setup code and data descriptors to a base of zero and max limit, then initialize the segment registers appropriately. That would cover a large number of use cases.

aquynh:
this sounds good. as usual, PR is welcome :-)

lunixbochs:
I've been slowly working on Unicorn GDT support in Usercorn, but it's not quite done.

Someone definitely needs to write up an example for setting up GDT and FS/GS registers in both 32-bit and 64-bit to make this easier as we seem to have a lot of interest.

JonathonReinhart:
@Nukem9 @Spl3en Yes, you guys are correct that it's a bit different in 64-bit mode. My comments about GDT were specific to 32-bit protected mode. 

In 64-bit mode, everything is flat, wide-open, with the exception of FS and GS. See [x86-64 on OSDev Wiki](http://wiki.osdev.org/X86-64).  There are MSRs for setting these, as well as the `SWAPGS` instruction (for context switching in/out of kernel mode).

lunixbochs:
I'm implementing this with shellcode but it'd be nice to be able to set MSRs on x86 without trampolining the emulator and manually running shellcode.

egberts:
I wonder if we can have a sneak preview as to its proposed Unicorn GDT API would look like; you know, vetting for flexibility.

PR #457, opened 2016-03-09
X86 - Access to FPU and GDT and friends from python bindings

feliam:
Read write FPU state from python and fix GDT from py


aquynh:
this looks good. can you add a testcase or sample using FPU & GDT as well?

include/unicorn/x86.h
```diff
@@ -73,7 +73,8 @@ typedef enum uc_x86_reg {
 	UC_X86_REG_R9D, UC_X86_REG_R10D, UC_X86_REG_R11D, UC_X86_REG_R12D, UC_X86_REG_R13D,
 	UC_X86_REG_R14D, UC_X86_REG_R15D, UC_X86_REG_R8W, UC_X86_REG_R9W, UC_X86_REG_R10W,
 	UC_X86_REG_R11W, UC_X86_REG_R12W, UC_X86_REG_R13W, UC_X86_REG_R14W, UC_X86_REG_R15W,
-	UC_X86_REG_IDTR, UC_X86_REG_GDTR, UC_X86_REG_LDTR, UC_X86_REG_TR,
+	UC_X86_REG_IDTR, UC_X86_REG_GDTR, UC_X86_REG_LDTR, UC_X86_REG_TR, UC_X86_REG_FPCW,
+    UC_X86_REG_FPTAG,
```
aquynh: indentation here is not correct

qemu/target-i386/unicorn.c
```diff
@@ -140,6 +140,57 @@ int x86_reg_read(struct uc_struct *uc, unsigned int regid, void *value)
...
+        case UC_X86_REG_FP0 ... UC_X86_REG_FP7:
+            {
+                floatx80 reg = X86_CPU(uc, mycpu)->env.fpregs[regid - UC_X86_REG_FP0].d;
+                cpu_get_fp80(value, value+sizeof(uint64_t), reg);
+            }
+            break;
+        case UC_X86_REG_FPSW:
+            {
+                uint16_t fpus = X86_CPU(uc, mycpu)->env.fpus;
+        		fpus  = fpus & ~0x3800;
```
aquynh: indentation here, too

bindings/python/unicorn/x86_const.py
```diff
@@ -248,7 +248,9 @@
 UC_X86_REG_GDTR = 243
 UC_X86_REG_LDTR = 244
 UC_X86_REG_TR = 245
-UC_X86_REG_ENDING = 246
+UC_X86_REG_FPTAGS = 246
+UC_X86_REG_FPCW = 247
```
aquynh: the order is incorrect here: FPCW=246, FPTAGS=247. if you should already detected this if you had a testcase for these new registers.

qemu/target-i386/unicorn.c
```diff
@@ -573,6 +624,38 @@ int x86_reg_write(struct uc_struct *uc, unsigned int regid, const void *value)
...
+            break;
+        case UC_X86_REG_FPSW:
+            {
+                uint16_t fpus = *(uint16_t*) value;
+                X86_CPU(uc, mycpu)->env.fpus = fpus & ~0x3800;
+                X86_CPU(uc, mycpu)->env.fpstt = (fpus >> 11) & 0x7;
+            }
+            break;
+        case UC_X86_REG_FPCW:
+            *(uint16_t*) value = X86_CPU(uc, mycpu)->env.fpuc;
```
aquynh: this is another bug here: you are reading, but not writing? this can be detected with a testcase/sample, too.

aquynh:
merged with some fixes for the above issues. still looking forward to testcases/samples for the added FPU registers, thanks!

PR #458, opened 2016-03-10
Updated MSVC bindings for new and changed functions.

xorstream:
Added uc_query(), uc_mem_map_ptr() and uc_mem_regions().
Updated uc_hook_add().


aquynh:
merged, thanks!

PR #467, opened 2016-03-18
Bugfix: uc_x86_mmr has no attribute limits

sashs:
uc_x86_mmr: Using of an attribute limits, but in Structure is named limit.


aquynh:
merged, thanks!

PR #469, opened 2016-03-20
Use base instead of selector for 32/64 segment regs - we might consid…

ohjeongwook:
…er making separate UC_X86_REG_*S_SELECTOR code for selector manipulation


JonathonReinhart:
Sorry, but this is wrong. This has been proposed several times before (#454). See #422 for an explanation.

@aquynh This needs to be put in a FAQ or something.

aquynh:
i am working on this. the target is solve this issue so that a FAQ is not needed anymore.

cseagle:
#478 may address some of this

aquynh:
this is addressed now, thanks!

PR #475, opened 2016-03-22
GDT regress test

sashs:
GDT regress test based on my GDT setting sample

### DIFF
--- a/bindings/ruby/unicorn_gem/ext/unicorn.h
+++ b/bindings/ruby/unicorn_gem/ext/unicorn.h
@@ -26,7 +26,6 @@ VALUE m_uc_reg_write(VALUE self, VALUE reg_id, VALUE reg_value);
 VALUE m_uc_mem_read(VALUE self, VALUE address, VALUE size);
 VALUE m_uc_mem_write(VALUE self, VALUE address, VALUE bytes);
 VALUE m_uc_mem_map(int argc, VALUE* argv, VALUE self);
-//VALUE m_uc_mem_map_ptr(VALUE self, VALUE address, VALUE size, VALUE perms, VALUE ptr);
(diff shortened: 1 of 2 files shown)


cseagle:
Is this test supposed to pass or fail? Is it demonstrating a current bug with segment registers? Just checking, because it fails for me and I may have a fix

lunixbochs:
I'm in email but I think it's supposed to pass and currently fails (assuming this is the sashs regress)

sashs:
Currently it should fails, because if that:

> recently @sashs post a good example on setting up GDT and FS register. there are some minor issues  > there, however. i am waiting for him to send a PR to put that example under tests/regress/, then i will     > make a commit to fix this issue in the core.

cseagle:
I address this bug in #478 with that PR, sashs' gdt regress test works, also posted sample C code that sets up a GDT and along with an FS segment

aquynh:
merged, thanks!

sashs:
@cseagle Did you test if my regress test works? After @aquynh merged the PR I tested it today, but for me it doesn't work.

cseagle:
Your regress test works for me when combined with my PR #478 which hasn't been merged yet. It doesn't work for me on current unicorn-master

sashs:
Oh, I thought your PR is merged. Then it is clear why it doesn't work.

sparktrend:
Hi Guys,

i truly appreciate your effort on improving unicorn. I have a Windows emulator that emulates Windows executable files, which in almost all cases access FS register. My emulator started failing due to the GDT setup requirements introduced in the recent unicorn build. Based on the conversations in this thread, it seems GDT is not fully working. Could you take a look at the following issue and comment it?

https://github.com/unicorn-engine/unicorn/issues/522

Much appreciated,
Sean

PR #478, opened 2016-03-23
fix x86 segment setup by updating cached segment registers on reg_write

cseagle:
Example gdt and segment register setup for 32-bit x86. Update segment cache when using x86_reg_write to modify a segment register


ohjeongwook:
This looks awesome. I'll test this over the weekend. Do you think we can work on shim layer soon that can put all these utility stuff?

samples/sample_x86_32_gdt_and_seg_regs.c
```diff
@@ -0,0 +1,290 @@
...
+#include <inttypes.h>
+#include <assert.h>
+#include <string.h>
+#include <stdlib.h>
+#include <unistd.h>
+
+struct SegmentDescriptor {
+   union {
+      struct {   
+# if __BYTE_ORDER == __LITTLE_ENDIAN
```
aquynh: i fail to understand why endianess is important here. the library & the samples should run on the sample hardware architecture, no?

aquynh:
i merged this into the "seg_regs" branch, and will look closer before merging it to master branch. meanwhile, please do your future requests based on that branch, thanks!

samples/sample_x86_32_gdt_and_seg_regs.c
cseagle: It matters because if I have built unicorn to run on a big endian machine, say an old ppc mac, the fields need to be packed in the descriptor in the proper order for the x86 emulator to pull them back out of memory. I don't have big endian hardware to test on, but if it truly doesn't matter, feel free to trim that part out.

aquynh: sorry but this looks pretty ugly. it seems this is another good reason why we should put the code dealing with architecture internals back into the core, so users do not need to think about endianess like this.

aquynh:
the reason i have not merged this PR yet is that you are enabling some user-only code to use for system emulator. this doesnt look right to me.

cseagle:
I am not really sure what you are saying. The code that you are complaining about is sample code and has nothing to do with the behavior of the unicorn core. It demonstrates how to setup and use a GDT if you wish to do so. People have been screaming for such an example and now you want to reject it.

The primary purpose for the PR is to fix what happens when uc_reg_write is called for x86 segment registers and has absolutely nothing to do with the sample code other than enable it to work properly.

aquynh:
no you misunderstood me. i dont reject it, and do not complain. i just try to understand if it is possible to use the user-only code for system emulator.

the function `cpu_x86_load_seg()` is for user-emulator of Qemu, and they have a good reason not to use that for system emulator, which Unicorn is based on.

cseagle:
Ah, thanks for the clarification. I have actually been wondering the opposite thing. Why didn't you base Unicorn on QEMU User in the first place since you cut out all of the hardware related code from QEMU it seems that what is left is essentially QEMU User. For that reason it looked like safe code to me since we have no hardware to deal with.

aquynh:
this is a good question. Qemu's user-emulator does not emulate the whole system, so while it would save me quite a lot of time when i started this project, i decided to base Unicorn on its system emulator instead. Unicorn is open for all kind of emulation, so i think it was a good decision.

aquynh:
by "this looks pretty ugly", and i did not mean your code. what i mean is that the fact the user code needs to care about endianess is ugly, and i think we should do better to avoid this. one solution is to move stuff like this into the core, like we can do with that option UC_OPT_WINDOWS_TIB at https://github.com/unicorn-engine/unicorn/pull/422

cseagle:
I'd rather see the struct layout made private in some contributed descriptor/gdt manipulation code, that includes a set of accessor functions that allow the user to be ignorant of the inner details of the descriptor layout. This type of things is an unnecessary option for the core, though it is more processor relevant than offering Windows specific options in the core (which I don't think is very elegant)

aquynh:
merged, thanks!

PR #481, opened 2016-03-24
execute cpus in same thread as uc_emu_start()

lunixbochs:
There's no change to samples_all.sh output, and all regression tests looked to pass the same.

note: I'm pretty sure this makes some dead code, but I wanted the least intrusive patch possible

### DIFF
--- a/qemu/cpus.c
+++ b/qemu/cpus.c
@@ -38,17 +38,13 @@ static void cpu_handle_guest_debug(CPUState *cpu);
 static int tcg_cpu_exec(struct uc_struct *uc, CPUArchState *env);
 static bool tcg_exec_all(struct uc_struct* uc);
 static int qemu_tcg_init_vcpu(CPUState *cpu);
-static void *qemu_tcg_cpu_thread_fn(void *arg);
+static void *qemu_tcg_cpu_loop(struct uc_struct *uc);
 
 int vm_start(struct uc_struct* uc)
 {
     if (resume_all_vcpus(uc)) {
         return -1;
     }
-
-    // kick off TCG thread
-    qemu_mutex_unlock_iothread(uc);
-
     return 0;
 }
 
@@ -99,7 +95,6 @@ int resume_all_vcpus(struct uc_struct *uc)
                 if (qemu_init_vcpu(cpu))
                     return -1;
             }
-            qemu_mutex_lock_iothread(uc);
         }
     }
 
@@ -107,6 +102,7 @@ int resume_all_vcpus(struct uc_struct *uc)
     CPU_FOREACH(cpu) {
         cpu_resume(cpu);
     }
+    qemu_tcg_cpu_loop(uc);
 
     return 0;
 }
@@ -125,13 +121,11 @@ int qemu_init_vcpu(CPUState *cpu)
 }
 
 
-static void *qemu_tcg_cpu_thread_fn(void *arg)
+static void *qemu_tcg_cpu_loop(struct uc_struct *uc)
 {
-    CPUState *cpu = arg;
-    struct uc_struct *uc = cpu->uc;
+    CPUState *cpu;
 
     //qemu_tcg_init_cpu_signals();
-    qemu_thread_get_self(uc, cpu->thread);
 
     qemu_mutex_lock(&uc->qemu_global_mutex);
     CPU_FOREACH(cpu) {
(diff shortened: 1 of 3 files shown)


cseagle:
Nice. I was also looking at converting the timer thread into a timer callback (CreateTimer on Windows) or alarm/itimer to get rid of that thread as well.

qemu/cpus.c
```diff
@@ -191,15 +169,10 @@ static int qemu_tcg_init_vcpu(CPUState *cpu)
         uc->tcg_halt_cond = cpu->halt_cond;
         snprintf(thread_name, VCPU_THREAD_NAME_SIZE, "CPU %d/TCG",
                 cpu->cpu_index);
-        if (qemu_thread_create(uc, cpu->thread, thread_name, qemu_tcg_cpu_thread_fn,
-                cpu, QEMU_THREAD_JOINABLE))
-            return -1;
+        qemu_thread_get_self(uc, cpu->thread);
```
aquynh: is this line necessary?

lunixbochs: I pulled it from `qemu_tcg_cpu_thread_fn` here https://github.com/unicorn-engine/unicorn/pull/481/files#diff-a4c1917cb81e3f2a58a18d86495f79b9L134

aquynh: unless i miss something, a less intrusive approach is to keep that line (call to `qemu_thread_get_self()`) in the old place, and simply replace `qemu_thread_create()` with `qemu_tcg_cpu_loop(uc)`.

does this work?

cseagle: It's not necessary as long as you pull the string and clean up all of the other places that might use it (which are also unnecessary)

lunixbochs: @aquynh that does not work as expected, due to the way the original code was laid out. At this point initialization is not done. My PR as submitted should even work with more than one CPU on a single Unicorn handle. If you put the CPU loop where the thread was, this flow happens:

1. init cpu
2. cpu loop
3. finish initing cpu

My submitted flow is:

1. init cpu
2. finish initing cpu
3. run cpu loop

In your suggested edit, if you have more than one CPU object on a uc handle for some reason, this happens:

1. init cpu
2. run cpu loop for all cpus
3. init cpu 2
4. run cpu loop for all cpus
5. finish initing cpus 1, 2

My flow is:

1. init cpu 1, 2
2. finish initing cpu 1, 2
3. run cpu loop for all cpus

Does that make sense?

cseagle: Which raises the question, is unicorn ever intended to emulate a multi-cpu system? Currently num_cpus and num_threads is hard coded to 1 each I believe.

lunixbochs: I want to be able to emulate multi-threaded code, but right now my best plan is to thread my host program and schedule `uc_emu_start()` threads manually.

I think Unicorn would do a much better job than me with cache coherency, as I have no idea what a memory fence/barrier/etc would do across two Unicorn instances with `mem_map_ptr`-shared memory.

cseagle: Yes, I think that once qemu_thread_create is remove, unicorn is locked into single cpu mode as there is no way that I see currently for cpus to yield to one another lacking threads. So there is a dilema: with threads, callbacks won't happen on the emu_start thread, and without threads, I think a single unicorn instance can't do multiple CPUs. Currently unicorn is hard coded for one cpu. Future uses of unicorn should be considered. My personal need is for a single CPU instance with a preference for callbacks on the emu_start thread.

lunixbochs: QEMU kinda supports multiple CPUs multiplexed onto a single thread, where it alternates calling the tcg exec functions https://github.com/lunixbochs/unicorn/blob/master/qemu/cpus.c#L200

Some of this might be lost in translation to Unicorn. It honestly might make the most sense to eventually remove all support for multiple threads and CPUs in Unicorn and expose any necessary sync primitives to the API, but I think we need a more clear use case and better idea of the upsides and downsides before someone works on it.

cseagle: I looked at that loop, which in turn calls tcg_cpu_exec -> cpu_exec which contains this infinite loop:

https://github.com/lunixbochs/unicorn/blob/master/qemu/cpu-exec.c#L100

which only seems to break on a stop_request or an exception/interrupt. If that is the case, any other CPUs would locked out until the current CPU hits a stop condition.

I agree that more thought needs to be put into a solution that supports multiple CPUs

zachriggle:
> If that is the case, any other CPUs would locked out until the current CPU hits a stop condition.

I'd rather have functional single-thread callbacks than multi-cpu functionality :smile:

lunixbochs:
@zachriggle I think the consensus on the comment thread here with @cseagle is that anything we'd break by removing multiple cpus in one unicorn instance is already broken. That's an aside and should honestly be opened as its own issue ("remove multi-cpu support and expose memory fencing / whatever as a sync api"), but doesn't block this PR at all.

PR #485, opened 2016-03-25
Nothreads

cseagle:
Extending lunixbochs' work to cut out all threading code, convert timeout code to use timers/callbacks


aquynh:
is this dependent on the PR of Ryan?

cseagle:
It builds on his PR. My PR contains all of his changes and more. Unfortunately this PR needs to be put on hold. I noticed that it is currently not building on the Travis-CI infrastructure for OS X. I need to find a way to do asynchronous timers on OS X, preferably without resorting to threads and polling.

aquynh:
sure, i will wait for your fix, and do not merge the PR of Ryan. thanks.

cseagle:
His PR is a good first step and doesn't break builds or timers I think. You could move ahead with that one I think.

lunixbochs:
@cseagle Why did you include my changes without my commit?

cseagle:
Because I was building on them and I don't know how to work git

lunixbochs:
OS X has `setitimer` but you can only register one timer for each type per process (or task?), which is kinda gross (as a Unicorn the library will now break anything else that tries to use a timer). Honestly the thread is cleaner there, it just breaks timeout if you fork (but timers aren't inherited either), and doubles the thread count.

Unicorn timeout is also opt-in. An application using Unicorn can choose to signal `uc_emu_stop()` asynchronously itself if there's a problem with Unicorn's method.

cseagle:
Yeah, I am actually going to implement a very crude asynch timer capability on top of the single itimer and see how that works out.

My concern is that if I want to have hundreds of unicorn instances in my process, that I need to have potentially hundreds of threads to support timeouts.

aquynh:
@cseagle, you can simply pull the branch you want to base on (Ryan's in this case), then work on it, so you still retain commits of other people. something like this (i use your branch as example):

    $ git checkout -b cseagle-seg_regs master
    $ git pull https://github.com/cseagle/unicorn.git seg_regs

then you work on this local branch, commit new changes to your local git repo, then push to your own branch on Github with:

    $ git push origin seg_regs

after that you can do PR from this "seg_regs" branch on the web interface.

lunixbochs:
I think if you want to run 200 Unicorn instances in your program, as you will already need to manage 200 threads for each `uc_emu_start()` call. At this point I think it's reasonable for a single extra thread in either Unicorn or your own program to manage the async timers.

`setitimer` is a weird global singleton and I don't think it makes any sense to use it in a library.

cseagle:
Excellent thanks. I will probably cancel this PR for the time begin because of the OS X timer stuff. When I sort that out (I really have a use case where I want no additional threads). then I will rebuild the PR using the process you describe.

lunixbochs:
If you don't want additional threads, I think you should manage the timers in your own code if you can, because my vote will always be against including a `setitimer` call inside Unicorn on OS X, as that can break a lot of code (and will break some profilers).

cseagle:
I think you need a single thread per timed unicorn as you need to provide a unique uc handle to each thread, so that doubles the number of threads to 400 in your example.

lunixbochs:
I was talking about threads in *your* code. You can't call `uc_emu_start()` more than once at a time without threading your own code.

cseagle:
No, if timers worked, unicorns can be paused and restarted based on timer expiration and my own selection logic for which unicorn to resume next. I suspect this can be done in as few as 2 threads with more worker threads added as desired, but no need for one thread per unicorn instance unless I decide I want that many workers.

lunixbochs:
That's still one thread per `uc_emu_start()`. I never said thread per instance.

cseagle:
Closing this request. Reworking to leave the timer thread stuff in because OS X has no async timers. Will resubmit when build works on all archs.

Issue #499, opened 2016-03-30
Both EFLAGS and single-stepping through conditionals is broken on amd64


zachriggle:
Feel free to dupe to #496 if you want, but here's a concrete trace.  **I did not contribute there since this is an actual emulation failure, rather than a failure to be coherent inside a callback.**

By using Binjitsu / Pwntools, you can easily assemble the following sequence of instructions:

```
$ asm -c amd64 'xor r14,r14; test r14d, r14d; jne $; hlt'
4d31f64585f675fef4
```

It's also pretty easy to launch it in a debugger.

```
$ asm -c amd64 'xor r14,r14; test r14d, r14d; jne $; hlt' --debug
```

When the debugger pops up, here's what we see:

```
pwndbg> x/4i $pc
=> 0x6000b0:    xor    r14,r14
   0x6000b3:    test   r14d,r14d
   0x6000b6:    jne    0x6000b6
   0x6000b8:    hlt    
```

By instantiating a Unicorn context object with the current state and stepping four times, we should hit the HLT instruction, but we never do.

Here's the full trace, below.  Note that everything between `emu_start` and `emu_stop` is processed in callbacks.

```
# Instantiating Unicorn for x86-64
uc = U.Uc(4, 8)
uc.reg_write(U.x86_const.UC_X86_REG_CS, 0x33)
uc.reg_write(U.x86_const.UC_X86_REG_SS, 0x2b)
uc.reg_write(U.x86_const.UC_X86_REG_DS, 0x0)
uc.reg_write(U.x86_const.UC_X86_REG_ES, 0x0)
uc.reg_write(U.x86_const.UC_X86_REG_FS, 0x0)
uc.reg_write(U.x86_const.UC_X86_REG_GS, 0x0)
Could not set register 'fsbase'
Could not set register 'gsbase'
uc.reg_write(U.x86_const.UC_X86_REG_AX, 0x0)
uc.reg_write(U.x86_const.UC_X86_REG_AH, 0x0)
uc.reg_write(U.x86_const.UC_X86_REG_AL, 0x0)
uc.reg_write(U.x86_const.UC_X86_REG_BX, 0x0)
uc.reg_write(U.x86_const.UC_X86_REG_BH, 0x0)
uc.reg_write(U.x86_const.UC_X86_REG_BL, 0x0)
uc.reg_write(U.x86_const.UC_X86_REG_CX, 0x0)
[...]

PR #501, opened 2016-03-31
Add test case for unicorn-engine/unicorn#499

zachriggle:
This currently fails because of #499

### DIFF
--- a/tests/regress/x86_64_eflags.py
+++ b/tests/regress/x86_64_eflags.py
@@ -1,0 +1,38 @@
+#!/usr/bin/python
+import regress
+import unicorn as U
+
+class WrongEFLAGS(regress.RegressTest):
+    def test_eflags(self):
+        # xor r14,r14
+        CODE = 'M1\xf6'
+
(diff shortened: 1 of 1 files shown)


aquynh:
merged, thanks!

aquynh:
this is not important, but why did you set RIP? that is not necessary, because later you start emulation, which will reset the RIP

aquynh:
so basically on physical machine the reserved flag (bit 1) is turned on, but Unicorn ignores this flag?

aquynh:
what is the value of EFLAGS with this XOR instruction in Qemu?

zachriggle:
The results within `qemu-user` are identical to the native results.  See https://github.com/unicorn-engine/unicorn/issues/499#issuecomment-204630258

zachriggle:
@aquynh: RIP is set because of how I initialize *all* registers in my emulation code from the actual machine state.  Easier to just set everything than have special cases to not-set-`$PC`-because-it'll-be-overwritten.

PR #503, opened 2016-04-01
Added stdint include to x86.h

adrianherrera:
include/unicorn/x86.h referenced types defined in stdint.h (e.g. uint16_t, etc.), but didn't actually include stdint.h

### DIFF
--- a/include/unicorn/x86.h
+++ b/include/unicorn/x86.h
@@ -8,6 +8,8 @@
 extern "C" {
 #endif
 
+#include <stdint.h>
+
 // Memory-Management Register for instructions IDTR, GDTR, LDTR, TR.
 // Borrow from SegmentCache in qemu/target-i386/cpu.h
 typedef struct uc_x86_mmr {


aquynh:
i see, but without this, what is wrong?

radare:
i dont think x86.h is suposed to be included directly

adrianherrera:
Hey guys,

Thanks for the comments. Sorry, I probably should have added more context to my PR. I'm currently working on some Haskell bindings (almost done!) that use the c2hs tool to automate a large amount of code generation by reading the header files. c2hs fails to generate code for x86.h because it does not know what a `uint16_t` is because stdint.h is missing.

It is probably important to add that because of `ifdef` guards this does not impact Unicorn in any way, shape or form. This is simply a robustness improvement.

Finally, if x86.h is x86.h is not supposed to be included directly, then it should not be exposed in the public include dir. Adding an stdint.h include in x86.h is essentially adding robustness (and, IMO, correctness). Actually, now that I've thought about it some more, should all of the CPU architectures be exposed via unicorn.h? If I'm only writing code that uses an X86/ARM/whatever emulator, then why do I need to see all of the others? It should really be up to the user to include the header files of the architectures that they actually need.

I hope I've convinced you enough to accept my PR! :-) I'm happy to answer any further questions.

aquynh:
does this mean you are working on a Haskell binding for Unicorn? do you plan to maintain it externally, or include it with Unicorn source like other bindings?

whatever your plan is, i would recommend to extend https://github.com/unicorn-engine/unicorn/blob/master/bindings/const_generator.py to support your binding. this script is very convenient and helpful to keep the bindings in sync with the core: whenever there is a minor change in the core (like adding or modifying some enums), i just do:

    cd bindings; make

and this script automatically updates all the xxx_const.xxx files in all bindings.

and you dont need to modify `x86.h` for this. what do you think?

adrianherrera:
Yes, I am working on a Haskell binding (about 90% of the way there!), and yes, I hope to submit a pull request and include it with Unicorn like the other bindings.

Ok cool, to be honest I hadn't looked at const_generator.py; I'll have a look and see if it solves my problem. If not, there is no harm in adding the stdint include to x86.h!

Regards,
Adrian

adrianherrera:
I had a look at `const_generator.py` and I will not need it for the Haskell bindings. Haskell already provides a tool, c2hs, that does code generation of constants, etc. based on the header files in `include/unicorn`.

By not including stdint.h in x86.h, x86.h is not self-contained and therefore is dependent on the ordering of include files to actually compile (Google "self-contained header file"). This is not good programming practice. By including stdint.h in x86.h, x86.h becomes self-contained and therefore avoids this problem. Additionally, it means that c2hs will operate correctly because it will know what an `uint16_t` is, etc.

aquynh:
On Apr 3, 2016 8:47 PM, "Adrian Herrera" <notifications@github.com> wrote:
>
> I had a look at const_generator.py and I will not need it for the Haskell
bindings. Haskell already provides a tool, c2hs, that does code generation
of constants, etc. based on the header files in include/unicorn.
>
> By not including stdint.h in x86.h, x86.h is not self-contained and
therefore is dependent on the ordering of include files to actually compile
(Google "self-contained header file"). This is not good programming
practice. By including stdint.h in x86.h, x86.h becomes self-contained and
therefore avoids this problem. Additionally, it means that c2hs will
operate correctly because it will know what an uint16_t is, etc.

Ok, can you send a new PR to handle all the rest <arch>.h files?

Thanks

adrianherrera:
The other <arch>.h files either (a) already include stdint.h (even though they don't actually use anything from stdint.h), or (b) don't need stdint.h because they just contain enum definitions and don't actually use stdint.h.

aquynh:
merged, thanks!

Issue #496, opened 2016-03-29
eflags issue on x86_64


jbremer:
Hi all,

Excellent framework. Found the opportunity to take advantage of it in the past few days, overall worked pretty well. Ran into one pretty awful bug, though. Namely that the `eflags` register is not correctly updated in certain cases. This is a Unicorn-only bug (as in, not in upstream Qemu).

First of all I modified a couple of things regarding the Python bindings of which I'd be happy to see merged upstream - if you're willing to take these changes (please let me know).
- Some code cleanup to improve code style (https://github.com/jbremer/unicorn/commit/b7d817d107790c0473abc787aa49d6145c40a356)
- Simplification of register access (x86-only currently). This allows one to read the `eax` register through `mu.eax`, write the `r8` register through `mu.r8 = 1234`, etc (https://github.com/jbremer/unicorn/commit/ecb8fd3b6160dd30b5b9eded84375d2a82bd5a97)
- Return the amount of executed instructions if desired; this functionality is currently not yet very well designed as it only works when providing a maximum number of instructions to be executed (i.e., `mu.emu_start(start, end, count=-1 % 2**32)` - mostly because that functionality almost directly results in exposing the amount of executed instructions (at the cost of inserting a `UC_HOOK_CODE` in the background, of course), see also https://github.com/jbremer/unicorn/commit/fbb82f5fa18ca521b7dc67faff21a670e9207622
- Basic register import/export (Python binding only); this is just a basic wrapper for reading and writing all of the general purpose registers at once https://github.com/jbremer/unicorn/commit/dafc95a2df1818dc3b0e707436d2ea8986812757 https://github.com/jbremer/unicorn/commit/a65c6c06e49ae6b320c7bb7adb1293de57a5d99e

This brings us to the actual bug. It can be identified through the following script https://gist.github.com/jbremer/6c0aa8253304f7f5b385. Simply by commenting the `hook_add` one can _make this script work_. Obviously commenting that line out removes all calls to the `dumper` callback and also - in the case of counting executed instructions - would get rid of that underlying functionality.

[...]

PR #505, opened 2016-04-04
remove call to restore_eflags (#496)

lunixbochs:
this outright breaks conditionals and needs to be done in a different way

### DIFF
--- a/qemu/target-i386/translate.c
+++ b/qemu/target-i386/translate.c
@@ -4784,7 +4784,6 @@ static target_ulong disas_insn(CPUX86State *env, DisasContext *s,
             changed_cc_op = true;
         }
         gen_uc_tracecode(tcg_ctx, 0xf1f1f1f1, UC_HOOK_CODE_IDX, env->uc, pc_start);
-        restore_eflags(s, tcg_ctx);
         // the callback might want to stop emulation immediately
         check_exit_request(tcg_ctx);
     }


aquynh:
does this fix the issues of https://gist.github.com/jbremer/6c0aa8253304f7f5b385 ? (reported in #496 by @jbremer)

lunixbochs:
yep, the 0xc7 case behaves correctly under this patch as described in the original issue

aquynh:
@zaddach, do you have any comments?

zachriggle:
This also fixes the issues I'm seeing in the emulator when single-stepping through conditionals #499, which is tested by #501 and #511.

The test in #501 passes after this change.

The test in #511 passes after this change.

jbremer:
Wouldn't it make sense to simply `git revert` 5fa6705d7af599275193d5eeaad5abf20fcc1fee? Or will this functionality be used later on again - is this a temporary workaround or permanent fix?

lunixbochs:
_(weird I swear I email replied to this)_

This fix is only intended to be temporary, as someone expressed interest in manually setting EFLAGS from hooks.

I think the correct solution might actually be to trampoline Unicorn stop/start when setting EFLAGS (assuming that works), and honestly it will have better performance than inserting a few extra instructions for every single hooked instruction anyway.

lunixbochs:
temp - modifying eflags is useful I hear.

different thought - isn't it likely way less expensive to stop/start trampoline unicorn to set eflags than to add code to every traced instruction? does tcg allow for that here?

zaddach:
I don't know too much about the eflags handling for x86, in ARM it's much
easier. Probably it's a better idea to exit the CPU loop and re-enter it
when EFLAGS registers have been changed. In this case you can simply revert
my earlier patch ...

aquynh:
i merged this, and commented out function restore_eflags() for now. we still need to find out how to sync EFLAGS for callbacks, which was why restore_eflags() was added.

@zachriggle: now the testcase #501 still fails here. are you sure it passes with this PR?

lunixbochs:
> we still need to find out how to sync EFLAGS for callbacks
does #512 not fix this?

PR #506, opened 2016-04-04
add batched reg access

lunixbochs:
add reg_read_batch and reg_write_batch API

this improves Usercorn's overall execution speed when doing a register trace by over 2x

related to #8

one of the reasons I want selective register access instead of all-at-once, is the syscall/function call ABI only requires a small subset of registers

### DIFF
--- a/bindings/go/unicorn/uc.c
+++ b/bindings/go/unicorn/uc.c
@@ -1,0 +1,23 @@
+#include <stdlib.h>
+#include <unicorn/unicorn.h>
+#include "_cgo_export.h"
+
+uc_err uc_reg_read_batch_helper(uc_engine *handle, int *regs, uint64_t *val_out, int count) {
+    void **val_ref = malloc(sizeof(void *) * count);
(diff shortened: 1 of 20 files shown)


aquynh:
apparently this still has some issues with Travis, would you take a look?

lunixbochs:
right, I added `for (int i = 0` reflexively: `‘for’ loop initial declarations are only allowed in C99 mode`

lunixbochs:
travis passes now

trufae:
didnt tested, but this thing will be great for r2 integration :)

aquynh:
this looks good! will you add some sample code (under samples/) to demonstrate these new APIs?

lunixbochs:
just pushed a sample

zachriggle:
Looks nice!  Can haz Python bindings? :D

aquynh:
merged, thanks!

@zachriggle, i leave that Python binding work to you :-)

Issue #508, opened 2016-04-04
Dynamic translator buffer issue


emdel:
Hi guys,
anyone experienced during the emulation of many instructions the message:

`Could not allocate dynamic translator buffer`

This message is from: qemu/translate-all.c:688
Unfortunately I'm not confident with the Qemu code so before digging into the 'softmmu' I ask 
for advices from the community.
By playing with the parameters of unicorn_code and stack I can reach different points.
From the code I understood I have to allocated at least one page and the granularity is at the page level, and I got always this message.

If it's necessary I can create a testcase.

Thanks a lot,

mariano

lunixbochs:
What's your host arch/OS? How much RAM do you have?

emdel:
Yep sorry. Host OS: Ubuntu; RAM: 8GB; arch: x86_64

lunixbochs:
Can you run under strace and post that?

aquynh:
@emdel, yes please send a testcase so we can test the fix.

emdel:
strace output (around the error):

```
stat("/home/emdel/projects/ropmemu/volatility/plugins/ropmemu/disass.py", {st_mode=S_IFREG|0664, st_size=1919, ...}) = 0
lseek(3, 62967808, SEEK_SET)            = 62967808
read(3, "\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"..., 4096) = 4096
lseek(3, 62976000, SEEK_SET)            = 62976000
read(3, "\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"..., 4096) = 4096
lseek(3, 62992384, SEEK_SET)            = 62992384
read(3, "\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"..., 4096) = 4096
lseek(3, 53821440, SEEK_SET)            = 53821440
read(3, "\376\377\377\17\267\22H\211KX\306CE\2\306CD\1\306C<\1\211S@\351\304\375\377\377f\220"..., 4096) = 4096
write(1, "\t | 0xffffffff8135469f \t| pop rd"..., 35     | 0xffffffff8135469f   | pop rdi 
) = 35
[...]

PR #510, opened 2016-04-05
test case for the dynmaic translator buffer issue

emdel:
Testcase for the issue described here: https://github.com/unicorn-engine/unicorn/issues/508


aquynh:
merged, thanks!

PR #511, opened 2016-04-05
Add test showing failure to track targets properly when single-stepping.

zachriggle:
This issue breaks emulation of conditional instructions on amd64.


zachriggle:
This is fixed by #505.

aquynh:
merged, thanks!

PR #514, opened 2016-04-06
Haskell bindings

adrianherrera:
These Haskell bindings make large use of c2hs to generate much of the Haskell code, so Unicorn's `const_generator.py` is not used.

The emulator is based on the Either monad transformer. The IO monad is used to run the underlying Unicorn library, while the Either monad is used to handle errors.

Instructions on how to build the bindings are located in bindings/haskell/README.TXT. The same samples found in samples/ can be found in bindings/haskell/samples. They should produce the same output, with slight differences in their error handling and messaging.


aquynh:
this is awesome, thanks for working on this!

a quick look, and an issue is that currently whenever the core has some minor changes on those enum, i simply do this to update for all the bindings under bindings/

    $ cd bindings
    $ make

this Haskell binding does not support that yet. can you modify `bindings/Makefile`, so it call your c2sh to do this?

thanks

adrianherrera:
Thanks for the comment! Changes have been made to the Makefile as requested.

aquynh:
i run "make" under bindings/ on Mac, and get below output. what do you think?

```
cd haskell && cabal configure && cabal build
Resolving dependencies...
Configuring unicorn-0.1.0.0...
cabal: At least the following dependencies are missing:
either >=4.4
make: *** [all] Error 1
```

adrianherrera:
Can you please tell me your version of ghc and cabal? It may be related to
versioning.

Regards,
Adrian

aquynh:
sure, they are all in the latest version, from Brew.

```
$ cabal --version
cabal-install version 1.22.6.0
using version 1.22.4.0 of the Cabal library 

$ ghc --version
The Glorious Glasgow Haskell Compilation System, version 7.10.3
```

adrianherrera:
Woops, not an issue with GHC/Cabal versions at all; it was an error in the Makefile. I've fixed the Makefile so it should all build fine now. Note that there is a known problem when c2hs is specified as a dependency in a cabal file, so you may have to install c2hs manually. Hopefully this is described clearly in `bindings/haskell/README.txt`.

Any further issues please let me know!

aquynh:
so after install cabal & ghc from Brew, i did this to install c2sh:

```
$ sudo cabal install c2hs
Resolving dependencies...
Configuring language-c-0.5.0...
Failed to install language-c-0.5.0
Build log ( /Users/imac/.cabal/logs/language-c-0.5.0.log ):
Configuring language-c-0.5.0...
setup-Simple-Cabal-1.22.5.0-x86_64-osx-ghc-7.10.3: The program 'happy' is
required but it could not be found.
cabal: Error: some packages failed to install:
c2hs-0.28.1 depends on language-c-0.5.0 which failed to install.
language-c-0.5.0 failed during the configure step. The exception was:
ExitFailure 1
```
do i miss any steps?

thanks

adrianherrera:
Yeah you basically need to recurse and manually install the dependencies that fail to install. E.g. `cabal install happy`, which will probably ask you to install alex (`cabal install alex`), and keep going until you can install c2hs successfully.

From http://community.hfm.io/hc/en-us/articles/216666257-Usage-of-the-command-line-tools:

> Building some Haskell packages, such as the C->Haskell binding generator c2hs, requires other Haskell build tools; in the case of c2hs, this is lexer generator alex and the parser generator happy (they are essentially lex and yacc for Haskell). These build tools are usually also available as Haskell packages. However, the cabal package manager is not able to install them automatically — instead, you need to install them explicitly before installing the package that depends on them. In our example, you need to install alex and happy before attempting to installing c2hs.

aquynh:
so to install c2hs, 2 packages "happy" & "alex" need to install as dependency. i think the output of "cabal instal" is not very intuitive.

would you mind adding this info to bindings/haskell/README, so in the future people can easily solve this problem without having to ask around?

besides, there is still this error when running "make" under bindings/

```
...
cd haskell && cabal install --dependencies-only && cabal configure && cabal build
cabal: /tmp/unicorn_sample: does not exist
make: *** [all] Error 1
```

it seems this depends on some binaries to be built before running "make". 
is there anyway to avoid this? (currently "make" under bindings/ simply process the prototypes in `include/`, and does not depend on anything.

thanks.

adrianherrera:
Updated Haskell README as requested.

However, I do not get the `/tmp/unicorn_sample` error. Can you please give me the exact steps to reproduce? If you look at the Makefile's `all` target, there is no reference to `TMPDIR`, so I do not understand why this error would occur.

aquynh:
i can fix this issue, but is this necessary to run `cabal install --dependencies-only` every single time, or just once is enough? if once is enough, is it better to put this into haskell/README?

aquynh:
i pushed a new branch "haskell" for this, with a fix for above issue of "missing /tmp/unicorn_sample" (on Mac, you simply run "make" under `bindings/` to reproduce this bug). please base your next PRs on this branch until we merge this into "master".

a question: `const_generator.py` can generates all the symbols such as UC_X86_REG_EAX automatically for all other bindings. but i cannot find these symbols in `haskell/`. how do you handle this?

thanks

adrianherrera:
Symbols such as UC_X86_REG_EAX are generated by c2hs. After `cabal build` is run, new Haskell files get generated in `haskell/dist/build/`. So UC_X86_REG_EAX appears in `haskell/dist/build/Unicorn/CPU/X86.hs`.

adrianherrera:
Also note that I am running Linux, and I do not get that error about "missing /tmp/unicorn_sample".

aquynh:
how can i test the samples under `haskell/` now?

adrianherrera:
Standard ghc build, using `ghc --make samples/SampleArm.hs`

aquynh:
a final request: could you put those commands to build & install in `haskell/README.TXT` into `haskell/Makefile`, so it is easier for users because they just need to run "make" or "make install" instead of copying/pasting from README.TXT? a "clean" target is also nice to have in `Makefile`.

thanks.

adrianherrera:
No worries, I have pushed some changes to the Makefile. Please let me know what you think!

I do not have permissions to push to the `haskell` branch, so I merged `haskell` into the `haskell-binding` branch so we can track everything in this PR.

aquynh:
if you have a `Makefile` under `bindings/haskell`, users can install it separately without having to install everything with `bindings/Makefile`. it is good to move the "clean" target there as well.

adrianherrera:
Sorry, I misread what you previously said! I will do that.

adrianherrera:
Actually, on second thought, if you are doing everything directly in "Haskell land", then you only really need to run `cabal install`. Cabal files take care of all the building and essentially act as Makefiles in Haskell. Therefore, I think it is redundant to include a Makefile as well. I have updated the readme to reflect this.

aquynh:
i think Haskell is a bit different from the rest: when we run "make" under `bindings/`, we want to generate the symbols for all the bindings. but Haskell seems to do this at build time, so this is not necessary.

so i think we can remove the below line from `bindings/Makefile`

        cd haskell && cabal install --dependencies-only && cabal configure && cabal build

aquynh:
but then we should move this line to `bindings/haskell/Makefile` instead ?

adrianherrera:
I do not believe a Makefile for the Haskell bindings is required. The cabal file essentially takes the role of a Makefile in "Haskell land".

Once you `cd bindings/haskell`, you can run `cabal install` to install and `cabal clean` to cleanup, so there is no need for a Makefile.

I agree that the Haskell bindings should be treated separately from the other bindings, because it uses its own tools (c2hs) to generate the required symbols. Therefore it doesn't really make sense to have it in `bindings/Makefile`. I will remove the Haskell build from `bindings/Makefile` and then I think we are done. What do you think?

aquynh:
if `cabal install` can implicitly take care `cabal install --dependencies-only && cabal configure && cabal build` then yes, we are done.

adrianherrera:
Cool! Hopefully this is the last commit then :-) Thanks Quynh

aquynh:
merged, thanks for all the amazing effort you put into this binding!

adrianherrera:
Great, no worries! You can probably delete the `Haskell` branch too. Please let me know of any API changes that would require an update to the Haskell binding.

aquynh:
yes i did. will ping you if the binding needs to be taken care of, thanks!

PR #518, opened 2016-04-08
Small improvements to align with hapstone (Haskell bindings for Capstone)

adrianherrera:
`Enum`s now derive `Bounded` and I've made use of `Applicative` in the Utils module.

### DIFF
--- a/bindings/haskell/src/Unicorn/CPU/Arm.chs
+++ b/bindings/haskell/src/Unicorn/CPU/Arm.chs
@@ -24,6 +24,6 @@ import Unicorn.Internal.Core (Reg)
     omit (UC_ARM_REG_INVALID,
           UC_ARM_REG_ENDING)
     with prefix="UC_ARM_REG_"
-    deriving (Show, Eq) #}
+    deriving (Show, Eq, Bounded) #}
 
 instance Reg Register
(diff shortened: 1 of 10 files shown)


ibabushkin:
Hapstone author here. Do you plan to write tests for your bindings? I did, using `quickcheck` and `hspec`, and it caught some bugs arising due to incompatibilities on `c2hs`' side. That's what the bounded instances are originally for (though they are useful elsewhere as well), because you can write simple `Arbitrary` instances like this:
```Haskell
instance Arbitrary CsArch where
    arbitrary = elements [minBound..maxBound]
```
So if you need tests, maybe hapstone's test suite is worth a look ;)

Cheers.

adrianherrera:
I did get the idea to write some tests after looking at hapstone. I will
look into it. I'm hoping there aren't any incompatibles with c2hs though!

Thanks for the comment.

ibabushkin:
The main issue with `c2hs` was the fact that it doesn't support C11, which capstone uses (most notably anonymous unions, which leads to problems with the `{#alignof#}`, `{#sizeof#}`, `{#offsetof#}` `{#get#}`, `{#set#}`.

aquynh:
merged, thanks!

adrianherrera:
@ibabushkin thanks for the heads up! Fortunately, there weren't any C11 features (that I was aware of anyway) in Unicorn that I had to verify.

lunixbochs:
Unicorn isn't even C99

Issue #523, opened 2016-04-15
When emulating 64-bit code on 32-bit systems, memory mapping fails for regions above 0x100000000


cseagle:
When emulating 64-bit code on 32-bit systems, memory mapping fails for regions above 0x100000000.

This appears to be because ram_addr_t is defined:
typedef uintptr_t ram_addr_t;
in qemu/include/exec/cpu-common.h

On 32 bit systems uintptr_t is only 4 bytes so begin addresses get truncated in memory_map:
MemoryRegion *memory_map(struct uc_struct *uc, ram_addr_t begin, size_t size, uint32_t perms)

I recommend changing ram_addr_t to uint64_t matching the the definition of hwaddr.

Test case is attached. NOTE: the test system must be built as a 32 bit binary to exhibit the failure.
[mem_64_c.txt](https://github.com/unicorn-engine/unicorn/files/220598/mem_64_c.txt)

aquynh:
this means Qemu cannot emulate 64bit on 32bit system? hmm i always think that Qemu can do this :-/

cseagle:
I have no idea about QEMU. I did provide you a test case that demonstrates the problem. Based on looking at QEMU source and Unicorn source, the problem appears to be due to your choice of ram_addr_t in your memory_map function in memory.c. That function calls the QEMU memory_region_add_subregion function and memory_region_add_subregion takes a hwaddr type. So, you have truncated the callers address in your memory_map before passing through to memory_region_add_subregion which is perfectly capable of accepting a 64 bit address.

As in my original submission, I suggest changing to a 64-bit data type for memory_map and memory_map_ptr by changing the begin argument to be a hwaddr instead of a ram_addr_t.

radare:
ida-style

aquynh:
another fix for this is to convert the data type of `begin` in `memory_map()` and `memory_map_ptr()` from `ram_addr_t` to `hwaddr`. i think this is a better solution. what do you think?

cseagle:
I actually tried that and things failed, so that approach to fixing it requires deeper changes than it first appears. I think the use of ram_addr_t extends deeper down the call tree so changing the arguments to memory_map just pushes the problem deeper without fixing it.

aquynh:
i just put your testcase into `tests/regress/`.

aquynh:
[...]

PR #526, opened 2016-04-18
promote ram_addr_t to uint64_t always to allow emulating 64-bit syste…

cseagle:
…ms on 32-bit systems. This addresses issue #523 which prevents unicorn from emulating full 64-bit address spaces on 32-bit systems.


aquynh:
i pushed a commit to fix this issue, but with a different approach: https://github.com/unicorn-engine/unicorn/commit/8b030ae51ab65f142a2f567ebbcd437c62c276ae.

please confirm, so we can close this pull req, thanks.

cseagle:
This seems to fix things for my test cases

aquynh:
thank you, Chris!

PR #530, opened 2016-04-20
Add ASAN build target with UNICORN_ASAN or via './make.sh asan'

zachriggle:
Note that using this with Python bindings (or, e.g. in GDB) will require adding `start_deactivated=1` to `ASAN_OPTIONS`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,6 +64,13 @@ else
 CFLAGS += -O3
 endif
 
+ifeq ($(UNICORN_ASAN),yes)
+CC = clang -fsanitize=address -fno-omit-frame-pointer
+CXX = clang++ -fsanitize=address -fno-omit-frame-pointer
+AR = llvm-ar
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -122,7 +129,7 @@ EXT = so
 VERSION_EXT = $(EXT).$(API_MAJOR)
 AR_EXT = a
 $(LIBNAME)_LDFLAGS += -Wl,-Bsymbolic-functions,-soname,lib$(LIBNAME).$(VERSION_EXT)
-UNICORN_CFLAGS += -fvisibility=hidden
+# UNICORN_CFLAGS += -fvisibility=hidden
```
aquynh: why did you comment out this line? i am not sure this is false positive, but Travis reported an issue for this PR.

aquynh:
merged, but with that line to set `UNICORN_CFLAGS +=-fvisibility=hidden` uncommented. if you have a good reason to do so, please send another PR. thanks.

Issue #532, opened 2016-04-24
Coredump


mrphrazer:
If something will be written into unmapped memory, qemu will crash.

Python bindings:
```python
from unicorn import *

code = "\x90"
ADDRESS = 0x1000000
mu = Uc(UC_ARCH_X86, UC_MODE_64)
mu.mem_write(ADDRESS, code)
```

The same holds for the following C testcase:

```C
#include <unicorn/unicorn.h>

#define X86_CODE64 "\x90"
#define ADDRESS 0x1000000

int main(int argc, char **argv, char **envp)
{
    uc_engine *uc;

    uc_open(UC_ARCH_X86, UC_MODE_64, &uc);
    uc_mem_write(uc, ADDRESS, X86_CODE64, sizeof(X86_CODE64) - 1);

    uc_close(uc);

    return 0;
}
```

Output
```
qemu: qemu_mutex_destroy: Device or resource busy
Aborted (core dumped)
```

Valgrind output:

```
==17239== Memcheck, a memory error detector
==17239== Copyright (C) 2002-2015, and GNU GPL'd, by Julian Seward et al.
==17239== Using Valgrind-3.11.0 and LibVEX; rerun with -h for copyright info
==17239== Command: ./test
==17239== 
==17239== Conditional jump or move depends on uninitialised value(s)
==17239==    at 0x56FB03E: tcg_target_init_x86_64 (tcg-target.c:2280)
==17239==    by 0x56FB5DD: tcg_context_init_x86_64 (tcg.c:367)
==17239==    by 0x56E9D15: cpu_gen_init_x86_64 (translate-all.c:132)
==17239==    by 0x56EA4C4: tcg_exec_init_x86_64 (translate-all.c:719)
==17239==    by 0x57861E4: tcg_init (accel.c:46)
==17239==    by 0x5786439: accel_init_machine (accel.c:116)
==17239==    by 0x5786229: configure_accelerator (accel.c:73)
==17239==    by 0x5785F9F: machine_initialize (vl.c:123)
==17239==    by 0x5D66CF9: uc_open (uc.c:263)
==17239==    by 0x40073E: main (in /tmp/test)
==17239==  Uninitialised value was created by a stack allocation
==17239==    at 0x4E6E35B: ??? (in /usr/lib/libglib-2.0.so.0.4800.0)
==17239== 
==17239== Conditional jump or move depends on uninitialised value(s)
[...]

PR #533, opened 2016-04-24
added regression test write before map

mrphrazer:
Regression test for https://github.com/unicorn-engine/unicorn/issues/532

### DIFF
--- a/tests/regress/write_before_map.py
+++ b/tests/regress/write_before_map.py
@@ -1,0 +1,25 @@
+#!/usr/bin/env python
+
+from __future__ import print_function
+from unicorn import *
+from unicorn.x86_const import *
+
+import regress
+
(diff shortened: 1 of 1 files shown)


aquynh:
merged, but i think you ran your testcase with an older version of Unicorn.
here is its output with latest Unicorn:

```
./write_before_map.py
E
ERROR: runTest (__main__.WriteBeforeMap)
  File "./write_before_map.py", line 21, in runTest
    mu.mem_write(ADDRESS, X86_CODE64)
  File "/Library/Python/2.7/site-packages/unicorn/unicorn.py", line 304, in mem_write
    raise UcError(status)
UcError: Invalid memory write (UC_ERR_WRITE_UNMAPPED)

Ran 1 test in 0.014s

FAILED (errors=1)
```

Issue #538, opened 2016-05-03
reg_read not working for SIMD registers on ARM


mkravchik:
Hi,
It looks that SIMD registers (UC_ARM_REG_D0 - UC_ARM_REG_D32) are not read by reg_read. They work correctly, but when I read them I always get 0. By working correctly I mean that if I read data into them using vld and them write it into memory using vsd the memory is updated as expected. Test attached. The expected result would be to see non-null values in the D6 and D7.

Has anyone encountered this? (I will add the code shortly)

aquynh:
can you send a PR to put your testcase in tests/regress/ ?

mkravchik:
Yes, where do I get this regress module from?

aquynh:
that comes with Python by default i think. if you can run any .py files there, you are already set.

mkravchik:
Done #539. Only vld.py file is relevant, the other one is from the closed issue. I just did not find a way to get rid of it (first experience with GitHub)

aquynh:
oh making PR is very simple. this is worth spending some little time, so you can follow below steps now:

- from inside your local git repo on your machine, pull the latest code of Unicorn, so your local copy has the latest code: `git pull` (be sure you are on "master" branch first)
- make a new branch named "vld" with `git checkout -b vld`
- now copy your testcase "vld.py" to `tests/regress/`
- push this "vld" branch to your Github fork with `git push origin vld`
- go to Github webpage, and make a new pull request

let me know if you have any issues this time.

mkravchik:
Thanks a lot for the detailed steps. I tried them and mostly succeeded, but the second file still got thru :-(

aquynh:
this is because you still works on a local repo that you already committed that "vmov.py" file. you can either clean your repo by removing that file, commit in your local repo, then pushing it to your Github fork. or whenever in doubt, fork a new Github and start on a fresh local repo with all the above steps.

aquynh:
the reason you dont get the right values for those registers is because they are not handled in arm_reg_read() function yet: https://github.com/unicorn-engine/unicorn/blob/master/qemu/target-arm/unicorn_arm.c#L47

this is easy to fix, so look forward to your next PR :-)

mkravchik:
Done, hopefully!

mkravchik:
Just tried again from a fresh branch to get rid of the conflict. Hope it will be OK and thanks for the guidance!

aquynh:
fixed, thanks!

PR #540, opened 2016-05-03
SIMD registers not read on ARM (Thumb mode)

mkravchik:
tests/regress/vld.py contains the unit test for the problematic case # #538 (the second file is irrelevant and can be discarded)


aquynh:
merged, with that vmov.py removed. thanks!

PR #543, opened 2016-05-04
Reading and writing NEON registers

mkravchik:
This is a fix for #538 
The vld.py regression test seems OK


aquynh:
looks good, but did you forget the new version of "vld.py"?

now you can just commit vld.py locally, then push to the same branch remotely with the same command "git push origin vld", and the commit will show up here. 

i will merge 2 commits at the same time, thanks.

mkravchik:
No need, the vld.py in the master is already correct. Maybe I had screwed it locally after previous commit.

aquynh:
merged, thanks!

PR #548, opened 2016-05-10
Add new bindings and fixed UC_HOOK_INSN hooking.

### LINKED ISSUE #541: uc_hook_add with UC_HOOK_INSN not working
Like it sounds, when I open the [example](https://github.com/unicorn-engine/unicorn/blob/master/samples/sample_x86.c) (via the visual studio MSVC project), and compile it, all code works, but any hook that uses UC_HOOK_INSN to hook on a specific instruction does not trigger.

Replacing the line with `uc_hook_add(uc, &trace1, UC_HOOK_BLOCK, hook_syscall, NULL, 1, 0);` generates the trigger fine, so i believe this is an error with UC_HOOK_INSN.

Alternatively, the sample code could be incorrect.

xorstream:
Added MSVC bindings for uc_reg_write_batch() and uc_reg_read_batch() and fixed UC_HOOK_INSN hooking.
Also fixed the use of stdbool.h in MSVC since it is not included in MSVC.
(This fixes #541)

### DIFF
--- a/include/unicorn/unicorn.h
+++ b/include/unicorn/unicorn.h
@@ -10,7 +10,7 @@ extern "C" {
 
 #include <stdint.h>
 #ifdef _MSC_VER
-typedef int bool;
+typedef unsigned char bool;
 #define false 0
 #define true 1
 #else
--- a/include/unicorn/unicorn.h
+++ b/include/unicorn/unicorn.h
@@ -9,7 +9,13 @@ extern "C" {
 #endif
 
 #include <stdint.h>
+#ifdef _MSC_VER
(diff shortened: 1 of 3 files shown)


include/unicorn/unicorn.h
```diff
@@ -9,7 +9,13 @@ extern "C" {
 #endif
 
 #include <stdint.h>
+#ifdef _MSC_VER
+typedef int bool;
```
aquynh: i think the right one is:

    typedef unsigned char bool;

aquynh:
merged, thanks!

PR #549, opened 2016-05-10
fixed go bindings typo

eqv:
Fixed broken type in go bindings

### DIFF
--- a/bindings/go/unicorn/unicorn.go
+++ b/bindings/go/unicorn/unicorn.go
@@ -154,7 +154,7 @@ func (u *uc) RegReadBatch(regs []int) ([]uint64, error) {
 }
 
 func (u *uc) MemRegions() ([]*MemRegion, error) {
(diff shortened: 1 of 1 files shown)


lunixbochs:
can't reproduce locally - I think you have a different problem. what's your Go version?

lunixbochs:
oh, it should work either way due to how uc_mem_region is defined. the change looks fine, but I'm still curious if this caused you any actual problems.

eqv:
go 1.2 which afaik is the default for the current ubuntu.

lunixbochs:
you should really be running 1.5+ (the binary distribution is very easy to use)

aquynh:
Ryan, so do you want to merge this PR?

lunixbochs:
yeah, I think the PR is right (not really a typo, but the "final" type is more correct even though they're equivalent)

aquynh:
merged, thanks!

Issue #552, opened 2016-05-17
flush tlb for multiple emulation


5lipper:
Memory mapping operations (memory_map/memory_map_ptr/memory_unmap) flush tlb only when we are in a running state. If we map/unmap a page between two emulations, tlb is not flushed since `uc->current_cpu` is NULL.

``` python
import unicorn

uc = unicorn.Uc(unicorn.UC_ARCH_X86, unicorn.UC_MODE_64)

#0:   48 c7 c0 00 20 00 00    mov    rax,0x2000
#7:   48 c7 00 34 12 00 00    mov    QWORD PTR [rax],0x1234
# e:   48 8b 00                mov    rax,QWORD PTR [rax]
code = '48c7c00020000048c70034120000488b00'.decode('hex')
code_addr = 0x1000
data_addr = 0x2000

uc.mem_map(code_addr, 0x1000)
uc.mem_map(data_addr, 0x1000)
uc.mem_write(0x1000, code)
uc.emu_start(code_addr, code_addr + len(code))
print 'rax = %#x' % uc.reg_read(unicorn.x86_const.UC_X86_REG_RAX)
uc.mem_unmap(data_addr, 0x1000)
uc.emu_start(code_addr, code_addr + len(code))
print 'rax = %#x' % uc.reg_read(unicorn.x86_const.UC_X86_REG_RAX)
```

This will cause a segfault in jitted code (write to unmapped memory).
Should we flush tlb at the beginning of each emulation?

aquynh:
yes it sounds right to me. can you send a PR?

thanks

5lipper:
#553

PR #553, opened 2016-05-17
flush tlb for each emulation issue #552

5lipper:


lunixbochs:
what are the performance implications of this if I'm rapidly stopping/restarting the emulator?

aquynh:
How about flushing whenever we unmap memory?

5lipper:
```c
void memory_unmap(struct uc_struct *uc, MemoryRegion *mr)
{
    int i;
    target_ulong addr;
    Object *obj;

    // Make sure all pages associated with the MemoryRegion are flushed
    // Only need to do this if we are in a running state
    if (uc->current_cpu) {
        for (addr = mr->addr; addr < mr->end; addr += uc->target_page_size) {
           tlb_flush_page(uc->current_cpu, addr);
        }
    }
```
Only flushing the unmapped pages is a better idea, but how can we get
the assigned cpu if we are not in a running state?

aquynh:
in the case the emulation is not running, we can fall back to the first solution: flush all the cache inside uc_emu_start().

lunixbochs:
why even use `uc->current_cpu`? why not just `CPU_FOREACH` and flush 'em all? (I don't think a single uc will or should use multiple cpus anyway)

5lipper:
@lunixbochs I did that in a commit before this one, but it's not merged. Then I think @aquynh is talking about flushing all cache when it's not running. I don't know if a single uc can only use one cpu :(

lunixbochs:
I really don't like the idea of flushing all cache when we know only a couple of pages need to be flushed - that's not efficient :)

lunixbochs:
diff --git a/qemu/memory.c b/qemu/memory.c
    index 3ab5c7a..a0c0b3e 100644
    --- a/qemu/memory.c
    +++ b/qemu/memory.c
    @@ -71,12 +71,13 @@ void memory_unmap(struct uc_struct *uc, MemoryRegion *mr)
         int i;
         target_ulong addr;
         Object *obj;
    +    CPUState *cpu;

         // Make sure all pages associated with the MemoryRegion are flushed
         // Only need to do this if we are in a running state
    -    if (uc->current_cpu) {
    +    CPU_FOREACH(cpu) {
             for (addr = mr->addr; addr < mr->end; addr += uc->target_page_size) {
    -           tlb_flush_page(uc->current_cpu, addr);
    +           tlb_flush_page(cpu, addr);
             }
         }
         mr->enabled = false;

Issue #571, opened 2016-06-09
ARM NEON/VFP support seems to exist but is disabled by default


petmac:
Hello,

Sorry if this is a known issue. I searched but didn’t notice it.

I’m trying to emulate an ARM processor with NEON (Cortex-A9). Emulation of the “vpush” instruction failed so I started investigating why.

According to the processor technical reference manual, in order to access VFP and NEON registers, both cp10 and cp11 of CPACR have to be enabled:

http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.ddi0388f/CIHGHGJE.html

And indeed, when floating point is disabled, QEMU generates an EXCP_UDEF exception in disas_vfp_insn.

When VFP is disabled (ARM_VFP_FPEXC bit 30 is clear), disas_vfp_insn returns failure for instructions like vpush.

Thanks,
Peter

aquynh:
can you send a testcase for this (put it under tests/regress/ or tests/unit/), so i can debug this problem?

thanks

petmac:
Thanks, I will do that.

I believe the reset state of the CPU is to have FP and VFP disabled, so I don't think it's a bug. It would be nice though to be able to enable them programmatically without having to run code to do it.

aquynh:
have you seen this issue yet? https://github.com/unicorn-engine/unicorn/issues/446

petmac:
Hi,

Sorry, no I didn't notice it. I actually do very similar though. Before I run user programs I run some code to enable FP and VFP.

Thanks for looking into it.

Peter

petmac:
I guess it would be nice to be able to set these registers without having to inject ARM/Thumb code to do it.

lunixbochs:
Yep, this comes up for me in x86 and mips as well. We definitely need an MSR API, which could just be special constants for reg_write() with an MSR prefix or something (no need to make a separate API function because it already uses opaque types)

aquynh:
We can think about how to enable VFP without extra guest code, but for now
can you modify your testcase so it passes?

neobrain:
The status of this issue seems a bit unclear to me.

@lunixbochs, have you gotten unicorn to successfully run VFP code? The test merged in #580 does not seem to work for me (throwing an `UC_ERR_FETCH_UNMAPPED` error), but from what was said here I'm not sure whether it's supposed to illustrate how to enable VFP or just to document that this feature does not work in unicorn yet. Have you gotten unicorn to emulate VFP code at all?

[...]

PR #572, opened 2016-06-10
Add arm_fp_vfp_disabled.py test.

petmac:
Test case supplied on request. Relates to issue #571.

Please let me know if you need more information or would prefer the test case to be written in C rather than Python, or any other changes.

Thanks for looking into this,
Peter


aquynh:
Python test is fine. merged, thanks!

PR #579, opened 2016-06-15
added ARM64 hang test

NUko32:
Added the test for AArch64 emulation hanging.


aquynh:
looks good, but there are some issues, can you please fix them?

- indentation: please dont mix tabs & spaces.
- this is not really an unit test following the existing cmocka form. you can see sample tests like tests/unit/test_pc_change.c for example. even though currently your test does not get to the assertion phase yet, we still need a proper one so it will pass when this bug is fixed.

NUko32:
I have fixed indentations and refactored the test code, it is now using cmocka.

aquynh:
merged, thanks!

PR #586, opened 2016-06-30
Python 3 compatibility: raise unicode exception messages

4577:
In order to make exceptions display well with the Python binding on Python 3.x, return exception message as unicode/str rather than str/bytes.

### DIFF
--- a/bindings/python/unicorn/unicorn.py
+++ b/bindings/python/unicorn/unicorn.py
@@ -156,7 +156,7 @@ class UcError(Exception):
         self.errno = errno
 
     def __str__(self):
-        return _uc.uc_strerror(self.errno)
+        return _uc.uc_strerror(self.errno).decode('ascii')
 
 
 # return the core's version


danghvu:
Can you post here an example that show the difference? Just for clarity

4577:
Wrong:

```>>> from unicorn import Uc
>>> Uc(42, 42)
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.5/site-packages/unicorn/unicorn.py", line 215, in __init__
    raise UcError(status)
unicorn.unicorn.UcError>>> ```

Good:

```>>> from unicorn import Uc
>>> Uc(42, 42)
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.5/site-packages/unicorn/unicorn.py", line 215, in __init__
    raise UcError(status)
unicorn.unicorn.UcError: ERROR: Invalid/unsupported architecture (UC_ERR_ARCH)
>>> ```

The str() function also fails with `TypeError: __str__ returned non-string (type bytes)`.

danghvu:
Thanks!

aquynh:
Yes, we fixed a similar bug in Keystone recently. Perhaps Capstone also has the same issue?

PR #588, opened 2016-07-11
add ifdef for compiling with mingw

cseagle:


danghvu:
Do you know if this flag works for all window version ?

cseagle:
I have no way to test across all Windows versions, but it fixes both the 32 and 64 bit builds on Windows 7.

danghvu:
I suppose removing the inclusion will work for you too?

cseagle:
That's fine as long as it doesn't break other things. You call munmap later in that file if I recall correctly.

danghvu:
Not anymore, see latest commit. Thanks!

Issue #391, opened 2016-01-22
ARM and THUMB interchange


xorstream:
ARM and THUMB modes are not mutually exclusive modes, it is very common for the mode to dynamically change during execution. Therefore it doesn't make sense to set them as a static mode in uc_open.

There are a probably few ways the mode changes but one of the most common is to branch via the BX instruction to an odd address to change to thumb mode or an even address for arm mode. When in thumb mode the bit value 0x20 is set in the processor status register (PSR), when in arm mode this bit is cleared. Quick link to PSR bits: https://en.wikipedia.org/wiki/ARM_architecture#Registers

Based on this information it makes sense to allow the arm/thumb mode to be set dynamically either via PC manipulation or PSR manipulation. It is much less of a headache to just set a bit in the status register instead of specially handling the "odd bit" of an instruction address in all cases.

Therefore I suggest that we allow dynamic selection of arm/thumb mode via the "thumb mode" bit in the status register. And that we do away with the static mode values for ARM/THUMB.

Note that the above suggestion does not currently work! If I emulate ARM code that jumps to THUMB mode, then I reposition the PC to some arm code and alter the PSR value to specify to run in ARM mode, it will still change itself back into thumb mode when it executes the instruction, thereby failing to execute the arm instruction.

aquynh:
> ARM and THUMB modes are not mutually exclusive modes, 

at initialize time (when we call `uc_open()`), it should be either ARM or THUMB, not both. this is what i mean when i said "mutually exclusive".

aquynh:
at the run-time, we will get the mode dynamically, but do not rely on the initialized mode. so there is no problem with the value UC_MODE_ARM=0.

aquynh:
for the last part of your issues, we need a  newmechanism (via a new API `uc_option()` as i once mentioned) to set the ARM mode at runtime.

xorstream:
If you are setting up an emulator for non-trivial cases you will likely call uc_open() to create your ARM emulator before you know whether the code you plan to execute is ARM or THUMB. I guess if there was a dynamic option for setting the mode, ie via some uc_option() function, then you could just select ARM during uc_open() and then dynamically alter the value to ARM/THUMB later via a call to uc_option() before then starting execution.

Btw do you consider the fact that I cannot change the value of the PSR via uc_reg_write() to be a bug?

aquynh:
yes, uc_option() should solve your problem, so you can dynamically change the mode at runtime.

[...]

PR #592, opened 2016-07-25
Set thumb mode based on PC value in ARM. Mask off last bit of PC.

esanfelix:
I tested this with code compiled using a cross-compiler from the Android NDK and seems to work fine. I don't see why it would fail, but it might be worth running more ARM tests on it before merging.

See issue #391


petmac:
It could be worth doing a quick pass over the test suite, to set the Thumb bit of the address passed to uc_emu_start for any tests that set Thumb mode with uc_open. Apologies if you have already done this and there are none.

aquynh:
merged, thanks.

if you find any issues with this change, please report.

PR #595, opened 2016-07-28
Added unit tests for X86 (SHL bug with the Parity Flag + ENTER)

samothtronicien:
There is a bug with the SHL instruction that occurs when using CL as source operand.
The parity flag is not calculated correctly, in the example below, the flag is set to 0 whereas it should be set to 1.

> mov eax, 3Ch
> mov cl, 2
> shl eax, cl <- do not set PF

The bug doesn't occured while using an immediate value.

> mov eax, 3Ch
> shl eax, 2 <- set PF correctly

The bug is present in 32/16/8 bits as illustrated in the tests.


aquynh:
would you please put this testcase under tests/regress, or (even better) tests/unit?

thanks.

samothtronicien:
Sorry about that, I've moved it to tests/regress.
I will add a unit test about this later today.

samothtronicien:
I've added more tests and noticed that the issue only occurs while we are retrieving EFLAGS in the hook.
Once the emulation is done, the value retrieved of EFLAGS is correct.

samothtronicien:
I've added n unit test for this issue (tests/unit/test_x86_shl.cpp).

aquynh:
Great, but:

- dont you add this to Makefile?

- all of our code is pure C, we never use C++ here. Would you mind converting your code to C?

Thanks

samothtronicien:
@aquynh : the test was ported to C and added to the makefile.

samothtronicien:
I've added some tests on ENTER/LEAVE to show some weird behavior on the ENTER instruction when the nesting level > 0.

aquynh:
do you want to remove tests/regress/test_x86_shl.cpp?

samothtronicien:
Thanks, forgot about that one. 
I've removed it.

aquynh:
merged, thanks!

PR #596, opened 2016-07-29
Get Python samples running on Python 3

andrew-d:


aquynh:
Merged, thanks a lot, Andrew!

Issue #313, opened 2015-12-12
Another issue with EFLAGS.


farmdve:
So I was executing code in Unicorn and noticed that the EFLAGS I get from uc_reg_read was inconsistent with real execution,furthermore if I were to single-step in Unicorn e.g with max_count = 1; the internal eflags did not properly sync and it affected branches and it didn't execute properly. So I started digging in and I was wondering if the following code is wrong in cpu.h

``` c
static inline uint32_t cpu_compute_eflags(CPUX86State *env)
{
    return env->eflags0 | cpu_cc_compute_all(env, CC_OP) | (env->df & DF_MASK);
}
```

Here we are using `env->eflags0`, but I think this must be `env->eflags`. In fact, after I changed it to `env->eflags`, `uc_reg_read` was returning correct results and my single-step was no longer executing wrong branches. E.g

``` c
static inline uint32_t cpu_compute_eflags(CPUX86State *env)
{
    return env->eflags | cpu_cc_compute_all(env, CC_OP) | (env->df & DF_MASK);
}
```

Cause in `x86_cpu_exec_exit` we have a redundant `env->eflags0 = env->eflags;` when  `cpu_compute_eflags` returns  `env->eflags0` which made me think that it was a simple typo.

``` c
void x86_cpu_exec_exit(CPUState *cs)
{
    X86CPU *cpu = X86_CPU(cs->uc, cs);
    CPUX86State *env = &cpu->env;

    env->eflags = cpu_compute_eflags(env);
    env->eflags0 = env->eflags;
}
```

Re-examining my old eflags_nosync test, it fails with my change, though. With my change ZF is set somehow, without it ZF is not set(which is the correct behaviour for that test).
So I further do this

``` c
static inline uint32_t cpu_compute_eflags(CPUX86State *env)
{
    env->eflags &= ~(CC_O | CC_S | CC_Z | CC_A | CC_P | CC_C | DF_MASK);
    return (env->eflags | cpu_cc_compute_all(env, CC_OP) | (env->df & DF_MASK)) & (~IOPL_MASK);
}
```

This eliminates the problem. Well, the other problem is that in Windows(ring3) the default value for EFLAGS is 0x202, if I use that eflags value, my eflags_nosync test does not fail.

Furthermore, the IOPL bits in eflags(bits 12-13) are set. In my own copy of Unicorn I clear them. However I know that people will use Unicorn to emulate firmware, and other non-OS specific code, so how do we unset these flags internally without manual patch?
[...]

PR #606, opened 2016-08-10
eflags patch

rhelmot:
This is a patch to the x86 eflags computation. I don't know what it does, exactly, since someone else on the angr team wrote this and I don't understand the diff, but it makes the unicorn eflags match the libVEX eflags in tests we run to ensure accuracy.


danghvu:
Could you please provide an tiny example to show the different of the output before/after the patch ?

Nukem9:
for reference: https://github.com/unicorn-engine/unicorn/issues/313#issuecomment-197729397

rhelmot:
Added a test.

aquynh:
merged, thanks!

PR #607, opened 2016-08-10
support xmm registers

rhelmot:
Support reading and writing to and from the x86 xmm registers. Only implements the python bindings.

### DIFF
--- a/bindings/python/unicorn/unicorn.py
+++ b/bindings/python/unicorn/unicorn.py
@@ -202,6 +202,14 @@ class uc_x86_float80(ctypes.Structure):
     ]
 
 
+class uc_x86_xmm(ctypes.Structure):
+    """128-bit xmm register"""
+    _fields_ = [
+        ("lowdword", ctypes.c_uint64),
+        ("highdword", ctypes.c_uint64),
+    ]
+
+
 class Uc(object):
     def __init__(self, arch, mode):
         # verify version compatibility with the core before doing anything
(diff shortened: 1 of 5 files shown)


aquynh:
this looks good, but would you mind writing some sample code for C & Python to demonstrate this new feature?

rhelmot:
Pushed samples and test cases.

rhelmot:
btw there's some real messed up shit going with the unit test build system, I had to make modifications to the makefile to get anything to build and then one of the tests failed to build

change to makefile:
```diff
diff --git a/tests/unit/Makefile b/tests/unit/Makefile
index 39f2747..deace76 100644
--- a/tests/unit/Makefile
+++ b/tests/unit/Makefile
@@ -52,4 +52,4 @@ test_hang: test_hang.c
 test_x86_shl_enter_leave: test_x86_shl_enter_leave.c

 ${ALL_TESTS}:
-       ${CC} ${CFLAGS} -o $@ $^
+       ${CC} $^ ${CFLAGS} -o $@
```

(I re-pushed the test commit because I accidentally committed this change)

failing build:

```
cc test_hang.c -Wall -Werror -Wno-unused-function -g -L ../../ -lcmocka -lunicorn -I ../../include -o test_hang
test_hang.c: In function ‘test_hang’:
test_hang.c:90:9: error: format ‘%llx’ expects argument of type ‘long long unsigned int’, but argument 2 has type ‘uint64_t {aka long unsigned int}’ [-Werror=format=]
  printf("x0: %#llx\n", x0);
         ^
test_hang.c:91:9: error: format ‘%llx’ expects argument of type ‘long long unsigned int’, but argument 2 has type ‘uint64_t {aka long unsigned int}’ [-Werror=format=]
  printf("x1: %#llx\n", x1);
         ^
```

bindings/python/unicorn/unicorn.py
```diff
@@ -202,6 +202,14 @@ class uc_x86_float80(ctypes.Structure):
     ]
 
 
+class uc_x86_xmm(ctypes.Structure):
+    """128-bit xmm register"""
+    _fields_ = [
+        ("lowdword", ctypes.c_uint64),
```
aquynh: what do you think about having "low_dword" & "high_dword" instead?

aquynh: what is your feedback?

rhelmot: Oh, dang, this completely dropped off my radar. I'm fine with that, let me make the change real quick.

rhelmot:
`lowdword` => `low_qword`, `highdword` => `high_qword`

aquynh:
merged with a minor fix for sample_x86.c, thanks!

Issue #266, opened 2015-11-14
Segmentation fault: "Invalid read of size 4" in cpu_tb_exec


practicalswift:
The code below segfaults Unicorn with:

```[…] testcase[6970]: segfault at fffffffffffffff4 ip 00007f84ad23df34 sp 00007f84a9a8e8c0 error 5 in libunicorn.so.0[7f84ad23d000+1000]```

valgrind output:

```
==6979== Invalid read of size 4
==6979==    at 0x59B5F04: ??? (in /usr/lib/libunicorn.so.0)
==6979==  Address 0xfffffffffffffff4 is not stack'd, malloc'd or (recently) free'd
```

Code:

```c
#include <unicorn/unicorn.h>

static void hook_block(uc_engine *uc, uint64_t address, uint32_t size, void *user_data) {
  printf("hook_block(…)\n");
}

#define BINARY "\x80\x05\xff\xff\xff\xff\x30\xeb\xf7\x30"
#define MEMORY_SIZE 2 * 1024 * 1024
#define STARTING_ADDRESS 0x1000000

int main(int argc, char **argv, char **envp) {
  uc_engine *uc;
  if (uc_open(UC_ARCH_X86, UC_MODE_64, &uc)) {
    printf("uc_open(…) failed\n");
    return 1;
  }
  uc_mem_map(uc, STARTING_ADDRESS, MEMORY_SIZE, UC_PROT_ALL);
  if (uc_mem_write(uc, STARTING_ADDRESS, BINARY, sizeof(BINARY) - 1)) {
    printf("uc_mem_write(…) failed\n");
    return 1;
  }
  uc_hook hook;
  uc_hook_add(uc, &hook, UC_HOOK_BLOCK, hook_block, NULL, 1, 0);
  printf("uc_emu_start(…)\n");
  uc_emu_start(uc, STARTING_ADDRESS, STARTING_ADDRESS + sizeof(BINARY) - 1, 0, 0);
  printf("done\n");
  return 0;
}
```

[...]

PR #608, opened 2016-08-10
Make normal self-modifying code work

rhelmot:
#266 was patched incorrectly. The fix made it so that the entire block would be re-lifted instead of just a single instruction, which discards the relevant self-modifications. The real fix is to keep an explicit flag for whether we've actually added the hook_block instrumentation code, and what index in the tcg arrays to insert it.


rhelmot:
I don't know how to add a test to your CI setup, but the test is that the following program, assembled with an executable stack, should return 65.

```bash
$ as --32 stuff.s -o stuff.o
$ ld -melf_i386 -z execstack stuff.o -o stuff
$ ./stuff
$ echo $?
65
```

here's stuff.s:

```asm
.intel_syntax noprefix

.global _start
_start:
    mov ebp, esp
    sub ebp, 0x4000
    mov edx, ebp

    lea esi, [self_modifying]
    mov edi, ebp
    mov ecx, 0x2d
    call memcpy
    add ebp, 0x2d
    xor ebx, ebx
    call edx

    mov eax, 1
    int 0x80

memcpy:
    cmp ecx, 0
    je _end
    dec ecx
    mov al, byte ptr [esi+ecx]
    mov byte ptr [edi+ecx], al
    jmp memcpy

_end:
    ret

self_modifying:
    inc ebx
    call $+5
    pop esi
    dec byte ptr [esi+11]
    xor edx, edx
    sub esi, 6
_loop_start:
    cmp edx, 5
    jz _loop_end

    mov edi, ebp
    mov ecx, 0x2d
    lea eax, [memcpy]
    call eax
    inc edx
    add ebp, 0x2d
    mov byte ptr [ebp], 0xc3
    jmp _loop_start

_loop_end:
```

rhelmot:
Added regression test with this testcase.

tests/regress/x86_self_modifying.s
```diff
@@ -0,0 +1,51 @@
+.intel_syntax noprefix
```
aquynh: a minor thing: would you please add some instructions on how to compile it into this file?

aquynh:
merged, but please send another pull req to add instructions on how to compile the asm file. thanks!

PR #609, opened 2016-08-10
Fix a qemu bug that causes infinite hangs on the square root of malformed long doubles

rhelmot:
I'll submit this patch to upstream qemu too!

### DIFF
--- a/qemu/fpu/softfloat.c
+++ b/qemu/fpu/softfloat.c
@@ -5355,8 +5355,12 @@ floatx80 floatx80_sqrt( floatx80 a STATUS_PARAM )
         z.high = floatx80_default_nan_high;
         return z;
     }
+    if ( aSig0 == 0 ) return packFloatx80( 0, 0, 0 );
+    while (aExp != 0 && (aSig0 & 0x8000000000000000) == 0) {
+        aExp--;
+        aSig0 <<= 1;
+    }
     if ( aExp == 0 ) {
-        if ( aSig0 == 0 ) return packFloatx80( 0, 0, 0 );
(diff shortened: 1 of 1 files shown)


aquynh:
can you please also provide a testcase for this issue?

rhelmot:
```asm
  xor eax,eax
  inc eax
  push eax
  push eax
  push eax
  fld tbyte ptr [esp]
  fsqrt
```

This will hang unless the patch is applied. How can I add a testcase to unicorn?

Nukem9:
Test case was already added by another user: https://github.com/unicorn-engine/unicorn/pull/590/commits/3eaceec07e3c603d3607fd9b13f2a3387058041d. Code is different, though.

rhelmot:
What the fuck? I discovered this bug on my own and made the original fix on July 10... It's plausible that someone could have discovered the bug independently, I guess?

Whatever. This is a fix, and I've submitted an identical patch to upstream qemu.

aquynh:
this looks good, as that existing testcase passed.

let see what the upstream says about this fix.

aquynh:
is there any feedback from Qemu on this patch yet?

rhelmot:
Qemu rejected the initial patch, requesting that we do something more
faithful to the Intel dev manual (return NaN and raise a fp exception), and
also that we apply this check to ALL floatx80 arithmetic. This patch is
currently going through code review. I'll cc you in future correspondence.

rhelmot:
Patch got accepted upstream! The final version is drastically different from this one, though, so I'll redo the pull request.

PR #612, opened 2016-08-11
fix free() -> g_free()

lunixbochs:
this makes precorn possible on glib < 2.46 :)


aquynh:
merged, thanks!

lunixbochs:
it's now working on linux hosts if you want to take a look :)

PR #614, opened 2016-08-16
make.sh: add nix32 option

idl3r:


aquynh:
Perhaps you will want to change this option to "linux32"?

idl3r:
The option is named "nix32" to comply with the options mentioned in docs/COMPILE-NIX.md:

> To cross-compile Unicorn on 64-bit OS to target 32-bit binary, run:
> $ ./make.sh nix32

Of course it can be linux32 if we update the document as well.

aquynh:
Please fix the docs, too.

PR #615, opened 2016-08-17
make.sh: add linux32 option

idl3r:
Add linux32 option to make.sh, also update COMILE-NIX.md to reflex this change.

### DIFF
--- a/docs/COMPILE-NIX.md
+++ b/docs/COMPILE-NIX.md
@@ -74,9 +74,9 @@ To build Unicorn on *nix (such as MacOSX, Linux, *BSD, Solaris):
 
         $ UNICORN_QEMU_FLAGS="--python=/path/to/python2" ./make.sh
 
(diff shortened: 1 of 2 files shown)


aquynh:
is there any missing dependencies that can be added to the docs as well? i cannot compile this on Ubuntu 16.04, 64bit.

idl3r:
Actually no non-documented dependencies are required. We just need to install their i386 version. For my 14.04 box following two packages were installed.
- linux-libc-dev:i386
- libglib2.0-dev:i386

The first one is generic support headers for all kinds of development. The second one is GLib 2.0 which is a documented requirement for Unicorn. I don't have a 16.04 box for testing but I believe both are avaiable for 16.04 as well.

aquynh:
merged, with an additional commit to add "libglib2.0-dev:i386" as dependency. thanks!

PR #617, opened 2016-08-18
Add Python-style dict for const values

stephengroat:
Use python-style dict so that accessing const values can be done with mutable strings.

I'm not wedded to this idea, just hate that I can't easily programmatically go through the constants; python dicts provide that function

Indexing through const values before:

``` python
for x in range(0,10):
    exec("y = UC_ARM_REG_R" + str(x))
    print "R" + str(x) + ": " + '%08x'%mu.reg_read(y)
```

Indexing through python dict:

``` python
for x in range(0,10):
    print "R" + str(x) + ":" + '%08x'%mu.reg_read(arm_const['UC_ARM_REG_D' + str(x))
```

### DIFF
--- a/bindings/const_generator.py
+++ b/bindings/const_generator.py
@@ -25,9 +25,9 @@ template = {
             'comment_close': '',
         },
     'python-dict': {
-            'header': "# For Unicorn Engine. AUTO-GENERATED FILE, DO NOT EDIT [%s_const_dict.py]\n%s_const = {\n",
+            'header': "# For Unicorn Engine. AUTO-GENERATED FILE, DO NOT EDIT\n%s_const = {\n",
             'footer': "}",
-            'line_format': '\'UC_%s\': %s\n',
+            'line_format': '\'UC_%s\': %s,\n',
             'out_file': './python/unicorn/%s_const_dict.py',
             # prefixes for constant filenames of all archs - case sensitive
             'arm.h': 'arm',
--- a/bindings/const_generator.py
+++ b/bindings/const_generator.py
@@ -24,6 +24,22 @@ template = {
             'comment_open': '#',
             'comment_close': '',
         },
+    'python-dict': {
+            'header': "# For Unicorn Engine. AUTO-GENERATED FILE, DO NOT EDIT [%s_const_dict.py]\n%s_const = {\n",
+            'footer': "}",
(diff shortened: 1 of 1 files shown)


aquynh:
with this PR, when i do this:

    $ cd bindings; make

it does not really generate <arch>_const_dict.py files under `bindings/python/unicorn/`.
does that work for you?

stephengroat:
I understand. I didn't update the `Makefile`, I was more looking for comments/approval on the concept. You can generate the new, updated files with a direct call to the python script

    python const_generator.py python-dict

If you think it's a good idea, I'll also spend the time to replace the old-style variables with the new dictionary style interface throughout the rest of the code.

aquynh:
yes please go ahead.

aquynh:
Oh no, i mean i still want to run "make" to build everything. Thanks

jroweboy:
While its probably a good idea to update it to use a dictionary, just wanted to point out that currently you can avoid the use of `eval` by using `global` instead. So the examples would look like this 

globals()['a']

aquynh:
oh right, so there is no need to have this dictionary anymore?

stephengroat:
globals are a little dangerous IMHO. dictionaries are much more "pythonic"

stephengroat:
I was looking at different ways of doing this. would it be better to have both (variables and a dict) during a transition period or do a hard switch of the entire repo over to a dict?

lunixbochs:
Making the primary method of accessing reg enums a dict will:

- Fragment the Python API to be less compatible with the other languages.
- Make current code more verbose and slower.

If you're already doing that I'd be more in favor of generating proper [enum types](https://docs.python.org/3/library/enum.html) (which unfortunately won't support Python 2 without some glue) or even going all in on a magic Python cpu wrapper class.

stephengroat:
Could you describe what you're thinking with the "magic python cpu wrapper class"? I'm not not follow on that ...

I think the Python 3 enum would be nice, but would probably be nasty for 2.x users. Since 2.7 is still so prevalent, that might be a harder pill for people to swallow

I know this isn't a small change, I just want to see if others think some sort of change away from variables would be a good idea. If not, I'll close

XVilka:
Shouldn't this be closed? Python 2 is removed, enums are the way to go, imho.

PR #620, opened 2016-08-20
Port patch from upstream to check for invalid long double encodings

rhelmot:
This is the fixed version of #609 that got accepted upstream.


aquynh:
this still fails to build? https://travis-ci.org/unicorn-engine/unicorn/builds/153772978

rhelmot:
Fixed :| That's what I get for pushing untested code at 4am...

aquynh:
merged, thanks!

PR #621, opened 2016-08-20
Basically, make python builds work as expected

rhelmot:
- Build the shared library on python build/install
- Package the shared library and static library with the python library
- Package the headers with the python library, this allows the python distribution to be used as a canonical distribution if your project that depends on unicorn requires both the python bindings and links directly against the C library (this is the case for angr)
- Generally, clean up the python setup process
- Drastically simplify the python shared library import process

This is a *lot* of invasive changes to the build process, but what I see is that they just make it _work_ instead of having to have an `install_unicorn.sh` in the root of all of our projects that depend on unicorn. This was the case for a while in our internal repos...

If you like these, I have a similar pull request I can open against capstone!

### DIFF
--- a/bindings/python/Makefile
+++ b/bindings/python/Makefile
@@ -1,30 +1,28 @@
 # Python binding for Unicorn engine. Nguyen Anh Quynh <aquynh@gmail.com>
 
-OBJDIR = ./build
-
-.PHONY: gen_const install install3 clean sdist sdist3 bdist bdist3
+.PHONY: gen_const install install3 clean sdist sdist3 bdist bdist3 sdist_win bdist_win
 
 gen_const:
 	cd .. && python const_generator.py python
 
 install:
-	rm -rf $(OBJDIR) src/
+	rm -rf src/ dist/
 	rm -rf prebuilt/win64/unicorn.dll
 	rm -rf prebuilt/win32/unicorn.dll
 	if test -n "${DESTDIR}"; then \
-		python setup.py build -b $(OBJDIR) install --root="${DESTDIR}"; \
+		python setup.py install --root="${DESTDIR}"; \
 	else \
-		python setup.py build -b $(OBJDIR) install; \
+		python setup.py install; \
 	fi
 
 install3:
-	rm -rf $(OBJDIR) src/
+	rm -rf src/ dist/
 	rm -rf prebuilt/win64/unicorn.dll
 	rm -rf prebuilt/win32/unicorn.dll
 	if test -n "${DESTDIR}"; then \
-		python3 setup.py build -b $(OBJDIR) install --root="${DESTDIR}"; \
+		python3 setup.py install --root="${DESTDIR}"; \
 	else \
-		python3 setup.py build -b $(OBJDIR) install; \
+		python3 setup.py install; \
 	fi
 
 # build & upload PyPi package with source code of the core
@@ -68,7 +66,7 @@ sdist3_win:
 	python3 setup.py sdist register upload
 
 clean:
-	rm -rf $(OBJDIR) src/ dist/ MANIFEST
+	rm -rf src/ dist/ build/ MANIFEST
 	rm -rf prebuilt/win64/unicorn.dll
 	rm -rf prebuilt/win32/unicorn.dll
 	rm -rf unicorn/lib unicorn/include
--- a/bindings/python/Makefile
+++ b/bindings/python/Makefile
@@ -32,7 +32,6 @@ sdist:
 	rm -rf src/ dist/
 	rm -rf prebuilt/win64/unicorn.dll
 	rm -rf prebuilt/win32/unicorn.dll
-	cp PKG-INFO.src PKG-INFO
 	python setup.py sdist register upload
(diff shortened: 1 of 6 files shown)


rhelmot:
Just pushed a bunch more code that makes the codebase pypi-ready. I was successfully able to build `sdist`, `bdist_wheel`, and `bdist_egg` distributions that worked correctly when installed with root, with --user, and in a virtual environment.

aquynh:
this looks good. i will spend some more time reviewing it before merging, thanks!

bindings/python/Makefile
```diff
@@ -45,31 +44,46 @@ sdist3:
...
-	cp README.pypi-src README
 	cp PKG-INFO.src PKG-INFO
 	python3 setup.py sdist register upload
 
+# build & upload PyPi package with precompiled core
+bdist:
+	rm -rf src/ dist/
+	rm -rf prebuilt/win64/unicorn.dll
+	rm -rf prebuilt/win32/unicorn.dll
+	cp PKG-INFO.src PKG-INFO
```
aquynh: looks like you copied this from Capstone? we dont have PKG-INFO.src in Unicorn

bindings/python/unicorn/unicorn.py
```diff
@@ -17,12 +17,12 @@
...
     range = xrange
 
-_lib_path = os.path.split(__file__)[0]
-_all_libs = (
-    "unicorn.dll",
-    "libunicorn.so",
-    "libunicorn.dylib",
-)
+if sys.platform == 'darwin':
+    _lib = "libunicorn.1.dylib"
```
aquynh: the part `.1` depends on the Unicorn version. for example, when we bump version to 2.0, the lib will be `libunicorn.2.dylib`. how about using libunicorn.dylib to avoid hardcode the version here (this is the current approach)?

bindings/python/Makefile
rhelmot: This wasn't copied from capstone, this was copid from the `sdist` code six lines up :P

rhelmot: Granted, it's wrong, and I'll fix it

bindings/python/unicorn/unicorn.py
rhelmot: I switched all references to the library name to the versioned name, since that's what you did with capstone. I can switch them all back.

rhelmot:
Pushed things to address your concerns.

aquynh:
on MacOS, this cannot be built. below is the later part of the output

```
$ cd bindings/python
$ sudo make install
.....
creating build/lib/unicorn/include
creating build/lib/unicorn/include/unicorn
copying unicorn/include/unicorn/arm.h -> ./build/lib/unicorn/include/unicorn
copying unicorn/include/unicorn/arm64.h -> ./build/lib/unicorn/include/unicorn
copying unicorn/include/unicorn/m68k.h -> ./build/lib/unicorn/include/unicorn
copying unicorn/include/unicorn/mips.h -> ./build/lib/unicorn/include/unicorn
copying unicorn/include/unicorn/sparc.h -> ./build/lib/unicorn/include/unicorn
copying unicorn/include/unicorn/unicorn.h -> ./build/lib/unicorn/include/unicorn
copying unicorn/include/unicorn/x86.h -> ./build/lib/unicorn/include/unicorn
running build_clib
building 'unicorn' library
creating build/temp.macosx-10.11-intel-2.7
ar rc ./build/temp.macosx-10.11-intel-2.7/libunicorn.a
ar: no archive members specified
usage:  ar -d [-TLsv] archive file ...
       	ar -m [-TLsv] archive file ...
       	ar -m [-abiTLsv] position archive file ...
       	ar -p [-TLsv] archive [file ...]
       	ar -q [-cTLsv] archive file ...
       	ar -r [-cuTLsv] archive file ...
       	ar -r [-abciuTLsv] position archive file ...
       	ar -t [-TLsv] archive [file ...]
       	ar -x [-ouTLsv] archive [file ...]
error: command 'ar' failed with exit status 1
make: *** [install] Error 1

$
```

rhelmot:
I don't know exactly what went wrong for you, since I don't have an OS X to test on, but I pushed some changes that should hopefully fix your problem - I moved all the compilation from `build_clib` to `build`, since `build_clib` happens after the `egg_info` step which means the shared library never gets picked up by the distribution, and I guess having the `libraries` setup key invokes build_clib automatically?

Regardless, there's an additional fix in there that makes `sudo make install` work correctly on my machine - I don't understand why, but running the `build -b` step causes the resulting installation to have files whose filenames are missing their first two letters. Removing `build -b` makes it work correctly for me and should have no repercussions, since `build` is executed automatically as part of `install`.

aquynh:
yes it builds on Mac now. let me do some more tests ...

aquynh:
merged, thanks!

rhelmot:
Perfect! Thank you so much :)

What would be the process/timeline of getting unicorn distributed on pip?

aquynh:
soon, i will do that after merging your remaining PR.

PR #622, opened 2016-08-20
New feature: registers can be bulk saved/restored in an opaque blob

rhelmot:
This is a feature that is required for UC integration in angr. We need to be able to efficiently roll back the state at a whim when we encounter symbolic data that cannot be processed by unicorn.

Originally, our internal code was just including the appropriate private header files and performing this operation manually, but as we're gearing up for distribution, this won't do. So, new feature!

I've done nothing but the bare minimum to demonstrate that it works and is useful since I don't know if this is a feature you'd appreciate having.


aquynh:
yes, this is useful, but my concern is that it overlaps with the existing APIs uc_reg_read_batch()  & uc_reg_write_batch(). have you considered these APIs?

rhelmot:
That would technically still work, but there are two problems preventing it from being feasible:

- It would require every single register in the entire state to be accessible through unicorn's API, which is clearly not the case right now and would take a LOT of work to make that happen
- Ultimately, this function is meant to be fast above all else, and not only does the structure of the batch reads/writes preclude the speed afforded by an optimized memcpy and requires a lot of additional effort to set up the structures that must be passed in, but unicorn performs a lot of additional computation to convert between the qemu internal state and the "real x86 values" that would be seen in e.g. gdb.

Our application needs this called once per basic block, so it needs to be as fast as possible.

danghvu:
+1 this change, seems to be something people want to do.

aquynh:
@lunixbochs, do you have any comments?

lunixbochs:
An opaque save API is useful. I'm mostly using the batch API for faster syscall/function dispatch. I don't think register read/write is mature enough yet to save the whole state, and there's more overhead in setting up the batch API.

rhelmot:
I wrote a simple test for save/restore but it seems to be failing? `uc->current_cpu` appears to be null even after stepping forward an instruction. Is this field only populated while emulation is active? What should I do to get a reference to the registers I'm trying to copy other than `uc->current_cpu->env_ptr`, which is a line of code that I myself didn't write, I just found it while porting the copy code from an internal hack to a real patch.

aquynh:
uc->current_cpu should not be NULL in this case.

can you post your testcase?

rhelmot:
I pushed the test case immediately before making that comment :P

rhelmot:
fixed it by using the `first_cpu` macro. Also added python bindings and an example of using them!

uc.c
```diff
@@ -1160,3 +1160,36 @@ uc_err uc_query(uc_engine *uc, uc_query_type type, size_t *result)
 
     return UC_ERR_OK;
 }
+
+size_t cpu_regs_size(uc_arch arch, uc_mode mode);
+size_t cpu_regs_size(uc_arch arch, uc_mode mode) {
```
aquynh: to be consistent with our code style, please put all the open function bracket `{` at the begin of the next line.

rhelmot:
Fixed braces

aquynh:
for consistency with existing API, what do you think about renaming these API?

- uc_restore_regstate() rename to uc_regstate_load()
- uc_save_regstate() rename to uc_regstate_save()

rhelmot:
I'm going to reject to `uc_regstate_load` - in other cases, "load" refers to loading values from emulated memory, where here it would refer to writing values to emulated memory. I will change them to `uc_regstate_save` and `uc_regstate_restore`.

bindings/python/sample_x86.py
```diff
@@ -329,7 +329,7 @@ def test_i386_reg_save():
         mu.emu_start(address, address+1)
 
         print(">>> save the register state")
-        saved_regs = mu.save_regs()
+        saved_regs = mu.regs_save()
```
aquynh: i think the method name should be consistent with the API name. so instead of `regs_save`, how about `regstate_save`?

danghvu:
One issue I found uneasy is the use of un-safe type `void*`. This enables user to shot themselves in the foot by exposing the raw data. We should discourage the access to the raw sate.

One suggestion is to add a `uc_regstate_alloc()` and `uc_regstate_free()` which returns/clean a opaque type `regstate_t` wrapped around the buffer (and size maybe). 

Then you can use this type inside `_save` and `_restore`. This also cleans the code path of passing a `NULL` pointer to `_save` which IMO is ugly.

aquynh:
on a second thought, `regstate` is still not the best name yet, as the internal structure contains more than just registers. so can we can change `regstate` to `state`? sorry for this trouble, and i can rename them for you.

rhelmot:
@danghvu: Those would work. I'd be more in favor of keeping everything the same as it is currently but with `typedef void* regstate_t` or maybe `typedef struct regstate {} regstate_t`, since I designed the current setup to work with my usecase (what I expect to be the common usecase) of save-many-times-restore-once. If you're strongly opposed to the current design regardless, it can be changed.

@aquynh: I considered that, but initially decided against it because I didn't want to imply that the methods would capture the current memory state as well. If you like, you can rename them!

aquynh:
you are right, but i dont think there is a perfect name.

another choice (suggested by @danghvu) is "context". what do you think?

rhelmot:
Context is good! Another alternative would be "processor_context", but I don't know if that's too lengthy.

danghvu:
@rhelmot Do you mean you need many different buffers for different saves? I agree in that case it  is probably slightly more verbose at the initialization, the clean up should be the same, you must free these buffers yourself in the current design too. Everything else you can keep the same.

On the other hand, calling native `free`  on  anything other than pointer is weird so I'm still in favor with an unicorn API for "context" `alloc/free`. Or we can just keep it as you have right now, I'm not strongly opposing it.

aquynh:
i renamed "_regstate_" to "_context_", and merged this PR to a new branch "context" at https://github.com/unicorn-engine/unicorn/tree/context. please base the next PRs on this branch, thanks.

rhelmot:
Added explicit alloc/free functions, also added a bit of error checking. Should be harder to shoot yourself in the foot now.

uc.c
```diff
@@ -1174,20 +1174,47 @@ size_t cpu_regs_size(uc_arch arch, uc_mode mode)
 }
 
 UNICORN_EXPORT
-void *uc_context_save(uc_engine *uc, void *buffer)
+uc_err uc_context_alloc(uc_engine *uc, uc_context **context)
 {
-    size_t sz = cpu_regs_size(uc->arch, uc->mode);
-    if (!buffer) {
-        buffer = malloc(sz);
+    size_t size = cpu_regs_size(uc->arch, uc->mode);
```
aquynh: do you want to rename cpu_regs_size to cpu_context_size, or something?

tests/unit/test_x86.c
```diff
@@ -742,6 +742,7 @@ static void test_x86_16(void **state)
 static void test_i386_reg_save(void **state)
 {
     uc_engine *uc;
+    uc_context *saved_regs;
```
aquynh: is it better to be "saved_context", or something?

include/unicorn/unicorn.h
```diff
@@ -273,6 +273,15 @@ typedef enum uc_query_type {
     UC_QUERY_PAGE_SIZE,
 } uc_query_type;
 
+// Metadata stub for the variable-size cpu context used with uc_context_*()
+typedef struct uc_context {
+   uc_arch arch;
+   uc_mode mode;
+   size_t size;
+   bool used;
```
aquynh: i have some concerns about this struct, but anyway i think user should be responsible for how uc_context_t is used, but should not push this to the framework with this "used" field.

bindings/python/unicorn/unicorn.py
```diff
@@ -443,27 +446,36 @@ def hook_del(self, h):
...
+        if status != uc.UC_ERR_OK:
+            raise UcError(status)
+
+        return SavedContext(ptr)
 
-    def context_restore(self, store):
-        if type(store) is not _ActivePointer:
-            raise TYpeError("Bad register store %s" % repr(store))
-        _uc.uc_context_restore(self._uch, store.pointer)
+    def context_update(self, context):
```
aquynh: what is this method for?

rhelmot:
Pushed fixes as requested. The python method you are asking about is for updating an existing saved context without allocating a new one. This is the cleanest way to do it in line with the way the API is currently set up.

aquynh:
i have few concerns about the struct **uc_context**: i feel that the fields **arch, mode & used** should be tracked by your app, but not by the core. i suppose that you need these info so you can save the emulation, and restore any time later. but keep in mind that future version of Unicorn may modify the internal context, thus break all the saved emulation. in that case **arch, mode & used** are not useful, and **uc_context** never has enough information that you need.

rhelmot:
Alright. Those fields serve no purpose to me, I only added them to be able
to provide additional sanity checks and prevent the user from shouting
themselves in the foot. I can remove them, but I'm not sure I see the point
of your example, since that structure is hidden from the user and can never
be written to disk...

danghvu:
This looks good to me.

btw: I don't know why my commits in the master branch are added here.. maybe something weird happens when you rebase/merge ?

aquynh:
Yes, just remove those fields, then i will merge this

bindings/python/sample_x86.py
```diff
@@ -312,6 +312,56 @@ def test_i386_inout():
...
+        mu.mem_write(address, code)
+
+        print(">>> set eax to 1")
+        mu.reg_write(UC_X86_REG_EAX, 1)
+
+        print(">>> execute 'inc eax'")
+        mu.emu_start(address, address+1)
+
+        print(">>> save the register state")
+        saved_regs = mu.context_save()
```
aquynh: do you want to rename this "saved_regs" variable as well?

rhelmot:
I've done the things you've asked, though I still truly do not understand why you want them. These fields cannot be accessed by the user, cannot be serialized to disk, and serve no purpose other than preventing the user from shooting themself in the foot. Honestly, I think they serve as justification for having there be a structure for the saved context instead of just a void\* - if we're not going to have safety considerations for the user, why do we care about giving them a void pointer instead of a structure? It's not like the user can actually do anything with the void\* that they can't do with an equally opaque pointer by casting it to a char array or whatever. And really, there's no need to store the size along with the data, it can just be recalculated every time. You seem to be pushing for a "compromise" between the two approaches (struct+safety/void pointer+simplicity) that discards the good parts of both of them.

Regardless, I'm beyond caring about design at this point; I just want this merged. If there are complications in the future related to design, I will offer my opinions again then.

aquynh:
sorry if that was not clear, but my point is: why do you want to keep arch/mode/used in the core, but do not maintain these info in your own app? the other fields "data" & "size" make sense, but the above 3 fields do not belong to the core, especially the "used" field.

rhelmot:
I do not maintain this information in my own app. I don't need to, since in my app these instances will always be attached to a class which is slaved to a given uc_engine instance.

I'm not sure what you're not understanding - by putting the struct definition in uc_priv.h, it is truly opaque - no user app will ever be able to see the arch, mode, and used fields - nobody will ever be able to see them except the core, which uses them to ensure that the user hasn't passed invalid data into the call.

aquynh:
yes i do understand that the struct is opaque (i designed uc_struct in the similar way), but adding those fields still cannot guarantee user cannot pass invalid data. that is why i think that trick is not helpful.

rhelmot:
In that case the pull request is complete and if you like it, it can be merged.

aquynh:
i just made a minor change to turn cpu_context_size() to a static function. merged, thanks!

this is yet another high quality pull req from you, thanks so much for the incredible effort: https://github.com/unicorn-engine/unicorn/commit/b991b71cf5603bbf7c7804f05a9948e1d54c4840

rhelmot:
I actually noticed that already - as soon as you merged it I pulled to push it to our copy of the repository, and saw the commit then! :D Glad to be able to contribute.

aquynh:
i just realized we only got a Python sample for this new API, but missed the C sample. would be great if you can send another PR for that, too!

PR #625, opened 2016-08-22
Run unit in travis

stephengroat:


aquynh:
i think cmocka unit is not even compilable on some Linux distros.

stephengroat:
I'm still going to test around in travis to see if I can get something to work, make just the regression tests

Issue #565, opened 2016-06-02
Support registering IO register read/write handlers


neobrain:
It seems to me that unicorn does not expose a way to expose "active" memory (i.e. MMIO), which triggers some sort of hardware action when being written to/read from. For instance, a game console emulator may want to recognize writes to a particular address as a rendering command to the console's GPU and then consequently trigger the GPU emulation code.

For best flexibility, I would propose unicorn to invoke a user-defined callback function whenever it encounters a read from/write to an unmapped address. Alternatively, the list of reads/writes invoking the callback function could be restricted to a previously defined list of IO registers.

I've been looking through the issue tracker and surprisingly found no existing requests for such a feature, even though I thought it would be fairly common use case in low-level emulation.

lunixbochs:
Unicorn has a hook for "unmapped read".

neobrain:
@lunixbochs I'm aware of the unmapped read/write hooks, but there are one small and one large problem with them, as far as I can see anyway:
* for writes, it's not clear to me what the `uc_cb_eventmem_t` documentation means with "continue"ing. Would it resume execution, skipping the current instruction? Or instead, would it re-execute the current instruction, expecting the previously unmapped address to be mapped now? In either case, how can I be sure the instruction operation has actually completed successfully (e.g. in the case of a stack push, which needs to modify the stack pointer after the memory access)?
* for reads, there is no simple way to actually return a value from the handler to unicorn (which of course would be necessary when the emulated application is reading an IO register value into a CPU register). I suppose you could stop emulation, map a new memory region, write the register value to the appropriate offset, then re-run unicorn for exactly one instruction, and then unmap that memory region, and finally resume emulation normally. That seems overly hacky, though.

neobrain:
@chriscz I don't want to map static memory, though - emulating IO registers properly requires a custom handler to be installable on memory access (to be able to make emulated hardware devices react appropriately to a particular read/write).

kitling:
@aquynh 

Is there any intention of implementing an API for this, or is there a (preferrably easy) way to do this in the state that unicorn is currently in?

(I'd like to implement GPIO and similar for my project.)

aquynh:
Yes, can you post your proposal? If we all agree, then plz send pull req to
implement it, i will merge.

kitling:
Warning: I don't design APIs, so this may be dirty.

[...]

PR #627, opened 2016-08-26
[RFC] Add uc_mmio_map for memory mapped devices

kitling:
This is an implementation of my second proposal for issue #565. I will admit, I am unfamiliar with the codebase, so there are likely many improvements that can be made.

I'll probably push a commit to my project, raspcorn, that uses this PR, soon.
### TODO
- [ ] Make an example, and perhaps some unit tests for this.

### DIFF
--- a/uc.c
+++ b/uc.c
@@ -1144,6 +1144,34 @@ uint32_t uc_mem_regions(uc_engine *uc, uc_mem_region **regions, uint32_t *count)
 }
 
 UNICORN_EXPORT
+uc_err uc_mmio_map(uc_engine *uc, uint64_t address, size_t size,
+        uc_cb_mmio_t callback, void *user_data)
+{
+    uc_err res;
+    struct mmio_data *data;
+
+    if (callback == NULL)
+        return UC_ERR_ARG;
+
+    if (uc->mem_redirect) {
+        address = uc->mem_redirect(address);
+    }
+
+    res = mem_map_check(uc, address, size, UC_PROT_ALL);
+    if (res)
+        return res;
+
+    data = g_new(struct mmio_data, 1);
+    *data = (struct mmio_data){
+        .user_data = user_data,
+        .callback = callback
+    };
+
+
+    return mem_map(uc, address, size, UC_PROT_ALL, uc->memory_map_io(uc, address, size, data));
+}
+
+UNICORN_EXPORT
 uc_err uc_query(uc_engine *uc, uc_query_type type, size_t *result)
 {
     if (type == UC_QUERY_PAGE_SIZE) {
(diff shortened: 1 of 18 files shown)


kitling:
Latest commit of raspcorn uses this in a basic way. Any changes to the API added in this PR will be reflected in my application as I develop it.

@aquynh I'm sorry if I'm being impatient, or if you're reviewing this in your mailing list, but I never really got the hang of mailing lists... >_>

aquynh:
Yes i do, but this is slow, so please be patient. Thanks for your
contribution!

aquynh:
would you please add a sample code using this new API? something like `samples/sample_mmio.c` will be great.

uc.c
```diff
@@ -1144,6 +1144,34 @@ uint32_t uc_mem_regions(uc_engine *uc, uc_mem_region **regions, uint32_t *count)
...
+
+    if (uc->mem_redirect) {
+        address = uc->mem_redirect(address);
+    }
+
+    res = mem_map_check(uc, address, size, UC_PROT_ALL);
+    if (res)
+        return res;
+
+    data = g_new(struct mmio_data, 1);
```
aquynh: where are we going to free this memory? is this memory leaking?

kitling: ... DOH. Can't believe that I forgot about this. It *is* leaking at the moment, isn't it? Need to: (A) remove the need for that struct. (B) be able to tell mmio apart from ram and free that memory when unmapping the mmio region. (C) create a seperate unmapping function.

You have any ideas on how to accomplish A? I don't mind changing the API to require 2 callbacks, for example, just depends on what you think is the best interface to present. I'll be happy that the API is there.

aquynh: are you relying on uc_mem_unmap() to free the resource for you? if then, then uc_mem_unmap() must be aware of this allocated memory to release that.

aquynh:
this adds a new API, however. another approach is to extend the existing API uc_mem_map(), so `perms` param accepts a new enum like UC_MMIO_MAP. so if you pass `UC_MMIO_MAP | UC_PROT_READ`, then your handler will be called upon reading that mmio area.

any objections?

aquynh:
in this pull req, your API set the callback via a new API, which is not consistent with the existing API: we add & delete the callbacks via `uc_hook_add()` & `uc_hook_del()`, but not in the mapping functions.

i think the API change proposed above keep it simple and consistent.

aquynh:
what do you have to say?

kitling:
I apologize, I've gotten busy as of late, between attempting to test the project (NTRPi) that I've been working on, and school being active for me.

That said, I still want to get something in, but I'm not working on this at the moment. >_>

lunixbochs:
I vote for MMIO extending the existing memory hook API. I might take a look at this.

markyi370:
Has anyone fixed the memory leak? Waiting for this to be implemented.

kitling:
Apologies once again. I haven't worked on this in... awhile. Just different project after different project. I ended up testing NTRPi on actual hardware, and found it to not be fast enough, so I have no real need for this anymore.

Since I don't find it likely that I'll come back to this, I'm closing the PR, as it has several issues.
 - It allocates memory without freeing it
 - It creates inconsistant new API
 - Even if it were a fine API, reading and writing should be seperate callbacks, anyway.
   - I think this could fix the memory leak, as iirc I only allocated memory so that I could squash the callbacks into the same function.

PR #630, opened 2016-09-04
Remove BUILDDIR-specific code

### LINKED ISSUE #629: Makefile deletes its work directory when used with Arch build system
It is not possible to install Unicorn using the yaourt package manager, because the Makefile deletes its own work directory.

This is due to this line from the Makefile:

    ifdef BUILDDIR
	    rm -rf $(BUILDDIR)
    endif

Here, BUILDDIR is an environment variable set to /tmp/yaourt-tmp-$HOME/aur-unicorn by makepkg, and the root for all files used by yaourt to perform. As $BUILDDIR is removed, the installation process fails with an error.

Thus I am not sure about the reason of the presence of this, BUILDDIR being not set anywhere in the source and seeming to be primarily an Archlinux/makepkg-specific variable. Can this be explained?

4577:
Fix #629

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -213,11 +213,7 @@ else
 endif
 endif
 
-ifndef BUILDDIR
 	@cd samples && $(MAKE)
-else
-	@cd samples && $(MAKE) BUILDDIR=$(BLDIR)
-endif
 endif
 
 config:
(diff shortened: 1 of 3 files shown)


aquynh:
merged, thanks!

Issue #638, opened 2016-09-16
sample_arm fails in thumb mode? 


disektor:
I am new to Unicorn.  When I run the sample_arm example, the ARM mode works but thumb mode seems to fail (Failed on uc_emu_start() with error returned: 8,  and the register value in SP is wrong).

Did I miss anything?

Here is the full output:

unicorn/samples$ ./sample_arm
Emulate ARM code
>>> Tracing basic block at 0x10000, block size = 0x8
>>> Tracing instruction at 0x10000, instruction size = 0x4
>>> Emulation done. Below is the CPU context
>>> R0 = 0x37
>>> R1 = 0x3456

==========================

Emulate THUMB code
>>> Tracing basic block at 0x10000, block size = 0x0
>>> Tracing instruction at 0x10000, instruction size = 0x4
>>> Tracing basic block at 0x10400, block size = 0x0
...
Failed on uc_emu_start() with error returned: 8
>>> Emulation done. Below is the CPU context
>>> SP = 0x1234

aquynh:
are you using the latest code from Github?

disektor:
Yes, a fresh clone from git on a x86_64,  Ubuntu 14.04.4 system.
Here is the uname and git status info.

~/unicorn/unicorn$ uname -a
Linux R320 3.13.0-37-generic #64-Ubuntu SMP Mon Sep 22 21:28:38 UTC 2014 x86_64 x86_64 x86_64 GNU/Linux

~/unicorn/unicorn$ git status
On branch master
Your branch is up-to-date with 'origin/master'.

nothing to commit, working directory clean

disektor:
Just checked out another copy.  Same behavior with sample_arm (good with ARM, but failure in thumb mode)

~/unicorn/unicorn$ ./samples/sample_arm
Emulate ARM code
>>> Tracing basic block at 0x10000, block size = 0x8
>>> Tracing instruction at 0x10000, instruction size = 0x4
>>> Emulation done. Below is the CPU context
>>> R0 = 0x37
>>> R1 = 0x3456

==========================

Emulate THUMB code
>>> Tracing basic block at 0x10000, block size = 0x0
>>> Tracing instruction at 0x10000, instruction size = 0x4
>>> Tracing basic block at 0x10400, block size = 0x0
... ...
>>> Tracing basic block at 0x20fc00, block size = 0x0
Failed on uc_emu_start() with error returned: 8
>>> Emulation done. Below is the CPU context
[...]

PR #639, opened 2016-09-20
Fixed ARM sample to properly run THUMB code #638

esanfelix:
Added change to sample_arm as discussed in #638


aquynh:
merged, thanks!

PR #640, opened 2016-09-20
bindings/python: fix install3

ret2libc:
before this patch:
```sh
$ sudo make install3
rm -rf src/ dist/
rm -rf prebuilt/win64/unicorn.dll
rm -rf prebuilt/win32/unicorn.dll
if test -n ""; then \
                python3 setup.py install --root=""; \
        else \
                python3 setup.py install; \
        fi
  File "setup.py", line 165
    print "Proper 'develop' support unavailable."
                                                ^
SyntaxError: Missing parentheses in call to 'print'
make: *** [install3] Error 1
```


aquynh:
merged, thanks!

PR #643, opened 2016-09-23
remove uc->cpus

lunixbochs:


aquynh:
it seems current_cpu can be removed. same for cpu_index?

aquynh:
merged, with some deadcode cleanups, thanks!

next, we can remove current_cpu & cpu_index?

PR #650, opened 2016-10-05
Fix samples make

stephengroat:
I'm starting to cleanup the make system, wanted your feedback. samples was a good first target

Was thinking of trying to do autoconf/autotools

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -213,7 +213,7 @@ else
 endif
 endif
 
-	@cd samples && $(MAKE)
+	+$(MAKE) -C samples
 endif
 
 config:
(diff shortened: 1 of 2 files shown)


danghvu:
Thank you, cleaning those up is great but I don't think we need a autoconf/tool.
p/s: It's breaking the build in travis & appveyor FYI, please fix.

stephengroat:
Maybe not autoconf/tool, but I'd really like  the tests (unit and regress) to be run by the CI system. It's great to have tests, but if you're not running them each time on your CI system, it's not doing as much good as it easily could.

I'm looking at building this tool into a larger flow, so I was trying to follow the GNU make conventions (source files go in `src`, builds go in `lib` or `bin`, etc)

stephengroat:
I'm closing this for right now, issue with my build systems. I'll re-open later

Issue #647, opened 2016-09-30
OSX compilation error


0xmilkmix:
Using either the 0.9 tarball or clone, I get this error while compiling using ./make.sh on OSX
/usr/local/Cellar/glib/2.50.0/include/glib-2.0/glib/gtypes.h:423:3: error:      '_GStaticAssertCompileTimeAssertion_0' declared as an array with a negative size

glib: stable 2.50.0 (bottled)
--universal

Any idea what could be wrong?

aquynh:
Did you follow the docs? There are some dependency that you must install
first.

Also, you should use the latest Github cide, not 0.9, since we fixed many
issues since.

0xmilkmix:
Yes, followed the docs and compiled glib in both 32 and 64 bits. pkg-config installed as well

aquynh:
Do you have same issue with Github code?

What is your clang version?

0xmilkmix:
Yes
glib header was included from unicorn/qemu/include/glib-compat.h:19
clang  is "Apple LLVM version 7.3.0 (clang-703.0.29)"

aquynh:
something must be wrong with your setup, as i have the same glib version, and use clang-703.0.31, but have no issues at all.

if you figure out what is wrong, please report back here.

0xmilkmix:
will continue looking into this and keep you informed. Thanks

0xmilkmix:
Solved it:
$ export CC=clang
$ ./make.sh macos-universal

aquynh:
interesting. what was the default CC then?

0xmilkmix:
strange thing is that env variable wasn't defined...

aquynh:
so you can compile now, but dont have universal binary? 

are you sure that you install glib with universal format? if you are not, then install again with:

    $ brew unlink glib
    $ brew install glib --universal

0xmilkmix:
unfortunately it does not solved the issue. Pulled the latest version this morning and it does not changed the outcome.

aquynh:
sorry, cant tell what is wrong with your system. Travis now compiles on OSX without any issues.

stephengroat:
@0xmilkmix if you look at the .travis.yml in #666, i now build MACOS_UNIVERSAL builds in travis

ludwigschwardt:
[...]

Issue #574, opened 2016-06-13
iOS compile issue


Nmzik:
When trying to compile Unicorn on Mac 10.11 i get this error: 

```
ERROR: zlib check failed
       Make sure to have the zlib libs and headers installed.
```

Do i need to compile zlib for arm? Any fix?
Thanks

lunixbochs:
@aquynh zlib has a good license, we could consider vendoring it into the source

Nmzik:
I just manually copied this test code to Xcode and successfully compiled it.
```
#include <zlib.h>
int main(void) { zlibVersion(); return 0; }
```
It's working on iOS. That means it's a problem with "checking" in "configure".
Anyway, i still get many errors when it starts building, like ```
/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS9.3.sdk/usr/include/sys/cdefs.h:707:2: error: 
      Unsupported architecture
#error Unsupported architecture
```

**Comment by radare** [comment]
why is zlib needed at all?

**Comment by lunixbochs** [comment]
probably used to implement a crc instruction or something

**Comment by radare** [comment]
There's no reason for unicorn to depend on zlib, i will send a pullreq removing this configure-time check and remove the includes because no function from zlib is called in any single file of the project

**Comment by radare** [comment]
https://github.com/unicorn-engine/unicorn/pull/576 <https://github.com/unicorn-engine/unicorn/pull/576>

**Comment by aquynh** [comment]
i merged that PR, and zlib is no longer a dependency.

thanks.

**Comment by Nmzik** [comment]
Thank you, but it doesn't solve any of compiling issues for ARM iOS. I still get errors in "checking" like
 ```
"ERROR: pthread check failed
       Make sure to have the pthread libs and headers installed." 
```
I have no building issues when i compile for mac, so its only happens on iOS. As i said before when i compile test code(in file configure) in Xcode, it successfully compiles.Thats means unicorn sets incorrect path for iOS SDK and Libraries.

stephengroat:
i'm trying to add iOS builds to travis. has anyone figured out a resolution to this so I can try to merge it in?

Qusic:
I am trying to fix it for iOS but seems that it's not only an issue with build configuration. Clang runtime library for iOS doesn't have support for `clear_cache` builtin and there seems no workaround for that. Btw I also don't think that self-modifying code is even allowed for App Store apps.

[...]

PR #651, opened 2016-10-05
Run Unit Test in Travis builds

stephengroat:
`osx` in Travis has `brew` available without `sudo`, easy to install `cmocka`

For `linux` and appveyor, I created a small script to download and compile `cmocka`.

Unit test `test_tb_x86_64_32_imul_Gv_Ev_Ib` now failing on all platforms

TODO: see if `clang` works in `cygwin` for builds, currently following up with cmocka

What about changing the `tests/unit` to pass and adding it to the build process? Maybe `bindings` too?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,7 @@ include config.mk
 include pkgconfig.mk	# package version
 
 LIBNAME = unicorn
+UNAME_S := $(shell uname -s)
 
 GENOBJ = $(shell find qemu/$(1) -name "*.o" 2>/dev/null) $(wildcard qemu/util/*.o) $(wildcard qemu/*.o) $(wildcard qemu/qom/*.o)\
 		 $(wildcard qemu/hw/core/*.o) $(wildcard qemu/qapi/*.o) $(wildcard qemu/qobject/*.o)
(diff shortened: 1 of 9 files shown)


aquynh:
great, please keep doing these experiments, and ping us when you are done! i know it is very tricky & tedious to make this work, especially with appveyor

stephengroat:
that's what beer is for!

stephengroat:
I've started talking with the appveyor folks about the problems with cmake in cygwin (needed for cmocka).

While that's going on, can anyone comment on why the `linux` and `osx` unit tests are failing?

stephengroat:
Last few builds tested python binding, those are also failing the sample `diff`s

danghvu:
I would skip the binding tests from CI for now, there is too often problems of output format mismatch. The tests there are mostly for the binding developer to use as reference.

stephengroat:
Bindings are second now, we can always quickly remove later. current build failures are from `unit/test`

stephengroat:
appveyor and windows now builds and runs tests (failing tests, like all other platforms) for cygwin and cygwin64, both using gcc

stephengroat:
@aquynh i now have the unit test `test_tb_x86_64_32_imul_Gv_Ev_Ib` failing on all platforms in the same place (expect cygwin `clang`, still working on getting that to compile). can you take a look? i'm not 100% sure what's going on there.

aquynh:
Yes, that is an unsolved case, so you can ignore it.

stephengroat:
@aquynh could you push to the branch to disable the test? i'm not 100% sure how to do it!

samples/Makefile
```diff
@@ -20,7 +21,7 @@ SAMPLEDIR = .
 OBJDIR = .
 LIBDIR = ..
 
-CFLAGS += -Wall -I$(INCDIR)
+CFLAGS += -Wall -Werror -Wno-unused-function -I$(INCDIR)
```
aquynh: if there is unused function, is it better to just clean it, rather than disable the warning?

aquynh:
i just have some minor comments. very nice work!

aquynh:
to disable that test_tb_x86_64_32_imul_Gv_Ev_Ib, you can comment out the line with `./test_tb_x86` in `unit/Makefile`

Makefile
```diff
@@ -100,65 +101,63 @@ VERSION_EXT =
 
 BIN_EXT =
 
-IS_APPLE := $(shell $(CC) -dM -E - < /dev/null | grep -cm 1 -e __apple_build_version__ -e __APPLE_CC__)
-ifeq ($(IS_APPLE),1)
+# Apple?
+ifeq ($(UNAME_S),Darwin)
```
stephengroat: Compiling on iOS or cross compiling for iOS on macOS?

samples/Makefile
stephengroat: agreed. cleaned

stephengroat:
new test failing, `test_x86_soft_paging`

aquynh:
yes i remember test_x86_soft_paging is another unsolved case.

stephengroat:
ok, i'll comment that out too.

i'm currently trying to get `macos_universal` building in a reasonable amount of time by caching the `brew install glib --universal` output (was thinking about making the `universal` the default?).

aquynh:
i think in most system glib is not compiled in universal mode, so we cannot compile with universal as default option.

stephengroat:
ok. reading the comments of `test_hang`, I'm guessing that's one to skip too?

stephengroat:
trying to integrate #574 and #647 into the `osx` builds

for #647, would `brew unlink` allow me to install `glib` and `glib --universal` side by side with brew?

Makefile
aquynh: i mean cross compile from Mac, but then it should work.

aquynh:
is this done yet?

stephengroat:
depends on what you're looking for. what tests are you looking to run?

`test_hang` seems to be failing on `clang` for `osx` and `linux` and `gcc` for `windows` and `test_x86_shl_enter_leave` fails for all

should both of those be commented out?

aquynh:
yes please comment them out for now. i will look at them later.

stephengroat:
squashed into a single commit that builds, will do separate PRs for some of the other stuff (mingw, bindings, etc.)

stephengroat:
Any ideas on the failing cross compiles in the latest build?

`pthread` error? seen it before?

aquynh:
are you done with this yet?

stephengroat:
yup. cleaned up commits and ready to push. comments still in place

tests/unit/test_tb_x86.c
```diff
@@ -222,7 +222,7 @@ static void test_tb_x86_64_32_imul_Gv_Ev_Ib(void **state)
 {
     uc_engine *uc = *state;
     uc_hook trace1, trace2;
-    void *mem;
+    //void *mem;
```
danghvu: What is this for ? If we exclude this test, let's leave this out from this commit.

stephengroat: Doesn't compile with it in

danghvu: I don't understand, which one doesn't compile with it and what is the error ?

tests/unit/test_tb_x86.c
```diff
@@ -245,8 +245,9 @@ static void test_tb_x86_64_32_imul_Gv_Ev_Ib(void **state)
     int64_t edi = 0x488ac239;
 #endif
 
-    mem = calloc(1, CODE_SPACE);
-    assert_int_not_equal(0, mem);
+    //mem = calloc(1, CODE_SPACE);
+    // TODO examine
+    //assert_int_not_equal(0, mem);
```
stephengroat: @danghvu referring to your comment on line 255, this assert is the issue. since `mem` is a `void*` and not an `int`, this doesn't compile. with `-Wall -Werror`, you then need to remove an unused variable

aquynh:
this looks great! for the record, can you summarize all the changes of this PR?

stephengroat:
Summary:
- in appveyor, install clang and cmake in cygwin, enable package upgrades, and build cmocka and enable testing for gcc only
- in `gitignore`, ignore generated cmocka folder
- in travis, use brew in osx to install cmocka, and enable testing for gcc and clang on os x and linux
- in `Makefile`, change to use `uname -s` to determine os type
- make `install-cmocka-linux.sh`, a simple shell script to download and install cmocka on linux
- in `bindings/Makefile`, enable `make -c` to call subdirectory makefiles instead of `cd [dir] && make` and include environment variables for runtime access to generated libraries
- in `samples/Makefile`, change to use `uname -s` to determine os type, remove `clean_bins` from `all` command, and include `Werror` for compile strictness
- in `tests/unit/Makefile`, add `cflags` for compile time access to cmocka headers and library, include execute vars for runtime access to cmocka and unicorn libs
- in `tests/unit/test_tb_x86.c`, comment out assert that would not compile 

I can move it to the commit message if necessary

aquynh:
merged, thanks for your fantastic work: https://github.com/unicorn-engine/unicorn/commit/867cd92c406cafe0117272d66c8733e86946fbb6!

stephengroat:
can you try triggering a re-run on travis?

aquynh:
Ok Travis passed now, that was false positive, thanks.

stephengroat:
looking at it

stephengroat:
appveyor is unhappy too

aquynh:
Appveyor had no complain in the latest build.

Issue #560, opened 2016-05-28
memory mapping speed regression


lunixbochs:
If I rapidly uc_mem_map / uc_mem_unmap, I hit a speed regression very quickly.

Here's a test case:

```
#include <unicorn/unicorn.h>
#include <stdio.h>
#include <time.h>

int main(int argc, char **argv) {
    uc_engine *uc;
    uc_open(UC_ARCH_X86, UC_MODE_32, &uc);

    int64_t count = 0;
    int64_t start = time(NULL);
    while (1) {
        count++;
        uc_mem_map(uc, 0x1000, 0x10000, UC_PROT_ALL);
        uc_mem_unmap(uc, 0x1000, 0x10000);
        int64_t now = time(NULL);
        printf("%.4f maps per second\n", count / (double)(now - start));
    }
}
```

lunixbochs:
I've hooked Usercorn up to AFL but both methods of forkserver are causing me problems-

If I reuse a single Unicorn instance but remap its memory every run, I hit this regression within seconds.
If I use a new Unicorn instance every time, memory usage balloons out of control (talking like 1TB virtual memory) very quickly.

aquynh:
interesting, what can be the reason of this speed issue?

lunixbochs:
No idea. I did some basic debugging and didn't find it. I see a slow memory leak as well, not sure if it's related.

lunixbochs:
Also happens with `mem_map_ptr()`

lunixbochs:
Speed increases from ~500 to ~20000 maps per second and the speed regression seems to disappear (albeit with a leak) if I stub out memory_region_unref and delete all the mr->name stuff in memory_region_init

aquynh:
Can Valgrind find this memleak?

Yes, Unicorn can be optimized to remove some dependency layers we dont really need. This is quite a lot of work however.

danghvu:
There are some memory that are not free-ed during uc_mem_unmap but eventually is free-ed during `uc_close` .. I am trying to track these down.

lunixbochs:
this issue is more about the speed than the leak, though both matter when trying to use unicorn as an afl forkserver

danghvu:
I'm more concerning the reduced in speed, which is due to memory leak so far as my investigation

ltfish:
I assume this issue has been fixed. Maybe it's time to close it?

PR #655, opened 2016-10-20
Remove unmapped/freed memory regions from the object property list.

ltfish:
This commit fixes the following issues:
- Any unmapped/free'd memory regions (MemoryRegion instances) are not
  removed from the object property linked list of its owner (which is
  always qdev_get_machine(uc)). This issue makes adding new memory
  mapping by calling mem_map() or mem_map_ptr() slower as more and more
  memory pages are mapped and unmapped - yes, even if those memory pages
  are unmapped, they still impact the speed of future memory page
  mappings due to this issue.
- FlatView is not reconstructed after a memory region is freed during
  unmapping, which leads to a use-after-free the next time a new memory
  region is mapped in address_space_update_topology().


lunixbochs:
I'm going to be so excited if this fixes #560, testing now :)

danghvu:
Great job finding this man.. I was looking at that issue before but couldn't find it. We should merge this ASAP.

aquynh:
merged, thanks!

PR #658, opened 2016-10-25
Fix leak and style in map_write.c

peter-harliman:
Fix potential leak, wrong text in log, and restyle
to a cleaner exit path


aquynh:
merged, thanks!

Issue #654, opened 2016-10-19
bindings: registers are signed values


crowell:
```
In [1]: from unicorn import *

In [2]: from unicorn.x86_const import *

In [3]: mu = Uc(UC_ARCH_X86, UC_MODE_64)

In [4]: mu.reg_write(UC_X86_REG_RIP, 0xffffff8000200000)

In [5]: mu.reg_read(UC_X86_REG_RIP)
Out[5]: -549753716736
```

registers should be unsigned values. at least in python you're using ctypes for `int64`, instead of `uint64`

It doesn't make sense for `$pc` to ever be negative, especially when do things like map kernels there ;)

aquynh:
Can you send a pull req for this?

mrphrazer:
I think that also holds for the memory hooks?

```python

UC_HOOK_MEM_INVALID_CB = ctypes.CFUNCTYPE(
    ctypes.c_bool, uc_engine, ctypes.c_int,
    ctypes.c_uint64, ctypes.c_int, ctypes.c_int64, ctypes.c_void_p
)
UC_HOOK_MEM_ACCESS_CB = ctypes.CFUNCTYPE(
    None, uc_engine, ctypes.c_int,
    ctypes.c_uint64, ctypes.c_int, ctypes.c_int64, ctypes.c_void_p
```

aquynh:
Pull req, please?

lowlyw:
https://github.com/unicorn-engine/unicorn/pull/661

for fix for ruby impl

aquynh:
@crowell: fixed now thanks to @mrphrazer. please report if you still have issues, thanks.

mrphrazer:
we have the same problems in the hooks, `value` can be negative:

```python
    def _hook_mem_invalid_cb(self, handle, access, address, size, value, user_data):
        # call user's callback with self object
        (cb, data) = self._callbacks[user_data]
        return cb(self, access, address, size, value, data)

    def _hook_mem_access_cb(self, handle, access, address, size, value, user_data):
        # call user's callback with self object
        (cb, data) = self._callbacks[user_data]
        cb(self, access, address, size, value, data)
```
A possible fix is
```python
value &= 2 ** (size * 8) - 1
```

aquynh:
i am not sure it is a good idea to do this conversion inside the framework, rather than doing it in your app. is this what everyone wants?

sh1ftchg:
[...]

PR #659, opened 2016-10-25
fixed python bindings signed/unsigned regs/mem

mrphrazer:
seems to fix the [unsigned issues](https://github.com/unicorn-engine/unicorn/issues/654) however, I am not sure if I "fixed" something that was correct, please review before making things worse


bindings/python/unicorn/unicorn.py
```diff
@@ -89,8 +89,8 @@ def _setup_prototype(lib, fname, restype, *argtypes):
 _setup_prototype(_uc, "uc_close", ucerr, uc_engine)
 _setup_prototype(_uc, "uc_strerror", ctypes.c_char_p, ucerr)
 _setup_prototype(_uc, "uc_errno", ucerr, uc_engine)
-_setup_prototype(_uc, "uc_reg_read", ucerr, uc_engine, ctypes.c_int, ctypes.c_void_p)
-_setup_prototype(_uc, "uc_reg_write", ucerr, uc_engine, ctypes.c_int, ctypes.c_void_p)
+_setup_prototype(_uc, "uc_reg_read", ucerr, uc_engine, ctypes.c_uint64, ctypes.c_void_p)
```
aquynh: the original prototype of uc_reg_read() is like this:

    uc_err uc_reg_read(uc_engine *uc, int regid, void *value);

so you cannot use c_uint64 here, but just c_uint.

lowlyw: this is incorrect. regid is just an index.

bindings/python/unicorn/unicorn.py
```diff
@@ -89,8 +89,8 @@ def _setup_prototype(lib, fname, restype, *argtypes):
...
+_setup_prototype(_uc, "uc_reg_write", ucerr, uc_engine, ctypes.c_uint64, ctypes.c_void_p)
```
lowlyw: again, this is just an index. please check the prototypes. they live in `include/unicorn/unicorn.h`

bindings/python/unicorn/unicorn.py
```diff
@@ -114,11 +114,11 @@ def _setup_prototype(lib, fname, restype, *argtypes):
 UC_HOOK_CODE_CB = ctypes.CFUNCTYPE(None, uc_engine, ctypes.c_uint64, ctypes.c_size_t, ctypes.c_void_p)
 UC_HOOK_MEM_INVALID_CB = ctypes.CFUNCTYPE(
     ctypes.c_bool, uc_engine, ctypes.c_int,
```
lowlyw: read the prototypes here 

```c
typedef void (*uc_cb_hookmem_t)(uc_engine *uc, uc_mem_type type,                
        uint64_t address, int size, int64_t value, void *user_data);
```

mrphrazer:
thanks, fixed!

bindings/python/unicorn/unicorn.py
```diff
@@ -114,11 +114,11 @@ def _setup_prototype(lib, fname, restype, *argtypes):
...
-    ctypes.c_uint64, ctypes.c_int, ctypes.c_int64, ctypes.c_void_p
+    ctypes.c_uint64, ctypes.c_uint, ctypes.c_uint64, ctypes.c_void_p
 )
 UC_HOOK_MEM_ACCESS_CB = ctypes.CFUNCTYPE(
     None, uc_engine, ctypes.c_int,
-    ctypes.c_uint64, ctypes.c_int, ctypes.c_int64, ctypes.c_void_p
+    ctypes.c_uint64, ctypes.c_uint, ctypes.c_uint64, ctypes.c_void_p
```
lowlyw: same, consult the prototypes.

bindings/python/unicorn/unicorn.py
```diff
@@ -255,7 +255,7 @@ def reg_read(self, reg_id):
                 return reg.low_qword | (reg.high_qword << 64)
 
         # read to 64bit number to be safe
```
lowlyw: this comment is misleading... can we do something about that?

bindings/python/unicorn/unicorn.py
```diff
@@ -284,7 +284,7 @@ def reg_write(self, reg_id, value):
 
         if reg is None:
             # convert to 64bit number to be safe
```
lowlyw: same. Nguyen?

bindings/python/unicorn/unicorn.py
aquynh: what do you mean, @lowlyw? it is hard to follow this. if you find something wrong, can you open a new issue?

aquynh: you are right, @mrphrazer should not change the API prototype.

aquynh:
merged, but i reverted the changes on the API prototypes, thanks!

PR #661, opened 2016-10-26
fix signedness issue with reg read/write in ruby bindings

lowlyw:
also fixup extra blank spaces at the end of lines.


aquynh:
great, lets see if @sashs has any comments.

lowlyw:
ping?

aquynh:
merged, thanks!

PR #664, opened 2016-10-28
python bindings: unsiged conversion in memory hooks

mrphrazer:
fixes the annotation that is described [here](https://github.com/unicorn-engine/unicorn/issues/654#issuecomment-256640923)


wtdcode:

PR #666, opened 2016-10-29
Make cleanup

stephengroat:
- Cleanup the unit test makefile
- add Mac OS Universal builds
- enable python bindings test
- General Makefile cleanup
- Build regress tests
- Cleanup samples makefile
- Python testing (compare against samples)
- MSYS2/MinGW build (no test/`make test`)
- start building `go` bindings

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,7 @@ before_install:
   - if [[ "$TRAVIS_OS_NAME" == "osx" && "$MACOS_UNIVERSAL" == "yes" ]]; then brew install glib --universal cmocka; fi
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then ./install-cmocka-linux.sh; fi
 script:
-  - make && make test
-# TODO make bindings enabled
-#  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then make clean && ./make.sh ios; fi
+  - make && make -C bindings/go && make -C bindings/go test && make test
 compiler:
   - clang
   - gcc
@@ -23,8 +21,6 @@ matrix:
   - os: osx
     compiler: gcc
     env: MACOS_UNIVERSAL=yes
-#  - os: osx
-#    compiler: x86_64-w64-mingw32-gcc
 addons:
   apt:
     packages:
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,42 +2,30 @@ language: c
 sudo: false
 before_install:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update; fi
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install glib cmocka; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" && "$MACOS_UNIVERSAL" != "yes" ]]; then brew install glib cmocka; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" && "$MACOS_UNIVERSAL" == "yes" ]]; then brew install glib --universal cmocka; fi
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then ./install-cmocka-linux.sh; fi
 script:
-  - if [[ $CC == *"x86_64"* ]]; then ./make.sh cross-win64; elif [[ $CC == *"i686"* ]]; then ./make.sh cross-win32; else ./make.sh && make test; fi
+  - make && make test
 # TODO make bindings enabled
-#  - ./make.sh && make test && make bindings
-# TODO make universal build
-#  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew remove glib && brew install glib --universal && make clean && ./make.sh macos-universal && make test; fi
-# TODO test iOS builds
 #  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then make clean && ./make.sh ios; fi
 compiler:
   - clang
   - gcc
-# TODO update mingw32 to gcc 4.7+ for compilation
-#  - i686-w64-mingw32-gcc
-#  - x86_64-w64-mingw32-gcc
 os:
   - linux
   - osx
-#matrix:
-#  exclude:
-#  - os: osx
-#    compiler: i686-w64-mingw32-gcc
+matrix:
+  include:
+  - os: osx
+    compiler: clang
+    env: MACOS_UNIVERSAL=yes
+  - os: osx
+    compiler: gcc
+    env: MACOS_UNIVERSAL=yes
 #  - os: osx
 #    compiler: x86_64-w64-mingw32-gcc
 addons:
   apt:
     packages:
-    - mingw-w64
-    - gcc-mingw-w64
-    - mingw-w64-dev
-    - gcc-mingw-w64-i686
-    - gcc-mingw-w64-x86-64
-    - binutils-mingw-w64-i686
-    - binutils-mingw-w64-x86-64
-# TODO are mingw32 builds necessary?
-#    - mingw32
-#    - mingw32-binutils
-#    - mingw32-runtime
+    # mingw-w64 packages too old in precise
--- a/Makefile
+++ b/Makefile
@@ -201,25 +201,22 @@ $(LIBNAME)_LDFLAGS += $(GLIB) -lm
 all: unicorn
 	$(MAKE) -C samples
 
-qemu/config-host.h-timestamp config.log:
+qemu/config-host.h-timestamp:
 	cd qemu && \
 	./configure --cc="${CC}" --extra-cflags="$(UNICORN_CFLAGS)" --target-list="$(UNICORN_TARGETS)" ${UNICORN_QEMU_FLAGS}
 	printf "$(UNICORN_ARCHS)" > config.log
-
-compile_lib: qemu/config-host.h-timestamp config.log
 	$(MAKE) -C qemu -j 4
 	$(eval UC_TARGET_OBJ += $$(wildcard qemu/util/*.o) $$(wildcard qemu/*.o) $$(wildcard qemu/qom/*.o) $$(wildcard qemu/hw/core/*.o) $$(wildcard qemu/qapi/*.o) $$(wildcard qemu/qobject/*.o))
 
 unicorn: $(LIBRARY) $(ARCHIVE)
 
-$(LIBRARY): compile_lib uc.o list.o
+$(LIBRARY): qemu/config-host.h-timestamp uc.o list.o
 	$(CC) $(CFLAGS) -shared $(UC_TARGET_OBJ) uc.o list.o -o $(LIBRARY) $($(LIBNAME)_LDFLAGS)
 	-ln -sf $(LIBRARY) $(LIBRARY_SYMLINK)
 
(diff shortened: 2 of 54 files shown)


make.sh
```diff
@@ -36,17 +36,6 @@ build_iOS() {
     ${MAKE}
 }
 
-build() {
```
aquynh: why do you remove these code in make.sh?

.travis.yml
```diff
@@ -2,42 +2,29 @@ language: c
 sudo: false
 before_install:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update; fi
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install glib cmocka; fi
```
aquynh: you removed this because mocka is already there?

make.sh
stephengroat: `UNAME` and `LIBARCHS` weren't being used in the `Makefile` at all. If a `MACOS_UNIVERSAL` build is wanted, wouldn't it be easier for people to run `MACOS_UNIVERSAL=yes make`?

aquynh: yes, but for convenience, can you retain the option "macos-universal"?

samples/Makefile
```diff
@@ -106,73 +104,15 @@ ifneq (,$(findstring m68k,$(UNICORN_ARCHS)))
 SOURCES += sample_m68k.c
 endif
 
-OBJS = $(addprefix $(OBJDIR)/,$(SOURCES:.c=.o))
-OBJS_ELF = $(addprefix $(OBJDIR)/,$(SOURCES:.c=))
-BINARY = $(addprefix $(SAMPLEDIR)/,$(SOURCES:.c=$(BIN_EXT)))
+OBJS = $(SOURCES:.c=.o)
+OBJS_ELF =$(SOURCES:.c=)
```
aquynh: is there a better name? for ex, Mac does not use ELF format

aquynh:
nice work, let me know when all the issues are solved, thanks.

samples/Makefile
stephengroat: changes to `BINS`

make.sh
stephengroat: done

make.sh
```diff
@@ -112,17 +101,13 @@ fi
 export CC INSTALL_BIN PREFIX PKGCFGDIR LIBDIRARCH LIBARCHS CFLAGS LDFLAGS
 
 case "$1" in
-  "" ) build;;
```
stephengroat: updated

Makefile
```diff
@@ -228,43 +214,21 @@ else
...
-	cd qemu && $(MAKE) -j 4
-	$(MAKE) unicorn
+	$(MAKE) -C qemu -j 4
 
-unicorn: $(LIBRARY) $(ARCHIVE)
+unicorn: compile_lib $(LIBRARY) $(ARCHIVE)
 
-$(LIBRARY): $(UC_TARGET_OBJ) uc.o list.o
-ifeq ($(UNICORN_SHARED),yes)
-ifeq ($(V),0)
```
aquynh: why do you remove this verbose option here?

samples/Makefile
```diff
@@ -106,73 +69,9 @@ ifneq (,$(findstring m68k,$(UNICORN_ARCHS)))
...
-	rm -rf libunicorn*.so libunicorn*.lib libunicorn*.dylib unicorn*.dll unicorn*.lib
-
-clean: clean_bins clean_libs
-
-$(BINARY): $(OBJS)
-
-$(SAMPLEDIR)/%$(BIN_EXT): $(OBJDIR)/%.o
-	@mkdir -p $(@D)
-ifeq ($(V),0)
-ifeq ($(UNICORN_SHARED),yes)
```
aquynh: can you keep this verbose option, too?

stephengroat: implicit rules are default verbose

aquynh: this is before this change, "make" printed out command with just "CC" & "LINK", without details. now linker prints out a huge command for "libunicorn.1.dylib". also for samples, the output is like:

    cc -Wall -Werror -I../include  -L..   sample_arm.c  -lpthread -lunicorn -lm -L/usr/local/Cellar/glib/2.50.1/lib -L/usr/local/opt/gettext/lib -lglib-2.0 -lintl -o sample_arm

Makefile
stephengroat: I'm actually still trying to move over to autotools. since autotools handles verbosity, i think it'd be a better option (https://autotools.io/automake/silent.html)

aquynh:
Oh no, we want to keep it simple, and can do without autotool. Just few
more lines in Makefile, and i dont see why we need that.

stephengroat:
@aquynh autotools is built to deal with system specific configurations that are currently in the Makefiles. autotools would `make` (pun intended) the build system more legible and easier to debug (i think). it would also follow gnu conventions, since just about every system out there is built with `./configure && make && make install`

Makefile
```diff
@@ -141,27 +139,25 @@ endif
 
 ifeq ($(UNICORN_SHARED),yes)
 ifneq ($(filter MINGW%,$(UNAME_S)),)
-LIBRARY = $(BLDIR)/$(LIBNAME).$(EXT)
+LIBRARY = $(LIBNAME).$(EXT)
 else ifneq ($(filter CYGWIN%,$(UNAME_S)),)
-LIBRARY = $(BLDIR)/cyg$(LIBNAME).$(EXT)
```
stephengroat: do you want to separate out the cygwin builds and other win builds? if so, the python needs to be updated to recognize the `cyg` prefix

aquynh: what do you mean by "separate out"?

stephengroat: sorry about the ambiguity. by "seperate out", I mean do you want cygwin dll to have the `cyg` prefix and for other dll not to?

I can understand if you do so that cygwin dll can be installed side by side with mingw dlls

aquynh: that "cyg" prefix was proposed by @iroiro123 (see https://github.com/unicorn-engine/unicorn/pull/323), who wanted that because "cyg" is a standard prefix for Cygwin DLL.

you are right that if we keep this, then yes Python must be changed to support this prefix. also, i am not really sure if the dependencies (supported DLLs) are the same, as with Msys.

stephengroat: looks like the `cyg` prefix is good to differentiate between cygwin and mingw. i'll change the makefile back and add to python

samples/Makefile
stephengroat: i don't think there's any reason to compile and link in two separate steps for what's being done here.

also, there's a huge advantage following GNU standards and using implicit rules. it helps with library and flag ordering, etc

stephengroat:
@aquynh take a look over, it's pretty good right now. there's quite a bit of work in here

aquynh:
fantastic work again! let me know when you are done

stephengroat:
i'm going to make one last mingw attempt tonight. i'll update after
tonight. if there's any sign of success, i'll ask to hold on before a
merge. otherwise, i'll just say merge and i'll figure out mingw builds later

bindings/python/unicorn/unicorn.py
```diff
@@ -17,12 +17,11 @@
...
-    _lib = "libunicorn.dylib"
-elif sys.platform in ('win32', 'cygwin'):
-    _lib = "unicorn.dll"
-else:
-    _lib = "libunicorn.so"
+_lib = { 'darwin': 'libunicorn.dylib',
+         'win32': 'unicorn.dll',
+         'cygwin': 'cygunicorn.dll',
+         'linux': 'libunicorn.so',
+         'linux2': 'libunicorn.so' }
```
stephengroat: `linux2` is for 2.x and 3.x kernels, a pythonism

stephengroat:
i'm getting SUPER close on the msys build, currently struggling to get cmocka built in msys. worth the work or just start with `make` only?

aquynh:
if you cannot compile cmocka in Msys, then ignore it. that not important i guess.

stephengroat:
headed to bed. getting a lot closer, don't merge yet. should be able to get `cmocka` to build, will just take a little bit of fanagaling

stephengroat:
@aquynh ready to merge. the msys2/`mingw` build works, just need to get the test in. since this commit has gotten huge, i'd rather get this merged and continue working on the msys issues in another PR

Makefile
radare: Cmake also handles verbosity. But im not a cmake or autotools fan at all. If you do custom makefiles and just want a configure step you may try acr instead. Its equally portable and way smaller and faster

radare: Linux and r2 doesnt use autotools and also support verbose builds

stephengroat: if you look at autotools, this is specifically what it was designed to handle. instead of modifying makefiles, the platform dependencies are put into the autotools tools (https://www.gnu.org/software/automake/manual/automake.html#GNU-Build-System)

aquynh:
is this ready to merge, or you still want to make some changes?

stephengroat:
@aquynh should be good now

bindings/go/unicorn/unicorn.go
```diff
@@ -7,8 +7,8 @@ import (
 )
 
 /*
-#cgo CFLAGS: -O3
-#cgo LDFLAGS: -lunicorn
+#cgo CFLAGS: -O3 -Wall -Werror -I../../../include
```
aquynh: @lunixbochs, this changes your Go binding, would you approve?

bindings/python/sample_arm.py
```diff
@@ -46,7 +46,7 @@ def test_arm():
         mu.hook_add(UC_HOOK_BLOCK, hook_block)
 
         # tracing all instructions with customized callback
-        mu.hook_add(UC_HOOK_CODE, hook_code)
+        mu.hook_add(UC_HOOK_CODE, hook_code, begin=ADDRESS, end=ADDRESS)
```
aquynh: why do you want these default params?

bindings/python/sample_x86.py
```diff
@@ -128,21 +135,21 @@ def test_i386():
...
         print("")
 
     except UcError as e:
         print("ERROR: %s" % e)
 
 
-def test_i386_loop():
-    print("Emulate i386 code with infinite loop - wait for 2 seconds then stop emulation")
+def test_i386_map_ptr():
+    print("Emulate i386 code - use uc_mem_map_ptr()")
```
aquynh: this is confused: i dont see where uc_mem_map_ptr() is used here (this API uc_mem_map_ptr() is supposed for C)

samples/Makefile
aquynh: can you please recover this part, so default option (quiet) still works?
same applies for the Makefile in root dir, too.

aquynh:
another fantastic work, merged, thanks!

i reverted the change on verbose option, however. and sorry but i think we are fine without autotools.

stephengroat:
i think that broke something in appveyor (the new msys builds). let me see
if i can fix it

aquynh:
i already fixed it.

lunixbochs:
This PR breaks the Go bindings in multiple ways.

aquynh:
please send in PR to fix that, thanks

PR #670, opened 2016-10-31
Update COMPILE-WINDOWS.md

CryptXor:


aquynh:
What do you think about putting this option into Makefile instead, at Mingw section?

aquynh:
see my latest commit https://github.com/unicorn-engine/unicorn/commit/9e5b9d746cc9253154ba001a882007b0cfc3f984. with this, now you dont need to pass the option in at compile time, thanks.

PR #672, opened 2016-11-01
Add fpu STx support

feliam:
In case it works. User must set TOP accordingly before using this.


aquynh:
cool, could you add a new sample code to samples/ for this new feature?

aquynh:
"In case it works. User must set TOP accordingly before using this."  <-- this is why we need a sample code, thanks.

aquynh:
we are on the way towards 1.0 release, and would be nice to have this available. can you supply the sample code, so i can test & merge this?

PR #673, opened 2016-11-01
changed '%02hhx' to '%02x' in sample_x86_32_gdt_and_seg_regs.c

chubbymaggie:
This PR is to fix the following error that occurred while compiling unicorn-engine using MinGW32
unicorn_error


aquynh:
merged, thanks!

Issue #671, opened 2016-10-31
Python 3.5 dynamic library load error


CryptXor:
I've been battling with running the python bindings on python 3.5.

I keep getting a `ImportError: ERROR: fail to load the dynamic library.` however I have checked that all libraries are present in .\unicorn\lib

```
    Directory: C:\Python\Lib\site-packages\unicorn\lib


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----       2016/09/24   7:23 PM         122036 libgcc_s_dw2-1.dll
-a----       2016/09/24   3:17 PM          83742 libgcc_s_seh-1.dll
-a----       2016/10/13   7:02 PM        1124259 libglib-2.0-0.dll
-a----       2015/06/29   1:25 PM        1050543 libiconv-2.dll
-a----       2016/03/31   8:50 AM         132717 libintl-8.dll
-a----       2016/10/09   5:46 PM          56978 libwinpthread-1.dll
-a----       2016/10/31   9:10 PM       26335738 unicorn.dll
```

The library does load the directories correctly
```
[-] path: 'C:\Python\lib\site-packages\unicorn\lib'
        [+] lib_file: C:\Python\lib\site-packages\unicorn\lib\libwinpthread-1.dll
        [+] lib_file: C:\Python\lib\site-packages\unicorn\lib\libgcc_s_seh-1.dll
        [+] lib_file: C:\Python\lib\site-packages\unicorn\lib\libgcc_s_dw2-1.dll
[-] path: 'C:\Python\lib\site-packages\unicorn\lib'
        [+] lib_file: C:\Python\lib\site-packages\unicorn\lib\libwinpthread-1.dll
        [+] lib_file: C:\Python\lib\site-packages\unicorn\lib\libgcc_s_seh-1.dll
        [+] lib_file: C:\Python\lib\site-packages\unicorn\lib\libgcc_s_dw2-1.dll
[-] path: ''
        [+] lib_file: libwinpthread-1.dll
        [+] lib_file: libgcc_s_seh-1.dll
        [+] lib_file: libgcc_s_dw2-1.dll
        [+] lib_file: libiconv-2.dll
        [+] lib_file: libintl-8.dll
        [+] lib_file: libglib-2.0-0.dll
[-] path: 'C:\Python\Lib\site-packages'
        [+] lib_file: C:\Python\Lib\site-packages\libwinpthread-1.dll
[...]

PR #675, opened 2016-11-05
Fixes to make python distribution for windows work

rhelmot:
I was successfully able to build and install working wheels for both mingw64 and mingw32.

Note that there's actually a bug in the msys distribution of python preventing bdist_wheel from working correctly. I submitted a patch to address this. https://github.com/Alexpux/MINGW-packages/pull/1899

Should address #671

### DIFF
--- a/bindings/python/setup.cfg
+++ b/bindings/python/setup.cfg
@@ -1,0 +1,2 @@
+[bdist_wheel]
+universal=1
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -4,30 +4,32 @@
 from __future__ import print_function
 import glob
 import os
+import subprocess
 import shutil
 import sys
+import platform
 
 from distutils import log
 from distutils.core import setup
+from distutils.util import get_platform
 from distutils.command.build import build
 from distutils.command.sdist import sdist
 from setuptools.command.bdist_egg import bdist_egg
 
-# prebuilt libraries for Windows - for sdist
-PATH_LIB64 = "prebuilt/win64/unicorn.dll"
-PATH_LIB32 = "prebuilt/win32/unicorn.dll"
-
-# package name can be 'unicorn' or 'unicorn-windows'
-PKG_NAME = 'unicorn'
-if os.path.exists(PATH_LIB64) and os.path.exists(PATH_LIB32):
-    PKG_NAME = 'unicorn-windows'
-
 SYSTEM = sys.platform
 VERSION = '1.0'
 
-# adapted from commit e504b81 of Nguyen Tan Cong
-# Reference: https://docs.python.org/2/library/platform.html#cross-platform
-IS_64BITS = sys.maxsize > 2**32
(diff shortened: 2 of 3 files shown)


aquynh:
some concerns about this approach:

- it seems users are forced to use Msys environment to install, and from my experience Msys & its Python is very slow, and buggy
- this always recompiles unicorn.dll

i still think that it is simpler if we ask users to explicitly copy all the required DLLs (including dependencies & `unicorn.dll`) into `prebuilt/` directory (which will be picked up to put into the same final directory with `unicorn.py`), then run `python setup.py install` from whatever they want.

thoughts?

rhelmot:
That's the thing, it packages the full set of dependencies! You can run a
binary distribution of this package on pure windows Python; I tested this.

I'm seeing the incremental builds work correctly...

aquynh:
to run this on native Windows Python, you have to copy dependencies to `c:\python35\bin`, which is not a nice approach. it is easier if we just put all the DLLS into `prebuilt/`, and get setup.py pick them up from there.

aquynh:
Or we can keep what this PR is doing, but add prebuilt/ support. So users can either run this in Msys (uses Msys DLLs) or from native Python (uses prebuilt DLLs).

But lets make the core compilation non-default, so users without Msys can still run it

rhelmot:
FYI: I have not forgotten this, but I'm currently overloaded with work and will be busy an traveling for the next week.

stephengroat:
after working a lot with the build system, i'm not sure this is the best idea.

python is a consumer of the C library, it seems a little bit backwards to have python create a dll. also, if a windows solution is put into place, a *nix and darwin solution should also probably be added

rhelmot:
Two responses:

- unfortunately, this library has to be responsible for its own C library, since without it it just doesn't work. Capstone, for example, used to be distributed without its C library, and every other issue I got on a tool that depended on capstone was "it doesn't work" which was because capstone refused to import because it got installed without its c library. 

- we already do support doing this for win32 and Darwin! This patch is an update to an existing system I designed to do this exactly.

rhelmot:
I pushed a thing so that people can copy the windows support DLLs into bindings/python/prebuilt and they'll be packaged up with the install.

aquynh:
great, but i still see the call to `subprocess.call(cmd, env=new_env)`, so the core is always compiled in any cases?

rhelmot:
If you don't want to compile the core, you shouldn't be using a source distribution - I specifically set up everything so binary distributions (`python setup.py bdist_wheel register upload` on the developer end, `pip install unicorn` on the user end) work correctly.

aquynh:
how about this scenario: Windows users copy the dependency libraries to prebuilt/, then run "python setup.py install". this would fail, because of this compilation requires Msys, which most users do not have on their machines.

rhelmot:
I'm not sure why you're insistent on supporting the scenario where a user who does not have a build environment set up uses anything other than a packaged binary distribution, but alright. This should support that.

aquynh:
merged, thanks!

PR #680, opened 2016-11-19
OpenBSD_Notes_re_DEP: Added an md file in docs/ ...

oblivia-simplex:
explaining how to disable W^X protections on the filesystem being used to run or test Unicorn applications, so that the OS will permit those applications to run.


aquynh:
merged, thanks!

Issue #652, opened 2016-10-08
ARM: read / write coprocessor via API


mothran:
I am currently trying to emulate a usermode ARM process and ran into some issues with the Android TLS system, in order to get thread locals such as thread ID and other components, the process attempts to read the coprocessor registers, example:

```
mrc     p15, #0, r2, c13, c0, #3
```

Which reads the Read only coprocessor for thread locals (http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.ddi0344k/Babeihid.html)

Tracking it down in QEMU I found that this specific variable SHOULD be the referenced registers:
https://github.com/unicorn-engine/unicorn/blob/master/qemu/target-arm/cpu.h#L215

Along side that should be the other coprocessor registers as well. 

I was going to start attempting to include these coprocessors into the uc_arm_reg structure and implement read / write functions for the registers I needed.  But first I wanted to open this issue and request feedback about naming convention for the coprocessor registers and if there is and better way I should implement this feature. 

I wanted to provide a more light weight method than introducing a injected code sub as per these two tickets: #446 and #571

Thanks.

aquynh:
yes, please go ahead to support these registers, and send pull req when you are done.

there are many registers inside cp15. perhaps we should expose this struct, and perform read/write to the whole struct?

mothran:
It would make the registers much cleaner in that case.  Would you expose it as a member of the uc_engine handle, or as a new API method?

Also the actual variable names inside of the cp15 struct are a little obtuse, if we implement them as individual registers it could a much more clear translation for the unicorn user.  For example:

```
mrc     p15, #0, r2, c13, c0, #3
```
Becomes (or something similar):
```
UC_ARM_REG_CP13_C0_3
```
vs:
```
cp15.tpidrro_el0
```
What are your thoughts?

aquynh:
I think we can do both.

mothran:
Great, do you have any examples of exposing structs like that via the API that I can try and model my code after?

aquynh:
you can copy the struct `cp15` to include/unicorn/arm.h, then add code so that we can read/write  to the whole struct with new register `UC_ARM_REG_CP15`. 

we did something similar for XMM register before, you can take a look at this: https://github.com/unicorn-engine/unicorn/commit/4a8f52ae7f24e9dbbf05d792a91825cdffe7b54d

jbcayrou:
Hi,

[...]

Issue #446, opened 2016-02-29
Unicord fails to emulate VMOV on ARM (actually Thumb)


mkravchik:
Hi,

I have ARMv7 code which Unicorn fails to emulate. Specifically, the code looks like this in IDA:
__text:00012DB2 C0 EF 10 00                 VMOV.I32        D16, #0 ; Vector Move

Unicorn fails with error:
ERROR: Invalid instruction (UC_ERR_INSN_INVALID)

Are vector instructions supported? What CPU architecture unicorn is based on for ARM? Is there any way to control it?

Thanks!

aquynh:
yes, VMOV should be supported. could you submit a testcase (put it under `tests/unit/` or `tests/regress/`) for this case, so we can confirm the issue & test the fix?

mkravchik:
Wow thanks for such a quick response.
I've never submitted a test case, how do I do that?
The following code fails on emulator but works on a real 32 bit iPhone 5c device

from __future__ import print_function
from unicorn import *
from unicorn.arm_const import *

# memory address where emulation starts
ADDRESS    = 0x10000

code = "\xC0\xEF\x10\x00" #                 VMOV.I32        D16, #0 ; Vector Move
# Initialize emulator in thumb mode
mu = Uc(UC_ARCH_ARM, UC_MODE_THUMB)

# map 2MB memory for this emulation
mu.mem_map(ADDRESS, 2 * 1024 * 1024)

# write machine code to be emulated to memory
mu.mem_write(ADDRESS, code)

# emulate machine code in infinite time
mu.emu_start(ADDRESS, ADDRESS + len(code))

aquynh:
this means you should send a pull request to put your testcase (a Python script) under directory `tests/regress`.

you can take this PR as an example: https://github.com/unicorn-engine/unicorn/pull/434.
it script is at https://github.com/unicorn-engine/unicorn/blob/master/tests/regress/jumping.py

mkravchik:
I just did. #447

aquynh:
anyway, looks like you are missing a step here: you should enable VFP first before running this instruction, no?

according to http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.ddi0449b/ch02s01s02.html, we have "... From reset, the VFP extension is disabled. Any attempt to execute a VFP instruction results in an Undefined Instruction exception being taken."

see above link for how to modify your code for VFP, so this sample can work.

mkravchik:
Hi,

[...]

PR #684, opened 2016-12-05
Added ARM coproc registers

mothran:
As per #652 I added some of the arm coprocessor register.  I did not map them all but this does address the two issues raised previously (#446 and #571) and allows for user to enable the VFP without starting the emu first.


aquynh:
is there any existing testcase that failed before, but passes now thanks to this update?

mothran:
No specific tests, I referenced the previous issues because the solution for those was to run a small set of instructions to enable features in the emulator.  I wanted to be able to directly access those registers instead, thus speeding up my initialization time.

For example I can enable the ARM VFP by simply doing:
```C
// Enable the CP registers for the VFP.
err = uc_reg_read(m_unicorn, UC_ARM_REG_C1_C0_2, &tmp_val);
if (err) {
	break;
}
tmp_val = tmp_val | (0xf << 20);
err = uc_reg_write(m_unicorn, UC_ARM_REG_C1_C0_2, &tmp_val);
if (err) {
	break;
}

size_t enable_vfp = 0x40000000;
err = uc_reg_write(m_unicorn, UC_ARM_REG_FPEXC, &enable_vfp);
if (err) {
	break;
}
```

I also needed write access to the ARM coprocessor that stores the TLS data for android code to run smoothly.

aquynh:
looks good, would you put the script above into suite/regress/, as an example?

aquynh:
ping ...

mothran:
Updated my branch to be in sync with master and added a regression test.

aquynh:
looks good now, but my only concern is that: by directly writing to VFP registers, you may skip some internal CPU setup that is performed by "official" initialization method, which can affect CPU operations later on. is it possible for this case?

mothran:
Possibly but my thought here is that the VFP is only a small subset of the co-processor registers.  I added the ones I needed currently but wanted to at least leave it as a model for people to be able to add more as they go.

As for VFP setup, I would argue that from the stand point of a emulator API it is much less clean to setup some CPU features using injected instructions vs a API call.

I will do some more digging and make sure this is the appropriate setup path for the VFP and update the test to match the setup setup.

aquynh:
Never doubt this is useful, just worry if it leaves CPU in a broken state. Let me know what you find.

hedger:
I'd love to see this in master. Waiting for merge

include/unicorn/arm.h
```diff
@@ -24,6 +24,9 @@ typedef enum uc_arm_reg {
 	UC_ARM_REG_FPSCR_NZCV,
 	UC_ARM_REG_FPSID,
 	UC_ARM_REG_ITSTATE,
+	UC_ARM_REG_C1_C0_2,
+	UC_ARM_REG_C13_C0_2,
```
aquynh: did you add this `UC_ARM_REG_C13_C0_2`, but never used that in `qemu/target-arm/unicorn_arm.c` ?

mothran:
Looks like I did, would you like me to remove it? or add an implementation?  I currently have no use for that register but I can add it for posterity if needed.

aquynh:
Yes please add it, together with a sample for it. I will merge this then.

mothran:
Great, I will update my branch soon.

As per my comment earlier, I dug into the ARM documents and found references for enabling the VFPU:
http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.ddi0439b/BEHBJHIG.html
and FPEXC:
http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0068b/BCFHHBDD.html

So I think the example setup should match official method for enabling the VFPU.

aquynh:
merged this for RC3 release, but please follow up with the sample for `UC_ARM_REG_C13_C0_2`, thanks.

PR #690, opened 2016-12-18
remove glib dependency by providing compatible replacements

cseagle:
Added glib compatible replacements for all glib functions used in unicorn. Removed glib dependencies in Makefiles.
I threw the list and hash table implementations together and have paid no attention to their performance. Feel free to replace the hash table implementation with something that performs better.
Need help testing. All samples and unit tests run with same results as master on the following platforms:
win32 (using mingw), OS X 64, Linux 32, Linux 64.
Have not done much testing beyond that.

### DIFF
--- a/qemu/glib_compat.c
+++ b/qemu/glib_compat.c
@@ -561,7 +561,8 @@ guint g_hash_table_size(GHashTable *hash_table)
 
 /* general g_XXX substitutes */
 
-void g_free(void *ptr) {
+void g_free(void *ptr)
+{
    free(ptr);
 }
 
--- a/qemu/glib_compat.c
+++ b/qemu/glib_compat.c
@@ -561,6 +561,10 @@ guint g_hash_table_size(GHashTable *hash_table)
 
 /* general g_XXX substitutes */
 
+void g_free(void *ptr) {
+   free(ptr);
+}
+
 void *g_malloc(size_t size)
 {
    if (size == 0) return NULL;
--- a/qemu/glib_compat.c
+++ b/qemu/glib_compat.c
@@ -27,32 +27,16 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 
 #include "glib_compat.h"
 
-#ifdef _WIN32
-   #ifdef _WIN64
-      #define __HAVE_64_BIT_PTRS
-   #else
-      #define __HAVE_32_BIT_PTRS
-   #endif
-#else
-   #ifdef _WIN64
-      #define __HAVE_64_BIT_PTRS
-   #endif
+#undef __HAVE_64_BIT_PTRS
+
+#ifdef _WIN64
+    #define __HAVE_64_BIT_PTRS
 #endif
 
 #ifdef __GNUC__
-   #ifdef __x86_64__
-      #define __HAVE_64_BIT_PTRS
-   #else
-      #ifdef __ppc64__
-         #define __HAVE_64_BIT_PTRS
-      #else
-         #ifdef __aarch64__
-            #define __HAVE_64_BIT_PTRS
-         #else
-            #define __HAVE_32_BIT_PTRS
-         #endif
-      #endif
-   #endif
+#if defined(__x86_64__) || defined(__ppc64__) || defined(__aarch64__)
+    #define __HAVE_64_BIT_PTRS
+#endif
 #endif
 
 /* All functions below added to eliminate GLIB dependency */
@@ -65,7 +49,7 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
    This may be marginally better than what glib does in their direct_hash
    but someone with some chops in this space should fix if it needs improving
 */
-guint g_direct_hash(const void *v)
+guint g_direct_hash(gconstpointer v)
 {
 #ifdef __HAVE_64_BIT_PTRS
    uint64_t hash = (uint64_t)v;
@@ -79,16 +63,11 @@ guint g_direct_hash(const void *v)
 #endif
 }
 
-int g_direct_equal(const void *v1, const void *v2)
-{
-   return v1 == v2;
-}
-
 /*
    djb2+ string hashing
    see: http://www.cse.yorku.ca/~oz/hash.html
 */
-guint g_str_hash(const void *v)
+guint g_str_hash(gconstpointer v)
 {
    const char *s = (const char*)v;
    guint hash = 5381;
@@ -99,7 +78,7 @@ guint g_str_hash(const void *v)
    return hash;
 }
 
-int g_str_equal(const void *v1, const void *v2)
+gboolean g_str_equal(gconstpointer v1, gconstpointer v2)
 {
    return strcmp((const char*)v1, (const char*)v2) == 0;
 }
@@ -108,7 +87,7 @@ int g_str_equal(const void *v1, const void *v2)
   Bob Jenkins integer hash algorithm
   see: http://burtleburtle.net/bob/hash/integer.html
 */
-guint g_int_hash(const void *v)
+guint g_int_hash(gconstpointer v)
 {
    guint hash = *(const guint*)v;
    hash = (hash + 0x7ed55d16) + (hash << 12);
@@ -120,7 +99,7 @@ guint g_int_hash(const void *v)
    return hash;
 }
 
-int g_int_equal(const void *v1, const void *v2)
+gboolean g_int_equal(gconstpointer v1, gconstpointer v2)
 {
    return *(const int*)v1 == *(const int*)v2;
 }
@@ -134,7 +113,7 @@ GList *g_list_first(GList *list)
    return list;
 }
 
-void g_list_foreach(GList *list, GFunc func, void* user_data)
+void g_list_foreach(GList *list, GFunc func, gpointer user_data)
 {
    GList *lp;
    for (lp = list; lp; lp = lp->next) {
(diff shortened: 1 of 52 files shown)


aquynh:
very nice! did you write this glib compatible code from scratch?

cseagle:
Yes, it's all from scratch.

cseagle:
I did use a couple of public hash algorithms.

aquynh:
could you point out the source?

what do you think about just reusing glibc code instead, so we dont need to review it?
looks like the related code in glibc is self-contain, so we can just copy it, with some simple modification to make it work for us.

cseagle:
The source of one hash is already noted as djb2. It's essentially the same hash that glib uses as it turns out with one minor change. The integer hash is from Bob Jenkins here: http://burtleburtle.net/bob/hash/integer.html.

If you want to copy from glib, that's fine. It will be more code with more dependencies that either need to be taken in their entirety or trimmed down. It will include a lot of code that is never used by unicorn/qemu. Someone will need to work out all of the inter-dependencies with the glib bits. For example, their hash table implementation and perhaps other things too make use of a custom memory allocator (gslice) that would need to be pulled in as well.

qemu/glib_compat.c
```diff
@@ -0,0 +1,707 @@
...
+   return hash;
+}
+
+int g_int_equal(const void *v1, const void *v2) {
+   return *(const int*)v1 == *(const int*)v2;
+}
+
+/* Doubly-linked list */
+
+GList *g_list_first(GList *list) {
```
danghvu: function style
```
{
}
```

qemu/glib_compat.c
```diff
@@ -0,0 +1,707 @@
...
+GList *g_list_prepend(GList *list, void* data) {
+   GList *n = (GList*)g_malloc(sizeof(GList));
+   n->next = list;
+   n->prev = NULL;
+   n->data = data;
+   return n;
+}
+
+GList *g_list_remove_link(GList *list, GList *llink) {
+   if (llink == list) list = list->next;
```
danghvu: `if (llink)` ?

aquynh:
now this cannot compile on Linux/GCC: https://travis-ci.org/unicorn-engine/unicorn/jobs/185052000

cseagle:
Your comment is useless. That's the prototype straight from glib. Take it up with them.

cseagle:
It compiles just fine for me. clock_gettime is a linux system call, are you telling me it is not present on the travis build system?

cseagle:
Feel free to change it.

aquynh:
i merged this into a new branch "noglib" for now.

lunixbochs:
Does QEMU build using _GNU_SOURCE defined? Some of the libc headers are gated under that, and glib would definitely have it defined.

aquynh:
https://travis-ci.org/unicorn-engine/unicorn/jobs/185133337: for some weird reason, Travis cannot find clock_gettime(), while it did before this patch. @stephengroat, do you have any ideas?

qemu/target-arm/helper.c
```diff
@@ -155,7 +155,7 @@ static void count_cpreg(gpointer key, gpointer opaque)
     }
 }
 
-static gint cpreg_key_compare(gconstpointer a, gconstpointer b)
+static int32_t cpreg_key_compare(gconstpointer a, gconstpointer b)
```
aquynh: i think `gint` is `int`, but not int32_t ? see https://developer.gnome.org/glib/stable/glib-Basic-Types.html#gint

stephengroat:
try adding `-lrt` to the `samples` Makefile library list and see if that helps. was probably included through glib before

stephengroat:
same yourself some build time, replace your travis file in your branch with this:

````
language: c
sudo: false
before_install:
  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update; fi
  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install cmocka; fi
  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then ./install-cmocka-linux.sh; fi
script:
  - make && make -C bindings/go && make -C bindings/go test && make test
compiler:
  - clang
  - gcc
os:
  - linux
  - osx
addons:
  apt:
    packages:
    # mingw-w64 packages too old in precise
````

File: qemu/include/glib_compat.h [hunk 4]
```diff
@@ -104,7 +103,7 @@ void g_hash_table_remove_all(GHashTable *hash_table);
 int g_hash_table_remove(GHashTable *hash_table, const void* key);
 void g_hash_table_unref(GHashTable *hash_table);
 GHashTable *g_hash_table_ref(GHashTable *hash_table);
-uint32_t g_hash_table_size(GHashTable *hash_table);
+guint g_hash_table_size(GHashTable *hash_table);
```
radare: guint is not always 32 bit

docs/COMPILE-NIX.md
```diff
@@ -8,24 +8,10 @@ To compile for Microsoft Windows, see [COMPILE-WINDOWS.md](COMPILE-WINDOWS.md)
...
-  64-bit), you need to have glib in universal format. To do that, build &
-  install glib with:
-
-        $ brew install glib --universal
-
-- For Linux, "glib2-dev" is needed.
-  Ubuntu/Debian users can install this with:
-
-        $ sudo apt-get install libglib2.0-dev
+For MacOS, Unicorn requires "pkg-config" package.
```
radare: Isnt pkgconfig required only for glib?

aquynh: pkgconfig is checked in qemu/configure, but looks like we dont need that anymore. i will remove this dependency.

aquynh:
@stephengroat, now `MACOS_UNIVERSAL=yes` is default, but you suggested to remove that `brew install glib --universal cmocka` in Travis setup, why?

aquynh:
@cseagle, i merged your latest commits to "noglib" branch now.

aquynh:
@stephengroat, adding -lrt does not help: https://travis-ci.org/unicorn-engine/unicorn/jobs/185361656.

looks like this only happens with older gcc. i dont have this issue on Ubuntu 16.04

stephengroat:
@aquynh you only use `LDFLAGS` in your compile step since you reversed my default compile changes 😄 

add the `-lrt` to the `LDFLAGS` line and you'll have more luck!

aquynh:
adding -lrt to LDFLAGS causes error on MacOS, like below:

```
  LINK    sample_arm
ld: library not found for -lrt
```

you are right, i just removed glib from Travis & Appveryor setup.

qemu/glib_compat.c
```diff
@@ -0,0 +1,761 @@
...
+
+#include <string.h>
+#include <stdlib.h>
+#include <stdio.h>
+
+#include "glib_compat.h"
+
+#ifdef _WIN32
+   #ifdef _WIN64
+      #define __HAVE_64_BIT_PTRS
```
aquynh: do we need these macros __HAVE_64_BIT_PTRS & __HAVE_32_BIT_PTRS? i cant find where they are used

aquynh: oops sorry, i missed its use in the hash function below.

aquynh:
now Travis can build Linux samples by adding `-lrt` to link flags: https://github.com/unicorn-engine/unicorn/commit/b34109fb0d64ba4b6b3f45f96afd391f81553999.

@lunixbochs, but it is Go's turn now, can you take a look: https://travis-ci.org/unicorn-engine/unicorn/jobs/186505305 ? i think Go also needs `-lrt`, but not sure where to put that.

lunixbochs:
LDFLAGS in a cgo comment. -lunicorn is in there somewhere. but it might need a linux specific tag? I don't know which platforms need it.

aquynh:
do you mean something like this https://github.com/unicorn-engine/unicorn/commit/af4c29c57fcc60c2fc7ffec5c799d5b799a6a0fe ?

it seems only Linux needs this flag `-lrt`.

lunixbochs:
Exactly that, but with a linux tag. I think like this: "#cgo linux LDFLAGS"

While we're at it that include is wrong. uc.h is the private header right? It should probably use <unicorn/unicorn.h> and fix that -I../ to be /include

aquynh:
to be sure (so that i will not make a mistake), can you send a PR here?

lunixbochs:
not in a good place to do that

aquynh:
@cseagle, i made some minor fixes to make the glib_compat API better compatible with the original glib API: https://github.com/unicorn-engine/unicorn/commit/6b57da1fc2122376bb0c9439c2908987b62c9a15

cseagle:
Ok thanks.

aquynh:
i am not really sure about the performance of this new glib_compat code, so i lifted some of the most complicated functions regarding hash, sort & string from Glib to replace the original code, and push to the new branch at https://github.com/unicorn-engine/unicorn/tree/noglib2. the rest is still from the code of @cseagle. please help to test this new branch.

lunixbochs:
What's the glib hash table function? Could be worth replacing it with something like murmur or city or whatever google is using these days while you're in there.

aquynh:
at least we are on the same level as before because we are still using glib code inside. i am not really sure replacing proven stable code with new one would improve anything here.

cseagle:
Did you do any concrete testing or was this "just because"

cseagle:
Make sure you pull in all the relevant copyright notices, the glib code has a number of different contributors and copyright notices

aquynh:
@cseagle no, except running our existing tests in Travis & Appveyor setup. i copied as-is code from Glib, so (hopefully) they are bug-free.

on the license, i will copy it over from glib. thanks for reminding.

cseagle:
The other issue this brings up is tracking upstream for bug fixes. If a bug is fixed in this portion of the glib code will you be tracking that for porting?

aquynh:
yes i will port the bugfixes over.

aquynh:
i merged this PR & the branch "noglib2" into "master", thanks!

PR #693, opened 2016-12-20
add g_free to make future integration easier

cseagle:
Added define for g_free so that integration w/ future versions of qemu won't require global search and replace of g_free to free


aquynh:
merged, thanks!

do you want to revert all the free() back to g_free(), too?

cseagle:
I think that would be a good idea. I will try to track them all down and revert those files where that was the only change and fix others that had other changes.

lunixbochs:
I converted a couple of free to g_free originally because they'd been used with g_alloc and I was overriding the glib allocator for precorn to work

aquynh:
i already merged the PR #695

Issue #691, opened 2016-12-19
Applying and removing hooks segfaults


mrphrazer:
Unicorn segfaults if a hook will be set and then unset immediately. I tested it for ```UC_HOOK_CODE``` and ```UC_HOOK_BLOCK```
 

```python
from unicorn import *
from unicorn.x86_const import *


X86_CODE64 = "\x48\x31\xff\x57\x57\x5e\x5a\x48\xbf\x2f\x2f\x62\x69\x6e\x2f\x73\x68\x48\xc1\xef\x08\x57\x54\x5f\x6a\x3b\x58\x0f\x05"

# memory address where emulation starts
ADDRESS = 0x1000000


# callback for tracing basic blocks
def hook_block(uc, address, size, user_data):
    print(">>> Tracing basic block at 0x%x, block size = 0x%x" %(address, size))

 # Initialize emulator
mu = Uc(UC_ARCH_X86, UC_MODE_64)

# map 2MB memory for this emulation
mu.mem_map(ADDRESS, 2 * 1024 * 1024)

# write machine code to be emulated to memory
mu.mem_write(ADDRESS, X86_CODE64)

# initialize stack
mu.reg_write(UC_X86_REG_ESP, ADDRESS + 0x200000)

# hook block
mu.hook_add(UC_HOOK_BLOCK, hook_block)

""" segfault here """
mu.hook_del(UC_HOOK_BLOCK)

# emulate machine code in infinite time
mu.emu_start(ADDRESS, ADDRESS + len(X86_CODE64))
```

aquynh:
is this on the latest github code? can you send a pull req to put this testcase under tests/regress/ ?

aquynh:
this segfault should be fixed, but your code on `hook_del` is incorrect. it should be something like:

    h = mu.hook_add(UC_HOOK_BLOCK, hook_block)
    mu.hook_del(h)

mrphrazer:
I noticed my mistake a few minutes ago, thanks! However, the code above still segfaults in the newest version (at least in my setups)

aquynh:
the returned hook handle is actually a pointer, which is referenced inside uc_hook_del(), thus triggers crash. unless we save all the handles, there is no way to verify invalid handle passed to uc_hook_del().

this issue is not a critical issue anyway.

lunixbochs:
We do save the handles into the linked list and shouldn't call free unless there was a handle in the list still. This is a bug that the code should handle.

[...]

PR #697, opened 2016-12-21
fix possible segfault in hook del (#691)

lunixbochs:
this will be slightly slower (`UC_HOOK_MAX * len(hooks)` instead of `num_types * len(hooks)`) for removing hooks because we can't safely dereference a hook struct to get its types.

this could be fixed by aligning the hook's pointer and storing the hook types as a mask in the pointer value.


aquynh:
merged, thanks!

PR #698, opened 2016-12-23
python bindings: restore FreeBSD support

ant1:


aquynh:
this is very interesting, but not future-proof: what about FreeBSD 13, 14 etc? i guess you dont want to come back in few years to send another pull req.

for this reason, could you send another PR that makes sure we would never need to do this again? one way is to use default value `libunicorn.so` if the system is not in the list, so this also works with NetBSD, OpenBSD, etc.

ant1:
patch updated

aquynh:
merged, thanks!

PR #699, opened 2016-12-27
Windows: export a static lib that can be used outside of mingw

rhelmot:
The .lib file produced by mingw can't be used outside of mingw, even with all the support DLLs present. We can work around this by using mingw to compile the DLL and then calling out to MSVC to produce the lib file.

If the user doesn't have MSVC installed, the operation will fail and the rest of the build will proceed. The original static lib produced by mingw is kept with a `.a` extension and is still used to compile the static examples.

This allows angr's unicorn support to build on windows.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,8 @@ unicorn.pc
 
 unicorn.lib
 unicorn.dll
+unicorn.exp
+unicorn.def
 unicorn_*.lib
 unicorn_*.exp
 unicorn_*.dll
--- a/Makefile
+++ b/Makefile
@@ -122,10 +122,12 @@ UNICORN_CFLAGS := $(UNICORN_CFLAGS:-fPIC=)
 # mingw?
 else ifneq ($(filter MINGW%,$(UNAME_S)),)
 EXT = dll
-AR_EXT = lib
+AR_EXT = a
 BIN_EXT = .exe
 UNICORN_QEMU_FLAGS += --disable-stack-protector
 UNICORN_CFLAGS := $(UNICORN_CFLAGS:-fPIC=)
+$(LIBNAME)_LDFLAGS += -Wl,--output-def,unicorn.def
+DO_WINDOWS_EXPORT = 1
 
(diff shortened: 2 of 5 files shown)


aquynh:
could you add few notes into `windows_export.bat` to explain what that does?

rhelmot:
Done!

aquynh:
merged, thanks!

aquynh:
on my 64bit Win10 Windows, i run "make.sh cross-win64", and get the error below:

```
...
  CC    sparc64-softmmu/target-sparc/vis_helper.o
  CC    sparc64-softmmu/target-sparc/unicorn64.o
make[1]: Leaving directory '/c/Users/Thinkpad/Documents/GitHub/unicorn3.git/qemu'
  GEN     unicorn.dll
ln: 'unicorn.dll' and './unicorn.dll' are the same file
make: [Makefile:217: unicorn.dll] Error 1 (ignored)
cmd /c "windows_export.bat x64"

C:\Users\Thinkpad\Documents\GitHub\unicorn3.git>FOR /F "usebackq tokens=3*" %A IN (`REG QUERY "HKEY_LOCAL_MACHINE\SOFTWARE\Wow6432Node\Microsoft\VisualStudio\12.0" /v InstallDir`) DO (set appdir=%A %B )
ERROR: The system was unable to find the specified registry key or value.

C:\Users\Thinkpad\Documents\GitHub\unicorn3.git>call "..\..\VC\vcvarsall.bat" x64
The system cannot find the path specified.

C:\Users\Thinkpad\Documents\GitHub\unicorn3.git>call lib /machine:x64 /def:unicorn.def
'lib' is not recognized as an internal or external command,
operable program or batch file.
make: *** [Makefile:218: unicorn.dll] Error 1

```

there are few things:
- my machine does not have VS 12.0 (but 14.0), so `windows_export.bat` cannot find it, thus compilation fails. this means if machine has no VS, we will have the same issue.
- we should make `windows_export.bat` works regardless of VS version.

PR #700, opened 2016-12-27
Support for MCLASS ARM cpu (Cortex-M3)

cojocar:
Support for Cortex-M ARM CPU already exists in Qemu. This patch just
exposes a "cortex-m3" CPU.

Example:
`uc_open(UC_ARCH_ARM, UC_MODE_THUMB | UC_MODE_MCLASS, &uc);`

Signed-off-by: Lucian Cojocar <lucian@cojocar.com>

### DIFF
--- a/include/uc_priv.h
+++ b/include/uc_priv.h
@@ -13,7 +13,7 @@
 
 // These are masks of supported modes for each cpu/arch.
 // They should be updated when changes are made to the uc_mode enum typedef.
-#define UC_MODE_ARM_MASK    (UC_MODE_ARM|UC_MODE_THUMB|UC_MODE_LITTLE_ENDIAN)
+#define UC_MODE_ARM_MASK    (UC_MODE_ARM|UC_MODE_THUMB|UC_MODE_LITTLE_ENDIAN|UC_MODE_MCLASS)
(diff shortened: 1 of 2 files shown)


aquynh:
merged, thanks!

aquynh:
for testing, would you mind providing a simple code that works under M3, but not under A15?

cojocar:
Cortex M3 has ARMv7-M architecture, while Cortex A15 has ARMv7-A architecture. So the difference between the two, from instruction emulation perspective, is that ARMv7-M does not support ARM instructions (i.e. always 4 bytes). Thumb & Thumb-2 are supported by both, but, IIRC, the instruction encoding are different in some cases.

I will try to come up with a test, but cannot promise a deadline for it now.

lunixbochs:
You could just write a test that THUMB2 works but ARM doesn't. Then test that ARM works on the default cpu.

aquynh:
this fix hardcodes M3 with MCLASS mode, so this means you cannot use other models with MCLASS. is this a problem?

we have a branch named cpu_model at https://github.com/unicorn-engine/unicorn/tree/cpu_model, in which we can specify the CPU model at initialization time in `cpu_model` argument. i havent merged that, because it breaks API compatibility in `uc_open()`. we may want to bring the idea back, so we dont need this kind of fix in the future.

PR #703, opened 2017-01-02
Make the call out to visual studio extremely resilient

rhelmot:
Apparently batch scripts return with the return code of their last statement. Who knew.

See comment added in windows_export.bat


aquynh:
this is the 1st ever PR in 2017, thanks!

PR #705, opened 2017-01-04
Python: Automatically conform to PEP440 for version numbers

rhelmot:


aquynh:
merged, thanks!

aquynh:
btw, later we can do the same thing for Capstone

PR #706, opened 2017-01-04
Static library copying nitpicks

rhelmot:
Copy prebuilt static library if available. Disallow uploads from builds that fail to provide the static library.

These are untested, they're just from me scrutinizing the code that's gonna get run as soon as my backup finishes.


aquynh:
merged, thanks!

PR #707, opened 2017-01-04
Allow the client to write to CPSR

agustingianni:
This PR adds the capability to set the `CPSR` register on the ARM architecture.


agustingianni:
Thank you!

PR #713, opened 2017-01-11
Fixed circular refs preventing Uc instances from being GC'd

hedger:
Added a test case, requires `objgraph` module.

### DIFF
--- a/bindings/python/unicorn/unicorn.py
+++ b/bindings/python/unicorn/unicorn.py
@@ -7,6 +7,7 @@ import pkg_resources
 import inspect
 import os.path
 import sys
+import weakref
 
 from . import x86_const, unicorn_const as uc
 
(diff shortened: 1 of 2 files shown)


danghvu:
Can't we go and cleanup the hook properly in `hook_del` instead? And since I (and maybe others) may not be familiar with `objgraph` could you elaborate why we need it ?

hedger:
My implementation allows GC to properly collect those instances where hook_del wasn't called.

Generally, implementing __del__ on your own classes with circular references (that's what we have in Unicorn for callbacks) is discouraged -- it breaks GC, making it keep the object forever. See [Python GC docs](https://docs.python.org/2/library/gc.html#gc.garbage) for details.

`objgraph` is only needed for running regression test. I use it to verify that no instances of `Uc` class are remaining after a GC sweep.

danghvu:
Thanks, you mean `assert(len(gc.garbage) == 0)` is not sufficient in some cases ? 
If this is absolutely necessary I may just write a bit of code similar to what `objgraph.by_type` did, to prevent adding extra dependency to the test.

(quoted from the objgraph source code)
```python
def by_type(typename, objects=None):
    if objects is None:
        objects = gc.get_objects()
    try:
        if '.' in typename:
            return [o for o in objects if _long_typename(o) == typename]
        else:
            return [o for o in objects if _short_typename(o) == typename]
    finally:
        del objects  # clear cyclic references to frame
```

hedger:
> Thanks, you mean `assert(len(gc.garbage) == 0)` is not sufficient in some cases ?
I did some refactoring by updating `self` refs to weakrefs, also tried storing callbacks in a `WeakKeyDictionary` -- and it removed `Uc` instances from `gc.garbage`. But still they were not collected. So, I used `objgraph` to be sure that no instances remain. 

Of course, it is possible to reimplement `by_type` to avoid an additional dependency.

danghvu:
Merged, thanks!

PR #715, opened 2017-01-13
VB6 bindings and dynload update w/ uc_context* and uc_free api,…

dzzie:
in dynload.c I included support for people who are using the latest available binary release 1.0 rc2 which still uses uc_context_free. didnt want to break anyone. Extra check is a single line and can be removed next binary release. No negative effect on newer releases.


aquynh:
@xorstream, do you have any comments on the MSVC binding part of this PR?

aquynh:
@dzzie, great work! would you mind posting a screenshot of your VB6 app?

what is the status of the VB6 binding?

dzzie:
ok I added a screen shot to the commit. its really just a listbox with log results. the log output can be found in the file example_output.txt which I also just updated. 

all of the supported unicorn api are listed in the readme.txt . it basically covers all of the api except for the 2 batch register read/write functions. VB6 should be able to support these as well.  The test program uses all of the api.

The stubs for the instruction hook framework is almost in place, but I did not implement the in/out instruction hooks. If there was a call/ret hook I would implement those. 

All the other hook types are implemented (Code/Block/Memory/Interrupt). 

Currently the sample only contains the x86 32bit constants since this is the most heavily used and natural for vb6 to work with. 

The C shim api does support the full 64bit types still, so it can support the x86 64bit emulator, but I would need to include an additional library to give vb6 easy access to working with larger numbers. (VB6 really only supports 32bit signed numbers natively)

I do not work with any of the other supported processors so I will leave it up to the community to implement whatever else they wish and have expertise with as needed. The api framework in the C shim dll will be fine for any other processor except for the disasm function (optional compile disabled by default) and the memory_write_block which currently only accepts a 32bit address but could easily be changed to a uint64_t. The only difference in this api from the normal one is that it automatically calculates an aligned base and size for the user specified address and buffer size and handles the mapping automatically. (just a bonus method)

aquynh:
nice, as soon as @xorstream approves changes on MSVC binding, i will merge this.

dzzie:
one other thought I had for dynload would be to add a null pointer check on each function pointer before returning true from uc_dyn_load. 

 just as a sanity check in case versions get mismatched. ie using the current dynload against an older dll where an export is missing or to detect/alert if it becomes out of sync with current.

a little more verbose option would be to additionally add a char* lastDynloadErr

gp_uc_mem_map_ptr = (uc_mem_map_ptr_t)DYNLOAD_GETFUNC(g_dyn_handle, "uc_mem_map_ptr");
if(gp_uc_mem_map_ptr==0){
        lastDynloadErr = strdup("failed to load uc_mem_map_ptr");
        return false;
}

changing DYNLOAD_GETFUNC from a macro to a function and handling it in there is probably the easy way to go

xorstream:
Yup looks fine to me.
It's a pretty straight forward process to add new functions and just follow the same pattern when doing so.

And yeh it wouldn't hurt to add some null pointer checks now that it is possible there could be conflicting versions of the unicorn.dll in existence.

aquynh:
merged, thanks!

aquynh:
updated CREDITS.TXT with your name: https://github.com/unicorn-engine/unicorn/commit/d6ba083c4580b99e70e056a95faca4f40be4adff. thanks for your amazing effort!

Issue #714, opened 2017-01-12
Weird behavior using python bindings


eltongo:
I'm experiencing really weird behavior in Unicorn engine using Python bindings.

At first I was having trouble with this instruction:

`ADD W9, W9, #1`

It seemed this instruction was not causing any side effects, because, although executed, upon reading the value of W9, I was always getting 0.

After some experimenting, I noticed that in fact X9 changes, but W9 doesn't reflect any changes. Take for example this code:

```
def hook_code2(uc, address, size, user_data):  
    fmt = """W8: 0x{0:2x}
    X9: 0x{1:2x}
    W9: 0x{2:2x}
    X11: 0x{3:2x}
    X12: 0x{4:2x}
    """.format(
        uc.reg_read(UC_ARM64_REG_W8),
        uc.reg_read(UC_ARM64_REG_X9),
        uc.reg_read(UC_ARM64_REG_W9),
        uc.reg_read(UC_ARM64_REG_X11),
        uc.reg_read(UC_ARM64_REG_X12)
    )

    print(fmt)
    
    return True
```

I hook this function:

`mu.hook_add(UC_HOOK_CODE, hook_code2) `

And I can see the register X9 (but not W9) incrementing (sample terminal output):

```
W8: ...
X9: 0x12d8
W9: 0x0
X11: ...
X12: ...

W8: ...
X9: 0x12d9
W9: 0x0
X11: ...
X12: ...
```

I'm very confused and I'm not sure what to make of this.

lunixbochs:
Post a full example script please

eltongo:
Here's a complete code listing. Below you will see the output for the script, too.

```
from keystone import *
from unicorn import *
from unicorn.arm64_const import *

def hook_code_fails(uc, address, size, user_data):  
    print("-------------- INSIDE hook_code_fails --------------")
    fmt = """
    W9: 0x{1:2x}
    """.format(
        uc.reg_read(UC_ARM64_REG_W8),
        uc.reg_read(UC_ARM64_REG_W9),
        uc.reg_read(UC_ARM64_REG_X11),
        uc.reg_read(UC_ARM64_REG_X12)
    )

    print(fmt)
    
    return True

def hook_code_works(uc, address, size, user_data):
    print("-------------- INSIDE hook_code_works --------------")  
    fmt = """
    X9: 0x{1:2x}
    W9: 0x{1:2x}
    """.format(
        uc.reg_read(UC_ARM64_REG_W8),
[...]

PR #716, opened 2017-01-14
reg_read and reg_write now work with registers W0 through W30 in Aarch64

eltongo:
Based on issue #714, I added the ability to read and write registers W0 through W30 for Aarch64 using the reg_read and reg_write functions.


danghvu:
Cool!

Could you please also commit a test to demonstrate how to use this and show that this fixes the existing issue ? (It can be the same test you wrote in the issue)

eltongo:
While writing my test, I noticed that when writing to a 32-bit register (e.g. W0), it zeroes out the value of X0.

It seems the macro `WRITE_DWORD` (see snippet below) has an issue when writing to a `int64_t` value:

`#define WRITE_DWORD(x, w) (x = (x & ~0xffffffff) | (w & 0xffffffff))`

As seen above, in `x & ~0xffffffff`, since `0xffffffff` is an `int32_t` literal, first `~0xffffffff` is evaluated to zero, and then it is `&`ed with x, zeroing the original value out.

Shall I include another macro, something like `WRITE_DWORD_ON_QWORD` (below)?

`#define WRITE_DWORD_TO_QWORD(x, w) (x = (x & ~0xffffffffLL) | (w & 0xffffffff))`

aquynh:
how about renaming your macro WRITE_DWORD_TO_QWORD to WRITE_DWORD, thus replace the original WRITE_DWORD, to make WRITE_DWORD work in every cases?

aquynh:
and please fix READ_DWORD macro at the same time.

eltongo:
Alright. The `WRITE_DWORD` macro is now fixed. `READ_DWORD` should be fine, though.

aquynh:
merged, thanks!

PR #722, opened 2017-01-20
Fixed double free in win32 threads and changed free() to g_free().

xorstream:
Fixed double free in win32 threads and changed free() to g_free().

### DIFF
--- a/qemu/util/qemu-thread-win32.c
+++ b/qemu/util/qemu-thread-win32.c
@@ -127,7 +127,9 @@ void *qemu_thread_join(QemuThread *thread)
     ret = data->ret;
     assert(data->mode != QEMU_THREAD_DETACHED);
     DeleteCriticalSection(&data->cs);
+    data->uc->qemu_thread_data = NULL;
     g_free(data);
+    data = NULL;
     return ret;
 }
 
--- a/uc.c
+++ b/uc.c
@@ -312,7 +312,7 @@ uc_err uc_close(uc_engine *uc)
 
     // Thread relateds.
     if (uc->qemu_thread_data)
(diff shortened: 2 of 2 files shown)


qemu/util/qemu-thread-win32.c
```diff
@@ -127,7 +127,9 @@ void *qemu_thread_join(QemuThread *thread)
     ret = data->ret;
     assert(data->mode != QEMU_THREAD_DETACHED);
     DeleteCriticalSection(&data->cs);
+    data->uc->qemu_thread_data = NULL;
     g_free(data);
+    data = NULL;
```
aquynh: why do you need to set "data" to NULL here?

aquynh:
can you explain how the double free happens?

xorstream:
The "data = NULL" isn't really needed but i copied it from qemu-thread-win32.c: win32_start_routine().
win32_start_routine() will free the data if detatched, otherwise it gets freed at the end of qemu_thread_join(). Based on this I figured I would make the code match to make it obvious that it is doing the same thing in these two places.

The double free was occurring in uc.c: uc_close().
These lines:
    if (uc->qemu_thread_data)
        g_free(uc->qemu_thread_data);
This would free the same data that gets freed in qemu_thread_join() and win32_start_routine(). The double free occurs when the thread isn't detatched and so qemu_thread_join() freed it without also setting uc->qemu_thread_data to NULL.

It could be that these lines are not needed at all, but I didn't want to remove them *just in case*.
    if (uc->qemu_thread_data)
        g_free(uc->qemu_thread_data);

aquynh:
merged, thanks. i will get back to fix these ugly broken code later.

PR #723, opened 2017-01-20
Msvc native

xorstream:
First big chunk of changes to support MSVC native builds in 32bit and 64bit.
This currently supports only x86/x64 CPU as target.
Once this is sorted the rest of the CPUs will follow.

There are LOTS of changes but some of the common ones are:
* Moved OS specific headers to platform.h
* Removed GNU/GCC specific __attributes__(()) and either inserted replacements inline or in compiler.h or whereever seemed to fit best for each case.
* Removed GNU/GCC specific macros that returned values, ie: #define FUNC(x) ({STUFF})
* Removed GNU/GCC specific struct initialisers, ie: var = (Type) { values };
* Removed GNU/GCC specific switch/case ranges, ie: case 0x10...0x20:
* Removed GNU-GCC specific struct initialise members by name, ie: struct s = { .name1=value1, .name2=value2 }
* Put "config generated" files config-host.h and config-target.h into bindings project dir so they wont be altered/removed by calls to configure.
* Moved variable declarations to the top of functions or blocks as MSVC won't let them be declared part way through a function.
* Had to change some macros that would declare a variable in place, in some cases I split them into 2 where the first would define the variable and then the second would use it. This was to satisfy the condition that variables must be declared at the start of a function/block.
* Lots of casts to remove warnings about data sizes changing size and so possible data being lost. In the end I "turned off" this warning because there were still lots of these. If required I can turn these back on and remove the rest.


aquynh:
awesome!

a quick comment: this fails to pass Travis now: https://travis-ci.org/unicorn-engine/unicorn/builds/193718149

aquynh:
Appveyor also fails to compile with Msys now: https://ci.appveyor.com/project/aquynh/unicorn/build/1.0.886/job/ms0h2pxis44l9281

xorstream:
looking into it..

aquynh:
file "include/platform.h" should be moved to "include/unicorn/", because you want this to be public

bindings/java/unicorn_Unicorn.c
```diff
@@ -20,7 +20,7 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 */
 
 #include <sys/types.h>
-#include <stdint.h>
+#include "platform.h"
```
aquynh: i suppose that we port the core to MSVC, but leave the bindings alone?

qemu/include/elf.h
```diff
@@ -1,9 +1,294 @@
 #ifndef _QEMU_ELF_H
 #define _QEMU_ELF_H
 
+#include "platform.h"
```
aquynh: i cleaned up this file recently, so this file is based on the older code.

xorstream: are you sure? i recall updating this just before doing pull request.

aquynh: yes, you can see the current file at https://github.com/unicorn-engine/unicorn/blob/master/qemu/include/elf.h

bindings/msvc/samples/main.c
```diff
@@ -22,7 +22,7 @@
 
 // posix specific
 #else // _MSC_VER
-#include "platform.h"
+#include "unicorn/platform.h"
```
aquynh: the line above is not needed, because we include `unicorn/unicorn.h` in the next line, which includes `platform.h` inside

aquynh:
it looks much better now! few more things:

- it seems you are using tab for indentation. would you change that to 4 spaces instead, to match the current coding style?

- the warning "suggest braces around initialization..." on qemu/target-i386/translate.c at https://travis-ci.org/unicorn-engine/unicorn/jobs/193757694 looks worrisome. is it possible to fix that?

- gcc also produces some warnings at https://travis-ci.org/unicorn-engine/unicorn/jobs/193757696

xorstream:
I did the leading tab to space conversion using "expand -i -t4" and did the commit separately just in case something goes wrong with it.

aquynh:
you did a tremendous job! i merged this to a new branch at https://github.com/unicorn-engine/unicorn/tree/msvc, please take the new pull req there. 

i will do more cleanup on that tree, too.

xorstream:
I notice some checks did not pass. Do you want me to fix these up?

Also what does it mean that it has been merged to https://github.com/unicorn-engine/unicorn/tree/msvc ?
Does that mean it is not part of the main/master branch?
Or that it is just being set aside for now until it is prettied up?

aquynh:
all checks passed now for the "msvc" branch.

for now, we need more review & test before get this merged into "master" branch.

meanwhile, please go ahead porting the rest architectures, and make pull req on the "msvc" branch.

xorstream:
I am a git n00b, so want to make sure I do this correctly.
Do I just continue to work on it like I have been, but when I do a pull request I specify into "msvc" branch instead of into "master" branch?

aquynh:
yes, that is correct: you just pull the latest Github code, switch to work on the "msvc" branch, make changes, then push to your Github fork. finally, you make the pull req from your own "msvc" branch to the "msvc" branch of this repo.

qemu/exec.c
```diff
@@ -1484,14 +1489,17 @@ static void tcg_commit(MemoryListener *listener)
...
+        mem_commit,
+        mem_add,
+        NULL,
+        mem_add,
+        NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL,
+        0,
     };
+    as->dispatch = NULL;
+    as->dispatch_listener = ml;
     memory_listener_register(as->uc, &as->dispatch_listener, as);
```
aquynh: this part (and many other places) can be rewritten like below to (1) keep the change minimal, so minimize mistake. (2) better compatible with future change on the struct of MemoryListener, thus make it easier to port to future Qemu version:

```
MemoryListener ml;

ml.begin = mem_begin;
ml.commit = mem_commit;
ml.region_add = mem_add;
ml.region_nop = mem_add;
ml.priority = 0;

as->dispatch = NULL;
as->dispatch_listener = ml;
```

xorstream: Yes the problem is that const structs cannot be initialised like this.
So instead of doing things in multiple different ways I just decided to choose the most generic way for all cases.

For some cases I did create C style constructors that would setup structs from member values, but again this only works for non-const ones.

aquynh: yes i saw what you did. totally agreed that must be ugly in the cases you mentioned.

but like in the code above we can avoid that.

aquynh:
this gets merged to the "msvc" branch, so lets close this.

PR #725, opened 2017-01-21
Added prebuild script to remove generated files and started adding projects for other CPUs.

xorstream:
Added prebuild_script.bat to remove leftover configure generated files before building.
Also added project files and MSVC copies of configure generated files for all supported CPUs.


xorstream:
This should fix at least the stdbool.h in qapi-types.h issue.
If we are lucky it will also fix the __int128 issue, but if it doesn't we need to further explore this.

I'm done for this evening though, I'll look into it some more tomorrow.

aquynh:
merged, thanks!

aquynh:
We need to have a script (to be called from Makefile) to update these auto-gen files, so any time there is a change, we can run it to sync MSVC port with the core.

Also, is it right to put this port under bindings/ dir? This is not really a binding, so lets move it to the root dir instead?

xorstream:
Does it build for you now in VS2015?

The config*.h files are created by configure.
The qapi* files are created by the qemu/Makefile calling scripts/qapi-*.py.
So in order to generate these files you need configure, make and python.
I think the best way would be to add an option to make.sh to update the generated files for msvc. This could then be called after performing a build using mingw64 or similar.

When I started this project I wasn't sure where to put it, so I just put it in bindings and called it "msvc_native" to differentiate from the existing "msvc" dir. But yeh we can move the current "msvc_bindings" dir to root and rename it "msvc".

PR #726, opened 2017-01-22
Moved ./bindings/msvc_native into ./msvc

xorstream:
Moved ./bindings/msvc_native into ./msvc
Did some dir moves inside msvc to separate samples and unicorn projects.
Added "msvc_update_genfiles" option to make.sh to copy all generated files into msvc dirs.


aquynh:
this conflicts with the current code, and messed up even when i tried to fix it. can you make a new pull req based on the latest code of "msvc" branch?

aquynh:
merged, thanks!

PR #733, opened 2017-01-23
Fix for MIPS issue.

xorstream:
Fix for MIPS issue.
Also converted leading tabs to spaces.


aquynh:
what was wrong? i looked thru this PR, but did not spot where you fixed the issue.

qemu/target-mips/cpu.c
```diff
@@ -151,15 +151,17 @@ void mips_cpu_register_types(void *opaque)
...
+        opaque,
         
-		mips_cpu_initfn,
-		NULL,
-		NULL,
+        mips_cpu_initfn,
+        NULL,
+        NULL,
+
+        NULL,
```
aquynh: ah it seems it broke before because you missed a NULL above? this is why we should avoid doing this struct initialization when possible (which is avoidable here)

aquynh:
merged, thanks!

qemu/target-mips/cpu.c
xorstream: Yes it was this missing NULL.
This is one of those const initialisations though.
You could remove the const, but then you are trading readability for performance.

aquynh: yes, but that is OK to be slower here, as mips_cpu_register_types() is only executed once, at init time (i.e uc_open()).

PR #736, opened 2017-01-23
Arm support ported.

xorstream:
Arm support ported.


xorstream:
Looks like the arm code needs another check over it.
It was a horrible beast to convert.
I will check over it tomorrow.

aquynh:
merged, thanks!

PR #738, opened 2017-01-24
Arm issue fix.

xorstream:
I ended up writing a small test program to compare the structs before and after porting because ARM has soo many. The issue ended up being an error in the porting of the struct for "ICIMVAU" instruction.

This means that all the ARM instruction structs have been compared and now match the original 100%.


qemu/target-arm/helper.c
```diff
@@ -1883,7 +1883,7 @@ static const ARMCPRegInfo v8_cp_reginfo[] = {
       ARM_CP_NOP, PL1_W },
     { "ICIALLU", 15,7,5, 0,0,0, 0,
       ARM_CP_NOP, PL1_W },
-    { "ICIMVAU", 15,7,5,0,1, 0,
+    { "ICIMVAU", 15,7,5, 0,0,1, 0,
```
aquynh: it seems this is the only place you fixed, but the rest jut got some accident changes, right?

xorstream:
The other changes were using whitespace to group sets of variables together to help show that struct members are correct.

These other changes are not needed to fix the issue, but they are useful for seeing at a glance what struct values are what.

aquynh:
merged, thanks!

PR #739, opened 2017-01-24
Finalise MSVC port

xorstream:
This should be everything done now for the MSVC port.


aquynh:
merged, thanks!

aquynh:
this copies to "distro\include\" some private files: list.h, qemu.h & uc_priv.h. but these files are all private, but not part of public API, thus should not be distributed.

aquynh:
i fixed this issue: https://github.com/unicorn-engine/unicorn/commit/2853b4b1cd5004cfa74042f9d174e2260e69c48f

aquynh:
looks like the static lib is compiled with Debug info, even if i choose Release mode for "Rebuild solution"? in "distro\", the size of "unicorn.dll" is about 4MB, but "unicorn_static.lib" is 68MB.

PR #740, opened 2017-01-25
usleep() change for windows to not use winsock

xorstream:
Changed usleep() to not use winsock anymore.
This also sets a 20ms resolution for the windows usleep() implementation.
Updated readme to remove winsock notes.


aquynh:
Do you want to update docs in unicorn.h on this API as well?

aquynh:
why do you choose roundup to 20ms, but not for ex 15ms?

xorstream:
I chose 20ms because some of the pages I read mention that the minimum wait time might be as long as 15.6ms, so 20ms was a nice round number that was above this.

Btw thinking back over my implementation I should probably make it do a minimum wait of 20ms, and then just round up to the next 1ms after 20ms. Currently I have it rounding up to multiple of 20ms.

aquynh:
ok, can you fix that, and update docs in unicorn.h about the difference on time resolutin on Windows?

btw, later we can move all the code in platform.h to somewhere else private, and leave only data types there.

include/unicorn/platform.h
```diff
@@ -151,9 +151,17 @@ typedef _W64 unsigned int  uintptr_t;
...
 #include <windows.h>
 
+// timeval is defined in winsock.h and winsock2.h
+// FD_ZERO is defined in both of these just above struct timeval.
+#if !defined(FD_ZERO)
+struct timeval {
+    long tv_sec;
+    long tv_usec;
+};
+#endif
```
aquynh: why do we need this struct here?

xorstream: Usually it is defined in sys/time.h which is a linux/posix only header.

Previously it was being defined inside the winsock header, for winsock posix compatibility, but since we are removing the winsock related code I just define that one struct.

vardyh:
why not using SetWaitableTimer() to implement usleep() on windows?
SetWaitableTimer() allows you to control the sleep time in 100-nanosecond units.

aquynh:
@vardyh, can you send a pull req to implement your idea?

vardyh:
@aquynh, sure, request #745 represents my implementation.

PR #741, opened 2017-01-25
Changes to reduce size of libs

xorstream:
After a bit of googling it seems the way to reduce the size of the libs is to turn off whole program optimisations.
After doing so the resultant lib files are about 1/3 to 1/4 the size that they were.

This explains the setting that made the difference:
https://msdn.microsoft.com/en-us/library/0zza0de8.aspx


aquynh:
merged, thanks!

PR #743, opened 2017-01-28
Fixed build script for *BSD

hedger:
Moved $UNAME variable initialization a few lines higher -- before it is used.


aquynh:
merged, thanks!

PR #746, opened 2017-02-03
MSVC port (vardyh)

vardyh:
Use waitable timer to implement usleep() on Windows and get rid of usage of socket library.

### DIFF
--- a/include/unicorn/platform.h
+++ b/include/unicorn/platform.h
@@ -169,20 +169,24 @@ static int gettimeofday(struct timeval* t, void* timezone)
 // unistd.h compatibility
 #if defined(_MSC_VER)
 
-// horrible kludge requiring winsock to get microsecond sleep resolution.
-// if this is removed then all winsock references can also be removed.
-static int usleep(uint32_t t)
+static int usleep(uint32_t usec)
 {
-    int ret, err_code;
-    long value = t; // time in microseconds
-    struct timeval tv;
-    FD_SET dummy_set;
-    FD_ZERO(&dummy_set);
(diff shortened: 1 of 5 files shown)


aquynh:
looks nice! @xorstream, do you have any comments?

vardyh:
barrier() implementation for MSVC in updated in commit 1a4d56d2dc3d5e54b983e1cb2ea885db833ac710

aquynh:
thanks, just a minor thing: can you indent with 4 spaces, which is the coding style that we are following?

lets see if @xorstream has any comments, so i can merge this.

xorstream:
Looks fine to me

vardyh:
@aquynh I adjusted the indentation and made a rebase.

aquynh:
merged, thanks!

aquynh:
@xorstream, can you confirm that now https://github.com/unicorn-engine/unicorn/pull/740 is not needed anymore?

xorstream:
#740 is no longer needed

xorstream:
Now that we have MSVC native support, should we remove the MSVC binding to reduce confusion?

aquynh:
yes, please pull req.

xorstream:
There is a "windows_export.bat" file in the root that is used to build a static lib usable outside of mingw.
This file was added by someone else so I have left it in place but you might want to check with its author if it is still needed or not.

aquynh:
that was from @rhelmot, but i think we still need it for Msys build. so we allow users to either use MSVC or Msys to compile Unicorn.

PR #747, opened 2017-02-05
Fix 16-bit address computations

### LINKED ISSUE #737: Segment register ignored while emulating 16-bit x86 assembly
It seems that the emulator does not handle the following instruction properly:

66 a3 xx 00     mov ds:XX, eax

Instead the instruction is shown as **mov dword ptr [0xXX], eax** which effectively ignores the segment register.  

Please, see the attached 16-bit x86 vbr1.bin file to reproduce the behavior (the instruction is at offset 0xa5)

The emulator was created as follows:

emu = unicorn.Uc(unicorn.UC_ARCH_X86, unicorn.UC_MODE_16)

[vbr.zip](https://github.com/unicorn-engine/unicorn/files/724035/vbr.zip)

cseagle:
changes to 16 bit segment registers need to update segment base as well as segment selector. This may fix #737


aquynh:
merged, thanks!

PR #751, opened 2017-02-08
Fix initial state of segment registers

cseagle:
more appropriate solution to initial state of x86 segment registers in 16-bit mode


aquynh:
merged, thanks!

PR #753, opened 2017-02-20
Update unicorn.h

gdbinit:
Make it clear that only very few instructions can be hooked.

This isn't very clear in documentation (or at least the include that users consult) and it should be to avoid users trying to hook and getting no results.


aquynh:
merged, thanks!

PR #755, opened 2017-02-23
x86: add and export uc_msr_read/uc_msr_write

asamy:
The x86 implementation should have those, as they are really useful, apart from that, it's kind of painful to write shellcodes manually for them, and seems like a stupid task, so I decided to just make those functions.

I probably missed some stuff, so let me know if there is anything I should do.
The bindings for other languages are kind of difficult for me to test, so I'll cc the people that did it, should they do it or so.

/cc @lunixbochs
/cc @cseagle


aquynh:
This is nice, but a problem is that these new API are for X86 only. We want arch-independent API (just like the rest), so can you come up with something else?

asamy:
I really have no idea what other architectures have to say about this stuff, the only architecture I worked on is x86, so I'd rather pass on that.

cseagle:
I agree that this is very x86 specific and probably not a good candidate for a public API.

The best idea I can come up with is something like:

uc_err uc_reg_write_ex(uc_engine *uc, int regid, ...);

Where arch specific handlers would then deal with arch specific things. In the simplest cases, regid just maps to an existing reg id and this just gets passed along to uc_reg_write(uc, regid, val) where val is the one one and only extra argument pulled from the optional arguments.

For x86, a new regid UC_X86_REG_MSR could be defined and handled in the x86 aarch specific code. In this example two additional arguments would be provided, an MSR number, and the value for that MSR.

I don't know if any of the other supported architectures have registers that are difficult to handle with the current uc_reg_write/read, but this approach might solve any such problems there as well.

asamy:
I actually thought about that, but why _ex?  Since uc_reg_write already has a void* param, we can just use that to extract 12 bytes, I guess both are fine.

So, tomorrow morning I will purge those commits and do this instead others agree.

cseagle:
_ex for "extended version of an existing function". My personal preference is a va_list over various unions/custom structures, but that's just me.

aquynh:
yes, just define new register id for MSR, and go with uc_reg_read/write. flexibility with all kind of registers is is exactly why i designed those API to use void * argument.

cseagle:
If that is the case, then please also add documentation that describes exactly what the void * is pointing at. Note that in the case of uc_reg_read, the void * will need to be both an in (msr#) and out (msr value) parameter.

aquynh:
Yes, we definitely need good documentation. Please provide sample code as
well.

asamy:
OK, should be fine now.  I don't think it's worth a sample code, maybe a test, so I'll look into doing a test soon.

cseagle:
I suggest changing the prototypes of uc_msr_read and uc_msr_write to just take the uc pointer and the msr pointer. Dereferencing two fields in the same struct every time you call a function seems redundant to me:

so

static int x86_msr_read(struct uc_struct *uc, unsigned int msr, uint64_t *val)
becomes
static int x86_msr_read(struct uc_struct *uc, uc_x86_msr *msr);
moving the dereferencing into x86_msr_read

static int x86_msr_write(struct uc_struct *uc, unsigned int msr, unsigned long long val)
becomes
static int x86_msr_write(struct uc_struct *uc, uc_x86_msr *msr)
moving the dereferencing into x86_msr_write

x86_msr_read(uc, msr->rid, &msr->value);
becomes
x86_msr_read(uc, msr);

x86_msr_write(uc, msr->rid, msr->value);
becomes
x86_msr_write(uc, msr);

qemu/target-i386/unicorn.c
```diff
@@ -156,10 +160,54 @@ void x86_reg_reset(struct uc_struct *uc)
...
+	/* The implementation doesn't throw exception or return an error if there is one, so
+	 * we will return 0.  */
+	return 0;
+}
+
+static int x86_msr_write(struct uc_struct *uc, unsigned int msr, unsigned long long val)
+{
+    CPUX86State *env = (CPUX86State *)uc->cpu->env_ptr;
+	uint64_t ecx = env->regs[R_ECX];
+    uint64_t eax = env->regs[R_EAX];
```
aquynh: indentation is wrong here (we are using 4-spaces for indentation)

qemu/target-i386/unicorn.c
```diff
@@ -156,10 +160,54 @@ void x86_reg_reset(struct uc_struct *uc)
     }
 }
 
+static int x86_msr_read(struct uc_struct *uc, unsigned int msr, uint64_t *val)
+{
+	CPUX86State *env = (CPUX86State *)uc->cpu->env_ptr;
+	uint64_t ecx = env->regs[R_ECX];
+    uint64_t eax = env->regs[R_EAX];
```
aquynh: indentation is wrong here

qemu/target-i386/unicorn.c
```diff
@@ -376,6 +424,10 @@ int x86_reg_read(struct uc_struct *uc, unsigned int *regs, void **vals, int coun
                         ((uc_x86_mmr *)value)->selector = (uint16_t)X86_CPU(uc, mycpu)->env.tr.selector;
                         ((uc_x86_mmr *)value)->flags = X86_CPU(uc, mycpu)->env.tr.flags;
                         break;
+				    case UC_X86_REG_MSR:
```
aquynh: looks like you are using tabs for indentation in several places?

aquynh:
oh this defines a new struct for MSR registers just to keep the ID & its value. what do you think about using the existing convention that we use for all kind of registers now? the MSR value must be 64bit variable, so you just pass the pointer of 64bit number to uc_reg_read() and uc_reg_write(). as long as we document this well, we will be fine.

cseagle:
I don't understand what you are proposing here. To set an MSR you need to specify an MSR number in the range of uin32_t, and then the MSR value which may be up to uint64_t, so a total of 12 bytes are necessary to specify an MSR. This is the purpose of the uc_x86_msr struct type. How does this this translate to passing a uint64_t* to uc_reg_read? uc_x86_msr* meets the need nicely.

aquynh:
what is the exact range of the MSR id?

cseagle:
A very quick check shows a minimum ID of 0, and a maximum ID of 0xC0000103, a uint32_t is necessary at a minimum

aquynh:
i see, then your proposal to optimize the code above looks good to me.

asamy:
Looks good now?

qemu/target-i386/unicorn.c
```diff
@@ -156,6 +160,49 @@ void x86_reg_reset(struct uc_struct *uc)
...
+    helper_rdmsr(env);
+
+    msr->value = ((uint32_t)env->regs[R_EAX]) |
+    	((uint64_t)((uint32_t)env->regs[R_EDX]) << 32);
+
+    env->regs[R_EAX] = eax;
+    env->regs[R_ECX] = ecx;
+    env->regs[R_EDX] = edx;
+
+	/* The implementation doesn't throw exception or return an error if there is one, so
```
aquynh: indentation here, too.

qemu/target-i386/unicorn.c
```diff
@@ -156,6 +160,49 @@ void x86_reg_reset(struct uc_struct *uc)
...
+    env->regs[R_ECX] = msr->rid;
+    env->regs[R_EAX] = (unsigned int)msr->value;
+    env->regs[R_EDX] = (unsigned int)(msr->value >> 32);
+    helper_wrmsr(env);
+
+    env->regs[R_ECX] = ecx;
+    env->regs[R_EAX] = eax;
+    env->regs[R_EDX] = edx;
+
+	/* The implementation doesn't throw exception or return an error if there is one, so
```
aquynh: indentation here.

aquynh:
can you document this new feature in https://github.com/unicorn-engine/unicorn/blob/master/include/unicorn/x86.h?

qemu/target-i386/unicorn.c
```diff
@@ -177,8 +177,8 @@ static int x86_msr_read(struct uc_struct *uc, uc_x86_msr *msr)
     env->regs[R_ECX] = ecx;
     env->regs[R_EDX] = edx;
 
-	/* The implementation doesn't throw exception or return an error if there is one, so
-	 * we will return 0.  */
+    /* The implementation doesn't throw exception or return an error if there is one, so
+     * we will return 0.  */
 	return 0;
```
aquynh: did you miss this line?

qemu/target-i386/unicorn.c
```diff
@@ -198,8 +198,8 @@ static int x86_msr_write(struct uc_struct *uc, uc_x86_msr *msr)
     env->regs[R_EAX] = eax;
     env->regs[R_EDX] = edx;
 
-	/* The implementation doesn't throw exception or return an error if there is one, so
-	 * we will return 0.  */
+    /* The implementation doesn't throw exception or return an error if there is one, so
+     * we will return 0.  */
 	return 0;
```
aquynh: you missed this line, too?

asamy:
Rebased and added an inline documentation in the commit to describe what this is about.

qemu/target-i386/unicorn.c
asamy: What do you mean I missed this line?  You asked for indentation fix.

aquynh: because that line should be indented, too?

asamy: It was indented...

qemu/target-i386/unicorn.c
```diff
@@ -156,6 +160,49 @@ void x86_reg_reset(struct uc_struct *uc)
...
+    msr->value = ((uint32_t)env->regs[R_EAX]) |
+    	((uint64_t)((uint32_t)env->regs[R_EDX]) << 32);
+
+    env->regs[R_EAX] = eax;
+    env->regs[R_ECX] = ecx;
+    env->regs[R_EDX] = edx;
+
+    /* The implementation doesn't throw exception or return an error if there is one, so
+     * we will return 0.  */
+	return 0;
```
aquynh: i mean the line 182 should be indented. same for line 203.

asamy:
Focusing a bit too much on indentation, but done...

aquynh:
i suppose you are using Vim, then you can just mark a block between `{` & `}`, and press `=`, and Vim does all the indentation for you, without missing anything.

asamy:
Yeah, but the issue is my C style settings are different, case blocks, etc.

aquynh:
merged, thanks!

PR #762, opened 2017-02-24
Makefile: Allow for overriding job threads for making 'qemu' dir

Conan-Kudo:
Providing a default level of parallelism is perfectly fine, but it should also be possible to set higher or lower levels of parallelism easily. This is particularly beneficial for people packaging the software for Linux distributions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,13 @@ include pkgconfig.mk	# package version
 
 LIBNAME = unicorn
 UNAME_S := $(shell uname -s)
+# SMP_MFLAGS is used for controlling the amount of parallelism used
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -9,6 +9,7 @@ include pkgconfig.mk	# package version
 
 LIBNAME = unicorn
 UNAME_S := $(shell uname -s)
+SMP_MFLAGS := -j4
```
aquynh: could you put in some comments for this line, so users can understand what it is, and how to use it?

Conan-Kudo: Done in amended commit.

aquynh:
merged, thanks!

PR #767, opened 2017-02-25
Haskell bindings update

adrianherrera:
Adds Haskell bindings for the `uc_reg_write_batch` and `uc_reg_read_batch` functions


aquynh:
merged, thanks!

Issue #773, opened 2017-03-06
Cannot read/write ARM64 Neon registers


stevielavern:
Hello,
Reading/writing ARM64 Neon registers don't seem to be implemented. For instance the following code:

```python
from unicorn import *
from unicorn.arm64_const import *

uc = Uc(UC_ARCH_ARM64, UC_MODE_ARM)

uc.reg_write(UC_ARM64_REG_X0, 1)  # General Purpose
print "X0 = 0x%x" % uc.reg_read(UC_ARM64_REG_X0)

uc.reg_write(UC_ARM64_REG_V0, 1) # Neon
print "V0 = 0x%x" % uc.reg_read(UC_ARM64_REG_V0)

uc.reg_write(UC_ARM64_REG_D0, 1) # Neon
print "D0 = 0x%x" % uc.reg_read(UC_ARM64_REG_D0)
```

produces:

```
X0 = 0x1
V0 = 0x0
D0 = 0x0
```

This seems to be because of `arm64_reg_read` and `arm64_reg_write` (cf https://github.com/unicorn-engine/unicorn/blob/master/qemu/target-arm/unicorn_aarch64.c#L45 ) which only handle General Purpose registers.
Would it be possible to add support for Neon registers?

aquynh:
yes, that is possible. pull req to support this is welcome.

PR #774, opened 2017-03-06
uc_reg_read & uc_reg_write now support ARM64 Neon registers

stevielavern:
Fixing #773, including Python bindings.

Test:

```
In [1]: uc.reg_write(UC_ARM64_REG_Q0, 0x00112233445566778899aabbccddeeff)

In [2]: hex(uc.reg_read(UC_ARM64_REG_Q0))
Out[2]: '0x112233445566778899aabbccddeeffL'

In [3]: hex(uc.reg_read(UC_ARM64_REG_D0))
Out[3]: '0x8899aabbccddeeffL'

In [4]: hex(uc.reg_read(UC_ARM64_REG_S0))
Out[4]: '0xccddeeffL'

In [5]: hex(uc.reg_read(UC_ARM64_REG_H0))
Out[5]: '0xeeffL'

In [6]: hex(uc.reg_read(UC_ARM64_REG_B0))
Out[6]: '0xffL'
```

Also aliased `UC_ARM64_REG_V0` to `UC_ARM64_REG_Q0` since they are the same.

### DIFF
--- a/bindings/python/unicorn/unicorn.py
+++ b/bindings/python/unicorn/unicorn.py
@@ -9,7 +9,7 @@ import os.path
 import sys
 import weakref
 
-from . import x86_const, unicorn_const as uc
+from . import x86_const, arm64_const, unicorn_const as uc
 
 if not hasattr(sys.modules[__name__], "__file__"):
     __file__ = inspect.getfile(inspect.currentframe())
@@ -223,6 +223,9 @@ class uc_x86_xmm(ctypes.Structure):
         ("high_qword", ctypes.c_uint64),
     ]
 
+class uc_arm64_neon128(uc_x86_xmm):
+    pass
+
 # Subclassing ref to allow property assignment.
 class UcRef(weakref.ref):
     pass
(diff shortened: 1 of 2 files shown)


bindings/python/unicorn/unicorn.py
```diff
@@ -9,7 +9,7 @@
 import sys
 import weakref
 
-from . import x86_const, unicorn_const as uc
+from . import x86_const, arm64_const, unicorn_const as uc
```
aquynh: are you sure this is necessary, as arm64_const is already imported in __init__.py?

bindings/python/unicorn/unicorn.py
```diff
@@ -223,6 +223,9 @@ class uc_x86_xmm(ctypes.Structure):
         ("high_qword", ctypes.c_uint64),
     ]
 
+class uc_arm64_neon128(uc_x86_xmm):
```
aquynh: because this struct is independent from uc_x86_xmm, do you want to create a new class, without reusing uc_x86_xmm? even if the content duplicates uc_x86_xmm, it is fine

bindings/python/unicorn/unicorn.py
aquynh: oh this is necessary. then this is OK, but we need to refactor the code to separate arch-dependent code into their own files later.

bindings/python/unicorn/unicorn.py
stevielavern: Not sure if duplicating code is something we want. Maybe we could have a mother class like `vector128` and then 2 child classes `uc_x86_xmm` and `uc_arm64_neon128` inherit from it?

aquynh:
Feel free to duplicate. We still need to refactor arch-dependent code to
separate files later, so this should not be the issue.

stevielavern:
Done.

aquynh:
merged, thanks!

stevielavern:
Thank you!

PR #775, opened 2017-03-07
fix register widths for MIPS64 reg_read/write

mattypiper:
In Unicorn's interface to qemu/target-mips, reg_read and reg_write were only operating on the lower 32-bits of a 64-bit register. This change allows reading/writing the full register.


aquynh:
merged, thanks!

PR #776, opened 2017-03-09
Add support for ARM application flags - APSR register

feliam:
Support for reading writing ARM_REG_APSR. Only 4 msb. (no GE)


aquynh:
merged, thanks!

PR #778, opened 2017-03-12
Homebrew

stephengroat:
start monitoring `homebrew` builds by building with `homebrew install --HEAD`, but allow failures (in case a change is needed in the `homebrew` script)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,9 @@ matrix:
     - os: osx
       script: brew install --HEAD unicorn
       compiler: clang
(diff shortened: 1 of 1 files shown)


aquynh:
merged, thanks!

aquynh:
btw, we are no longer depend on glib. could you please send a pull req to Brew to remove this dependency?

stephengroat:
Homebrew/homebrew-core#10931

aquynh:
awesome, thanks!

PR #780, opened 2017-03-13
armeb: Add support for ARM big endian.

xizhizhang:
Hi,
I have tried supported ARM's big endian mode for Unicorn. And I have also added sample_armeb.

Thanks.


bindings/python/sample_armeb.py
```diff
@@ -0,0 +1,105 @@
+#!/usr/bin/env python
+# Sample code for ARM of Unicorn. Nguyen Anh Quynh <aquynh@gmail.com>
```
xizhizhang: I am sorry,  samples/.gitingore ingored my C sample, and I had not found it.

aquynh:
nice, but you still need to add this new sample to https://github.com/unicorn-engine/unicorn/blob/master/samples/Makefile, or it doesnt not get compiled

xizhizhang:
I have already modified samples/Makefile, just as:
` SOURCES += sample_arm.c
 SOURCES += sample_armeb.c`

I have tested, and it can be compiled.
Is this not a suitable modify?

aquynh:
sorry i missed that.

will look closer before merging, thanks.

qemu/target-arm/unicorn_arm.c
```diff
@@ -9,8 +9,9 @@
 #include "unicorn_common.h"
 #include "uc_priv.h"
 
-
+#ifndef TARGET_WORDS_BIGENDIAN
```
aquynh: this line means that for big-endian, we don't have ARM_REGS_STORAGE_SIZE, why?

aquynh:
this looks great! few quick comments:

- you should always run "make header" to generate a complete qemu/armbe.h. 
- please put your name into https://github.com/unicorn-engine/unicorn/blob/master/CREDITS.TXT

i fixed some minor issues, merged & pushed to the new branch https://github.com/unicorn-engine/unicorn/tree/armeb. please address the remaining comments, send pull req to this "armeb" branch, and i will merge this back to "master" branch.

thanks.

qemu/target-arm/unicorn_arm.c
xizhizhang: I did it, since target-mips/unicorn.c had do a similar one.
The root cause is:  both little-endian and big-endian are using the same global definition. Then, at the link stage, compiler would report error as redefinition.

Maybe, we can define different REGS_STORAGE_SIZE for different endian, it may looks more beauty.

aquynh: then Mips should fix this issue, too.

for duplicated definition, you can put the related variables into qemu/header_gen.py, then run "make header" to generate new header files, and the problem will be fixed.

xizhizhang:
- yes, I should run "make header", I have create a bad armbe.h.
- I have just pull req this one

Thanks

qemu/target-arm/unicorn_arm.c
aquynh: can you fix this issue, so i can merge the armeb branch to master?

xizhizhang: yeah, I have just fix it. it seems easy.

PR #787, opened 2017-03-17
Fix huge memory leak on uc_mem_protect()

gdbinit:
A memory region is allocated inside split_region() that was only freed in error case but not on success case, leading to huge memory leak if the region size was significant.


aquynh:
merged, thanks!

PR #798, opened 2017-04-17
handle not having a path in python bindings

tylerni7:
Very silly issue: I am using the python bindings, and for various reasons I don't have a PATH set in my environment. Right now I can't `import unicorn` because of that. That is stupid, this makes it better.


aquynh:
merged, thanks!

PR #801, opened 2017-04-21
Added armbe support to MSVC branch.

xorstream:
Main change here was to add a new msvc project for armeb-softmmu and then link that into the unicorn libraries when building.

Some other small changes were:
* Added project file for "sample_armeb"
* Fixed make.sh wrong paths when copying qapi-*.c files
* Fixed make.sh by removing reference to non-existent i386-softmmu
* Added armeb-softmmu to prebuild.bat
* Tidied up internals of msvc project files to keep them consistent


aquynh:
merged, thanks!

PR #808, opened 2017-04-25
Removed MSVC binding.

xorstream:
Removed MSVC bindings since we can now build for MSVC natively.
This should reduce the confusion for new users who want to use unicorn with msvc.


aquynh:
@dzzie, this modifies your VB binding, can you ack? (we now have native MSVC support, so you dont need to use the MSVC binding anymore)

dzzie:
as long as the dll compiles the rest of the code will run with DYNLOAD set or not (DYNLOAD should also be removed from the project properties). I use VS2008 and I think I had to actually regenerate the lib file for it to work for me with a static link. Even though it adds some complexity I actually liked the dynamic loading but I made the code so it would work either way.

dzzie:
we could also keep a copy of the dynamic loader in the vb bindings folder so it doesnt get lost to the ages if anyone cares. I have a version that added null checks after every lookup and sets a char* if it fails to load correctly

https://github.com/fireeye/unicorn-libemu-shim/blob/master/Unicorn/unicorn_dynload.c#L119

xorstream:
@dzzie the old msvc binding was forced to use the "dynamic loading work around" because the files were not built with msvc. I wrote the dynamic loader so that it was as close to using an msvc native dll as possible. The main difference was that you had to call uc_dyn_load() to do the magic.

The new msvc support means that both a static lib and a dynamic dll can be built natively with msvc. So the dynamic loading hack is no longer required. To link statically you would link in unicorn_static.lib. To link dynamically you link in in unicorn.lib and it will load unicorn.dll automatically at runtime.

It is still possible to use the dynamic loading hack with the new msvc dll. It's just that there is no need to do so.

dzzie:
file: ucvbshim.vcproj 
just remove DYNLOAD; from the two places it is set: (debug and release settings)
PreprocessorDefinitions="WIN32;_DEBUG;_CONSOLE;DYNLOAD;"

the vbshim dll and vb6 code will be fine then

aquynh:
merged, thanks!

@dzzie, could you send a pull req to update VB binding now?

dzzie:
ok vb bindings pull req made

PR #811, opened 2017-04-26
Update comments on UC_MEM_*_UNMAPPED

bulaza:
Updated the comment on UC_MEM_*_UNMAPPED to clarify what happens if "true" is returned


aquynh:
merged, thanks!

PR #814, opened 2017-05-02
add arm64 CPACR_EL1 register support

lunixbochs:
This was necessary to enable floating point support on arm64. I wasn't able to do it from inside the emulator, perhaps I needed to be higher than EL1 to set CPACR.


aquynh:
merged, thanks!

PR #816, opened 2017-05-03
arm64eb: arm64 big endian also using little endian instructions.

xizhizhang:
swap fetched instruction code before disassemble it


aquynh:
Very nice, but then the sample code now does not really exercise this big endian mode. Can you change the input binary code to something else accessing big endian data?

aquynh:
merged, thanks!

PR #819, opened 2017-05-04
Implemented read and write access to the YMM registers

saelo:
I've so far only modified the python bindings to work correctly with the YMM registers. The others will likely have to be modified as well.


aquynh:
merged, thanks!

PR #826, opened 2017-05-08
Fix UC_HOOK_INSN handling of invalid instructions

bulaza:
1) clarified usage of uc_hook_add for x86 (only supports IN and OUT)
2) Added a check in uc_hook_add that if x86, only IN or OUT is being hooked. Returns UC_ERR_HOOK if an invalid hook is requested.


uc.c
```diff
@@ -1036,6 +1036,16 @@ uc_err uc_hook_add(uc_engine *uc, uc_hook *hh, int type, void *callback,
         hook->insn = va_arg(valist, int);
         va_end(valist);
 
+
+#ifdef UNICORN_HAS_X86
```
aquynh: in general, we try best to isolate arch-dependent code in uc.c, usually by moving them to qemu/target-xxx/unicorn.c. can you somehow move this code there?

include/unicorn/unicorn.h
```diff
@@ -251,7 +251,7 @@ typedef enum uc_hook_type {
 #define UC_HOOK_MEM_VALID (UC_HOOK_MEM_READ + UC_HOOK_MEM_WRITE + UC_HOOK_MEM_FETCH)
 
 /*
-  Callback function for hooking memory (UC_MEM_READ, UC_MEM_WRITE & UC_MEM_FETCH)
+  Callback function for hooking memory (READ, WRITE, and FETCH)
```
aquynh: i dont see how this new comment is more clear?
same for the comment below on UNMAPPED & PROT.

bulaza: The comments were causing warnings to be emitted from const_generator.py during CI builds.

uc.c
bulaza: I will look into it tomorrow.

include/unicorn/unicorn.h
aquynh: oh i see. can we fix const_generator.py instead, because we may have the same issue in the future?

uc.c
aquynh: there are also few other code in the uc.c that should be isolated in their architectures. while you are at it, if you can fix them all, that will be great!

bulaza: Would a function that calls out to validate the hook for a particular architecture be ok? 

For example something like:

#ifdef UNICORN_HAS_X86
if (arch == UC_ARCH_x86 && !x86_validate_instruction_hook(insn)){
 return error
}
#endif
#ifdef UNICORN_HAS_ARM
if(arch == UC_ARCH_ARM && !arm_validate_instruction_hook(insn){
...
#endif
....

aquynh: would be better if we move everything out, so there is no more "ifdef UNICORN_HAS_X86" in uc.c.

you can think of adding a new function pointer to `uc_struct`. see `include/priv.h` for examples.

include/unicorn/unicorn.h
bulaza: Created an issue for this problem here: https://github.com/unicorn-engine/unicorn/issues/829

bulaza:
I am going to close this so I can cleanly submit two different PRs, one to fix the const_generator.py issue, and another with changes for hook validation

PR #828, opened 2017-05-09
fix arm64 hang (fix #827)

### LINKED ISSUE #827: ARM64 really likes to hang
```
from keystone import *
from unicorn import *

asm = '''
mov x0, 100
l1: sub x0, x0, 1
    bge l1
'''

uc = Uc(UC_ARCH_ARM64, UC_MODE_ARM)
ks = Ks(KS_ARCH_ARM64, KS_MODE_LITTLE_ENDIAN)
sc = ''.join(map(chr, ks.asm(asm)[0]))

uc.mem_map(0x1000, 0x1000)
uc.mem_write(0x1000, sc)
uc.emu_start(0x1000, 0x1000 + len(sc))
```

See also #813

I don't know if my assembly is right, but I can't use Unicorn to check because it hangs forever (and won't call a HOOK_CODE callback). It doesn't seem to be generating any interrupts either.

Edit: I think that should be a `subs` to set the condition flags, but that's unrelated

lunixbochs:


aquynh:
this is interesting! does the latest Qemu still have this bug??

lunixbochs:
this is a bug in added unicorn code for addr_end

aquynh:
merged, thanks!

PR #830, opened 2017-05-11
fix Ruby bindings

misson20000:
I haven't tested whether `query` actually works after this, I just noticed that it was set to `m_uc_hook_del` instead of `m_uc_query`. Maybe someone copy/pasted and forgot to change it?


aquynh:
@sashs, can you ack?

misson20000:
Turns out there were also issues with the Procs we get for adding hooks getting freed. The solution I've implemented works, but [has a memory leak](https://github.com/unicorn-engine/unicorn/pull/830/commits/29f6a4fb0271429abb433da36c096a3f1607b051#diff-48a405038e37e52364d7552341a229c7R430) that I haven't the inclination to figure out how to solve properly right now.

misson20000:
I think the proper way to manage the memory for the hook structs would be to keep a list of `VALUE`s of hook structs.

If we implement  a `mark` function for the hook structs...
```
VALUE m_uc_hook_add(int argc, VALUE* argv, VALUE self){
    ...
    uc_hook_add(...);
    VALUE hook_value = Data_Wrap_Struct(Hook, mark_hook, free, hook);
    ADD_TO_LIST(hook_value); // not sure how you guys want to implement this
}

...

VALUE m_uc_hook_del(int argc, VALUE* argv, VALUE self){
    ...
    uc_hook_del(...);
    REMOVE_FROM_LIST(hook_value);
    ...
}

...

static void mark_hook(void *p) {
    struct hook *hook = (struct hook *)p;
    rb_gc_mark(hook->cb);
    rb_gc_mark(hook->ud);
}
```

And a `mark` function for `UcClass`...

```
VALUE m_uc_initialize(VALUE self, VALUE arch, VALUE mode) {
    ...
    VALUE uc = Data_Wrap_Struct(UcClass, mark_uc, uc_close, _uc);
    ...
}

...

static void mark_uc(void *p) {
    FOREACH_HOOK_VALUE(hook) {
        rb_gc_mark(hook);
    }
}
```

That should prevent any memory issues. I'm not sure what kind of list you guys would like to use, though.

sashs:
Hey,
thx for your work. Are you still working on it? If not I will look at your fix.

misson20000:
I don't think there's anything left that I need to fix, so it should be good to go

misson20000:
Looks like I spoke too soon. The callback `cb_hook_intr` had the wrong signature and would segfault upon being invoked.

sashs:
@aquynh I think that looks good.

aquynh:
merged, thanks!

PR #834, opened 2017-05-13
Aarch64 add thread registers

misson20000:
Pretty similar to #684, this adds support for reading and writing the thread registers (TPIDR_EL0, TPIDRRO_EL0, and TPIDR_EL1)


qemu/target-arm/unicorn_aarch64.c
```diff
@@ -79,6 +79,15 @@ int arm64_reg_read(struct uc_struct *uc, unsigned int *regs, void **vals, int co
                 case UC_ARM64_REG_CPACR_EL1:
                     *(uint32_t *)value = ARM_CPU(uc, mycpu)->env.cp15.c1_coproc;
                     break;
+                case UC_ARM64_REG_TPIDR_EL0:
+                        *(int64_t *)value = ARM_CPU(uc, mycpu)->env.cp15.tpidr_el0;
```
aquynh: these new lines do not align with other blocks in the same switch block. can you fix this?

(just in case, we are using spaces (not tabs) for indentation)

misson20000: They align with all the blocks except for the UC_ARM64_REG_CPACR_EL1, so I'm thinking it's the UC_ARM64_REG_CPACR_EL1 block that's misaligned. It does seem a little strange to me that all the other blocks are indented 9 spaces, given that the rest of the file is indented 4 spaces

aquynh: yes 4 spaces, please.

aquynh:
merged, thanks!

PR #835, opened 2017-05-13
Go: improve hook callback speed by 30% and add a HOOK_CODE benchmark

lunixbochs:
Before: BenchmarkX86Hook-4    2000000    905 ns/op
    After:  BenchmarkX86Hook-4    2000000    683 ns/op


aquynh:
merged, thanks!

PR #837, opened 2017-05-17
Update UC_HOOK_MEM_VALID

bulaza:
UC_HOOK_MEM_READ fires before handlers for invalid memory reads, so UC_HOOK_MEM_VALID would technically also be receiving invalid memory reads. Switching to UC_HOOK_MEM_READ_AFTER ensures that only actually valid reads are hooked

### DIFF
--- a/include/unicorn/unicorn.h
+++ b/include/unicorn/unicorn.h
@@ -248,7 +248,9 @@ typedef enum uc_hook_type {
 // Hook type for all events of illegal memory access
 #define UC_HOOK_MEM_INVALID (UC_HOOK_MEM_UNMAPPED + UC_HOOK_MEM_PROT)
 // Hook type for all events of valid memory access
-#define UC_HOOK_MEM_VALID (UC_HOOK_MEM_READ_AFTER + UC_HOOK_MEM_WRITE + UC_HOOK_MEM_FETCH)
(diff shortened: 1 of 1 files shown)


aquynh:
i get your idea, but this move may break some old code, which we want to avoid.

if you really want this, perhaps it is better to create a new macro, and document its feature.

include/unicorn/unicorn.h
```diff
@@ -248,6 +248,8 @@ typedef enum uc_hook_type {
 // Hook type for all events of illegal memory access
 #define UC_HOOK_MEM_INVALID (UC_HOOK_MEM_UNMAPPED + UC_HOOK_MEM_PROT)
 // Hook type for all events of valid memory access
+// NOTE: UC_HOOK_MEM_READ is triggered before UC_HOOK_MEM_READ_PROT and UC_HOOK_MEM_READ_UNMAPPED, so
+//       this hook will technically trigger on some invalid reads.
```
aquynh: "this hook **will** technically trigger on some invalid reads" -> do you want to change "will" to "may" ?

bulaza:
@aquynh can you rerun the CI checks? It appears that the error in the Travis CI build was not related to my code:

"Error: Fetching /usr/local/Homebrew/Library/Taps/caskroom/homebrew-cask failed!"

aquynh:
merged, thanks!

Issue #727, opened 2017-01-22
MIPS fails to branch when single stepping


cseagle:
When running mips code one instruction at a time, the CPU fails to follow any branches.

The following was placed in a loop:

`uc_emu_start(uc, pc, 0xffffffff, 0, 1);`

Refer to the attached mips_trace.txt and mips_disasm.txt to see details.

In short, branches, jumps and jal at 0x004014AC, 0x004014CC, 0x004014E4, and 0x00401580 are not followed (along with many others.

Links to code may be found below. Build with

`gcc -o mips_test mips_test.c pages.c -lunicorn -lpthread`

This may happen on other architectures as well.

[mips_test.c](https://gist.github.com/cseagle/5cbe2fdfdc287b4035b9fa105d5ae39a#file-mips_test-c)
[pages.c](https://gist.github.com/cseagle/7a1b016b3346f4a9b34374cb4a14ba25#file-pages-c)
[mips_disasm.txt](https://github.com/unicorn-engine/unicorn/files/721641/mips_disasm.txt)
[mips_trace.txt](https://github.com/unicorn-engine/unicorn/files/721642/mips_trace.txt)

smuniz:
I can confirm this issue on MIPS when single stepping.

Any workaround possible?

aquynh:
@smuniz, can you provide a simpler minimal testcase (than the testcase of @cseaglea)?

smuniz:
Yes. Please find attached a compressed Python script. The code (inlined in the script) should return $v0=0x124 as a consequence of performing the JAL's to the two internal functions instead of just ignoring the instructions.

[sample_mips.zip](https://github.com/unicorn-engine/unicorn/files/1075202/sample_mips.zip)

aquynh:
nice, but can you make the input code (MIPS_CODE_EB) as small as possible, so it is easier to debug?

smuniz:
I've attached a more condensed example. By switching the parameter count between 1 and 0 the changed value of $v0 can be observed.

The shellcode in the script belongs to the following C:

```
/* A very simple function to return a constant 0x123. */
#include "stdio.h"

int caller() {
    return 0x123;
}

int main(void) {
    return caller();
}

```

[sample_mips.zip](https://github.com/unicorn-engine/unicorn/files/1075392/sample_mips.zip)

Thanks!

aquynh:
your code is not minimal yet. below is a better input with all irrelevant instructions removed:

```
[...]

PR #840, opened 2017-05-19
add uc_hook_ex() to allow front insertion

lunixbochs:
this is required for sane operation of a tracing interactive debugger with unicorn, as otherwise you can't insert watch/breakpoints before trace hooks fire and you get duplicate data

### DIFF
--- a/bindings/go/unicorn/hook.c
+++ b/bindings/go/unicorn/hook.c
@@ -1,16 +1,14 @@
 #include <unicorn/unicorn.h>
 #include "_cgo_export.h"
 
-uc_err uc_hook_add(uc_engine *uc, uc_hook *hh, int type, void *callback,
-                void *user_data, uint64_t begin, uint64_t end, ...);
+uc_err uc_hook_add_ex(uc_engine *uc, uc_hook *hh, int type, void *callback, bool front, void *user_data, uint64_t begin, uint64_t end, ...);
 
-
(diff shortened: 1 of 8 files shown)


aquynh:
i feel that adding a new API just to change where to put the hooks is not really a neat approach. time to get back to that uc_option() we discussed before, where we can customize how the core works?

lunixbochs:
hook_option is global state for something I need to control *per hook*, which isn't good

aquynh:
i just have a commit https://github.com/unicorn-engine/unicorn/commit/fe466d003a8921e83962da1634eadd4461a0ed17 related to this in the "test" branch, coming out of the issue https://github.com/unicorn-engine/unicorn/issues/727. 

i think we can have a better solution than having this new API uc_hook_ex(). lets discuss how we should extend the current API to allow what you want here.

aquynh:
and looks like you do not initialize list->tail in your list_insert() function.

lunixbochs:
Have we learned nothing from opengl?

This is needlessly stateful. I really don't like it:

uc_option(hook_insert, true)
hook_add()
uc_option(hook_insert, false)

lunixbochs:
I'm retracting this request. I am comfortable tracking and manage hooks independently in my own projects, or maintaining my own unicorn fork, if it comes up again as a requirement.

uc_option cannot ever solve my use case, which is for a debugger in another thread to insert breakpoint hooks. It's a guaranteed race which cannot be fixed, even if uc_hook_add were thread safe.

PR #841, opened 2017-05-20
aarch64: add exception syndrome (ESR) pseudo-register

### LINKED ISSUE #838: AArch64 Exception Syndrome Inaccessible
When an exception is taken, Unicorn will [invoke the interrupt hooks](https://github.com/unicorn-engine/unicorn/blob/513075e061e6642af696380f189d17a84b95241e/qemu/cpu-exec.c#L134), passing `cpu->exception_index` to the `uc_cb_hookintr_t`.

For my use case, I want to hook `SVC` instructions on AArch64, which take an exception to EL1. On an actual piece of hardware, this writes the immediate value provided in the `SVC` instruction to an [exception syndrome register](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.den0024a/CHDEIEBG.html). The QEMU code to do this is in [`aarch64_cpu_do_interrupt`](https://github.com/unicorn-engine/unicorn/blob/513075e061e6642af696380f189d17a84b95241e/qemu/target-arm/helper-a64.c#L489), and I would submit a pull request adding [the exception syndrome registers](https://github.com/unicorn-engine/unicorn/blob/master/qemu/target-arm/cpu.h#L198) to reg_read and reg_write, but `cc->do_interrupt(cpu)` is only run for x86, and on AArch64, it messes with some CPU state that we probably don't want it to.

I'm not sure how exceptions work on other architectures, but for AArch64, I think the best solution would be to add a `uint32_t syndrome` parameter to [`uc_cb_hookintr_t`](https://github.com/unicorn-engine/unicorn/blob/master/include/unicorn/unicorn.h#L171). This would break ABI, but I'm also not sure how much that matters to Unicorn.

misson20000:
Closes #838


aquynh:
This is not a real register, so it is confused. I rather having a new API to read/write CPU context.

misson20000:
Is that the sort of thing `uc_query` is for, and if so, would this be better off as a `uc_query` option?

aquynh:
but "query" is not supposed to write/change context, right?

misson20000:
the exception syndrome shouldn't be set by anything other than the emulator when it encounters an exception, which is why I didn't implement reg_write for it

misson20000:
Unless you're emulating something with more privileges and exception handlers I guess, but then you'd want to set the exception syndrome for a specific exception level (ESR_EL1, ESR_EL2, and ESR_EL3), and those *are* real registers (in the `cp15` struct) so it makes sense to have those in `uc_reg_read` and `uc_reg_write`.

I don't have a way of telling what exception level an exception gets taken to though, so I wouldn't know which of those specific registers to read which is what I wanted a more general exception syndrome pseudo-register for.

SomeoneWeird:
Any update on this?

wtdcode:
Closed due to inactivity.

wtdcode:
But still link this to #1449 if I have more time or some one could assist me.

PR #842, opened 2017-05-20
Add uc_query for architecture

misson20000:
Instead of keeping track of what architecture the engine is, I wanted a way to query the engine's architecture in order to fix the Ruby bindings to only [treat some x86 registers specially](https://github.com/unicorn-engine/unicorn/blob/master/bindings/ruby/unicorn_gem/ext/unicorn.c#L116) if the architecture is actually x86.


aquynh:
merged, thanks!

PR #844, opened 2017-05-25
x86::trans: handle illegal case for opc c6/c7

vardyh:
Reference Intel software developer manual vol2 Appendix A Table A-6 for
detailed decoding information.

Signed-off-by: vardyh <vardyh.dev@gmail.com>

### DIFF
--- a/qemu/target-i386/translate.c
+++ b/qemu/target-i386/translate.c
@@ -6005,9 +6005,15 @@ static target_ulong disas_insn(CPUX86State *env, DisasContext *s,
         ot = mo_b_d(b, dflag);
         modrm = cpu_ldub_code(env, s->pc++);
         mod = (modrm >> 6) & 3;
+        reg = ((modrm >> 3) & 7) | rex_r;
         if (mod != 3) {
+            if (reg != 0)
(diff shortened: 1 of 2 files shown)


aquynh:
can you provide the input triggering this issue?

another question: is this also issue for the latest Qemu version?

vardyh:
@aquynh Did you mean a comment in the code or commit, or a testcase in sample_x86?
Regarding original Qemu, I'll check it out later.

vardyh:
@aquynh Confirmed that original Qemu has the same issue.

aquynh:
Yes, i mean the testcase triggering this issue.

vardyh:
@aquynh testcase committed, it's kinda tricky, but it enumerates all the circumstances.

aquynh:
merged, thanks!

btw, i commented out function test_i386_invalid_c6c7() so CI checks pass. in fact, the tests should be put into tests/ directory, rather than in samples/.

vardyh:
yeah, thanks! :-D
I didn't notice the CI failure after I made the last commit. Sorry about that man :-p

Issue #645, opened 2016-09-27
invalid-shift-exponent  in normalizeFloatx80Subnormal_x86_64 


felixgr:
We found a UBSan error which crashes unicorn. I don't think it has security impact.

`echo '2MbYKLr+uri6ura6usnYAAyn0Ngp2CgMICUnFokS3RQnIbiNyigVedghySHW1dXV1dXVgtXV1dXV1dXVCgoK3g==' | base64 -d > repro_normalizeFloatx80Subnormal_x86_64`

Offending line:

`*zSigPtr = aSig<<shiftCount;`

`unicorn/src/qemu/fpu/softfloat.c:672:20: runtime error: shift exponent 64 is too large for 64-bit type 'uint64_t' (aka 'unsigned long')`

```
#0 in normalizeFloatx80Subnormal_x86_64 unicorn/src/qemu/fpu/softfloat.c:672:20
#1 in addFloatx80Sigs_x86_64 unicorn/src/qemu/fpu/softfloat.c:4996:13
#2 in floatx80_add_x86_64 unicorn/src/qemu/fpu/softfloat.c:5100:16
#3 in helper_fadd_ST0_FT0 unicorn/src/qemu/target-i386/fpu_helper.c:413:11
```

aquynh:
could you provide the source code of the testcase `repro_normalizeFloatx80Subnormal_x86_64`?

aquynh:
do you have this issue with original Qemu?

felixgr:
I didn't check original Qemu.

```
#define ADDRESS 0x10000

struct CleanUpHandle {
  uc_engine *handle;
  ~CleanUpHandle() { uc_close(handle); }
};

extern "C" int LLVMFuzzerTestOneInput(const uint8_t* code_data,
                                      size_t code_size) {
  CleanUpHandle H;
  uc_err err;
  int ecx = 0x12345678;

  err = uc_open(UC_ARCH_X86, UC_MODE_32, &H.handle);
  if (err != UC_ERR_OK) {
    return 0;
  }

  err = uc_mem_map(H.handle, ADDRESS, 0x1000, UC_PROT_ALL);
  if (err != UC_ERR_OK) {
    return 0;
  }

  err = uc_mem_write(H.handle, ADDRESS, code_data, code_size);
  if (err != UC_ERR_OK) {
    return 0;
  }

  err = uc_reg_write(H.handle, UC_X86_REG_ECX, &ecx);
[...]

PR #847, opened 2017-06-09
Remove undefined behaviour with out-of-bounds shift count.

c01db33f:
Testing with sanitizers reports that the two patched locations can be
reached with an out-of-bounds shift count, the patch should not alter
the behaviour of the code except to prevent the undefined behaviour.

### DIFF
--- a/qemu/fpu/softfloat.c
+++ b/qemu/fpu/softfloat.c
@@ -669,7 +669,7 @@ static void
     int8 shiftCount;
 
     shiftCount = countLeadingZeros64( aSig );
-    *zSigPtr = aSig<<shiftCount;
+    *zSigPtr = aSig<<(shiftCount&63);
     *zExpPtr = 1 - shiftCount;
 
 }
--- a/qemu/translate-all.c
+++ b/qemu/translate-all.c
@@ -1420,7 +1420,7 @@ void tb_invalidate_phys_page_fast(struct uc_struct* uc, tb_page_addr_t start, in
         unsigned long b;
 
         nr = start & ~TARGET_PAGE_MASK;
-        b = p->code_bitmap[BIT_WORD(nr)] >> (nr & (BITS_PER_LONG - 1));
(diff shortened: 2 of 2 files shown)


aquynh:
does the latest Qemu code have this issue?

c01db33f:
In latest Qemu, the PageDesc->code_bitmap has type long, so the case in translate-all.c is correctly masked upstream.

The softfloat code seems to have the same issue upstream though, I'll submit a second pull request to them. Does unicorn pull from upstream often, or should I leave the change in here?

aquynh:
No we dont update from stream often, so you should include all the fixes from latest Qemu if possible.

I am travelling, so will be slow to review & merge this, sorry.

aquynh:
which input can trigger this bug?

c01db33f:
Unfortunately the tests that are failing seem to be running a number of different chunks of code, and as it's not my code it's not trivial for me to reduce them without spending a while trying to figure out their code.

Anyway, is seems that one is this issue: https://github.com/unicorn-engine/unicorn/issues/645

The other issue would logically be triggered by self modifying code of some sort, so I had a quick try and sure enough, this code will trigger the issue:

``#define ADDRESS 0x100000
#define CODE "\xb1\x20"         /* mov cl,0x20 */ \
             "\x80\x34\x0f\x41" /* xor byte [rdi + rcx], 0x41 */ \
             "\xe2\xfa"         /* loop 0x2 */

uc_engine* handle;
uint64_t rdi = ADDRESS + 0x10;

uc_open(UC_ARCH_X86, UC_MODE_64, &handle));
uc_mem_map(handle, ADDRESS, 0x1000, UC_PROT_ALL));
uc_mem_write(handle, ADDRESS, CODE, sizeof(CODE) - 1));
uc_reg_write(handle, UC_X86_REG_RDI, &rdi));
uc_emu_start(handle, ADDRESS, ADDRESS + sizeof(CODE) - 1, 0, 1000));

wtdcode:

PR #850, opened 2017-06-20
[RFC] Add support for mapping MMIO through a new function.

### LINKED ISSUE #565: Support registering IO register read/write handlers

kitlith:
Could fix #565.

I don't expect this to be accepted, given complaints before about a new function like this creating inconsistant API, but I figured that I'd polish up and fix the issues with my old PR (#627) before starting on anything else, as I feel like this is the best/cleanest way to do it with the facilities available.

This comes with an example of using the new function created, `uc_mmio_map`.

I have verified that no memory leaks are caused when used in this manner. However, perhaps I need to do something more, as currently some memory is leaked if you do not unmap the IO memory before closing the engine. (maybe I need to add my freeing code to `memory_free` as well as `memory_unmap`?)

### DIFF
--- a/include/uc_priv.h
+++ b/include/uc_priv.h
@@ -66,6 +66,8 @@ typedef void (*uc_args_uc_u64_t)(struct uc_struct *, uint64_t addr);
 
 typedef MemoryRegion* (*uc_args_uc_ram_size_t)(struct uc_struct*,  hwaddr begin, size_t size, uint32_t perms);
 
+typedef MemoryRegion* (*uc_args_uc_ram_size_io_t)(struct uc_struct *uc, ram_addr_t begin, size_t size, uc_cb_mmio_read read_cb, uc_cb_mmio_write write_cb, void *user_data);
+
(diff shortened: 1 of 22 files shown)


kitling:
It appears that I am failing the tests because I haven't implemented the bindings and the equivilant example for the bindings. This might take a while...

wtdcode:

PR #851, opened 2017-06-21
Start moving examples in S files

stephengroat:
allows for easier understanding and editing


aquynh:
with this Appveyor does not pass?

stephengroat:
sorry about that, caught the error

aquynh:
merged, thanks!

PR #855, opened 2017-06-26
target-i386: validate lock prefix for all instructions

vardyh:
According to Intel Developer's Manual:

" The LOCK prefix can be prepended only to the following instructions and
  only to those forms of the instructions
  where the destination operand is a memory operand: ADD, ADC, AND, BTC,
  BTR, BTS, CMPXCHG, CMPXCH8B,
  CMPXCHG16B, DEC, INC, NEG, NOT, OR, SBB, SUB, XOR, XADD, and XCHG. If the
  LOCK prefix is used with one of
  these instructions and the source operand is a memory operand, an
  undefined opcode exception (#UD) may be
  generated. An undefined opcode exception will also be generated if the
  LOCK prefix is used with any instruction not
  in the above list. "

Then it possible to list all available valid opcodes with the help of
Intel Developer's Manual, like so:

ADD         00 01 80/0 81/0 83/0
ADC         10 11 80/2 81/2 83/2
AND         20 21 80/4 81/4 83/4
BTC         0fba/7 0fbb
BTR         0fba/6 0fb3
BTS         0fba/5 0fab
CMPXCHG     0fb0 0fb1
CMPXCH8B    0fc7/1
CMPXCHG16B  0fc7/1
DEC         fe/1 ff/1
INC         fe/0 ff/0
NEG         f6/3 f7/3
NOT         f6/2 f7/2
OR          08 09 80/1 81/1 83/1
SBB         18 19 80/3 81/3 83/3
SUB         28 29 80/5 81/5 83/5
XOR         30 31 80/6 81/6 83/6
XADD        0fc0 0fc1
XCHG        86 87

Signed-off-by: vardyh <vardyh.dev@gmail.com>


aquynh:
cool, i guess upstream Qemu does not handle all the cases yet? any plans to send the patch there?

vardyh:
Upstream Qemu indeed haven't fixed these cases yet. I still may have several fixes on Unicorn to be applied, I'll send the proper patches to Qemu when these are done.

wtdcode:

PR #856, opened 2017-06-29
call handle_mmu_fault from CPUClass

vardyh:
Always fill handle_mmu_fault in CPUClass with arch-specific
handler, and calls from there in tlb_fill().

This brought the possiblity to overriding the builtin
memory management machenisms in the future.

Signed-off-by: vardyh <vardyh@huorong.cn>

### DIFF
--- a/qemu/target-arm/cpu.c
+++ b/qemu/target-arm/cpu.c
@@ -1062,9 +1062,8 @@ static void arm_cpu_class_init(struct uc_struct *uc, ObjectClass *oc, void *data
     cc->cpu_exec_interrupt = arm_cpu_exec_interrupt;
     //cc->dump_state = arm_cpu_dump_state;
     cc->set_pc = arm_cpu_set_pc;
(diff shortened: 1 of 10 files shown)


aquynh:
what is the problem without this?

vardyh:
I may override the builtin memory management in my project, this patch introduces a posibility.

aquynh:
cool, can you elaborate: how do you override the builtin MM while this is not exposed via API? and you are doing this for what purpose?

vardyh:
well, more modifications have to be made and i'm working on it. it's a private project that's gonna be using the uc component and interfaces.

aquynh:
so i understand that there is more pull req following up, otherwise this is no use. then i can wait to merge them all at once.

vardyh:
alright, just hold for now. i made several modifications last night but were not ideal. maybe i'll have to expose some internal interfaces or data structures and if that must be so, i may close the pull request, 'coz the patches will be more specific.

PR #864, opened 2017-08-01
uc_mem_unmap returns proper error code & fixes #863

### LINKED ISSUE #863: uc_mem_unmap not returning expected error code
According to the documentation in [unicorn.h](https://github.com/unicorn-engine/unicorn/blob/master/include/unicorn/unicorn.h#L621-L622) uc_mem_unmap should return UC_ERR_ARG when the specified size is not a multiple of 4KB, however it returns [UC_ERR_MAP](https://github.com/unicorn-engine/unicorn/blob/master/uc.c#L955-L956).

FICTURE7:
It should now return the expected error code when the size specified is not a multiple of 4KB.


aquynh:
merged, thanks!

PR #873, opened 2017-08-22
python: Fix exception in finalizer at exit

bennofs:
Sometimes, the finalizer for an `UcRef` runs so late that the members of the
module have already been set to `None`. We need to make sure that we don't
depend on anything in the module, or we risk getting a Exception when we try
to access the `release_handle` method of `None` (`Uc`).

### DIFF
--- a/bindings/python/unicorn/unicorn.py
+++ b/bindings/python/unicorn/unicorn.py
@@ -251,11 +251,26 @@ class UcCleanupManager(object):
     def register(self, uc):
         ref = UcRef(uc, self._finalizer)
         ref._uch = uc._uch
+        ref._class = uc.__class__
         self._refs[id(ref)] = ref
 
     def _finalizer(self, ref):
+        # note: this method must be completely self-contained and cannot have any references
+        # to anything else in this module.
+        #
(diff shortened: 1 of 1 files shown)


aquynh:
this is interesting, but how can we reproduce this issue?

bennofs:
@aquynh I added a script (with docker, in case that this is python-version specific but you can also try to run it without the script) + source code at https://github.com/bennofs/repros/tree/master/unicorn/%23873

Note that reproducing this is quite fragile. For example, just changing the name of the `angr.state_plugins.unicorn_engine`  module to `angr2.state_plugins.unicorn_engine` or `angr.s.unicorn_engine` makes the error disappear. I suspect that there is some internal hash table and thus the order in which modules are dropped is depending on a hash value.

bennofs:
Sample run:

```
d-cube:/c/r/u/#873 (master=) python2 none-type-error.py 
Exception AttributeError: "'NoneType' object has no attribute 'release_handle'" in <bound method UcCleanupManager._finalizer of <unicorn.unicorn.UcCleanupManager object at 0x7fdc62355d90>> ignored
```

aquynh:
merged, thanks!

PR #875, opened 2017-08-24
Don't clobber EIP after non INT/SYSCALL exception

### LINKED ISSUE #872: Instruction pointer being reset on exceptions
https://github.com/unicorn-engine/unicorn/blob/513075e061e6642af696380f189d17a84b95241e/qemu/cpu-exec.c#L147

In our interrupt callback, we call uc_stop and roll the registers back to the start of the previous block, so that angr can take over execution to handle the runup to the exception natively. However, the instruction pointer is clobbered by this line. I tried moving the register rollback to after uc_start returns, but that caused additional issues. It would be one thing if it were being set to a meaningful address, but in my case, `exception_next_eip` seems to be zero.

What is the purpose of this line? Do we actually want to be clobbering the instruction pointer in the case that the interrupt was handled by a user's callback? How can I accomplish what I want to accomplish?

rhelmot:
Fixes #872

### DIFF
--- a/qemu/cpu-exec.c
+++ b/qemu/cpu-exec.c
@@ -143,8 +143,10 @@ int cpu_exec(struct uc_struct *uc, CPUArchState *env)   // qq
                     }
                     cpu->exception_index = -1;
 #if defined(TARGET_X86_64)
-                    // point EIP to the next instruction after INT
-                    env->eip = env->exception_next_eip;
+                    if (env->exception_is_int) {
+                        // point EIP to the next instruction after INT
+                        env->eip = env->exception_next_eip;
+                    }
 #endif
 #if defined(TARGET_MIPS) || defined(TARGET_MIPS64)
                     env->active_tc.PC = uc->next_pc;


qemu/target-i386/seg_helper.c
```diff
@@ -939,6 +939,7 @@ void helper_syscall(CPUX86State *env, int next_eip_addend)
     CPUState *cs = CPU(x86_env_get_cpu(env));
 
     cs->exception_index = EXCP_SYSCALL;
+    env->exception_is_int = 1;
```
aquynh: i need to look at the code closer, but this worries me, as "exception_is_int" is also used elsewhere, so you set this here may trigger some unwanted behaviour.

aquynh: to be safe (and future-proof), is it better to create a new field in env for this?

rhelmot:
I looked at this a little more carefully and it seems that function I added that line to was guarded by `#ifdef CONFIG_USER_ONLY`, so it wouldn't have done anything. In the corresponding system mode function, changes have already been added for unicorn that hack around this issue by setting env->eip directly, which makes sense since this is meant to be the "fast-syscall" instruction and handled inline.

I pushed the change removing that line, this can be merged now.

aquynh:
merged, thanks!

PR #876, opened 2017-08-28
Changed constatns in ruby gdt example

felberj:
I think those two numbers are wrong, see http://wiki.osdev.org/Global_Descriptor_Table


aquynh:
@sashs, can you ack?

sashs:
Yes, I can ack. Thx!

aquynh:
merged, thanks!

PR #878, opened 2017-08-29
Use the qemu helpers to get/set the x86 eflags

rhelmot:
Previously, every time I tried to get the eflags out of unicorn, it would return zero. This switches to using the methods provided by qemu for correctly manipulating the internal flags to not run afoul of qemu's lazy flags calculation.

I am _extremely_ suspect of the setting of eflags0 - my guess is that it was added as a hack to work around this issue in a different way. Removing the lines setting it causes no regressions in my testing but I have no idea what the effect is deeper in qemu, especially in the translation optimization mechanism.

### DIFF
--- a/qemu/target-i386/unicorn.c
+++ b/qemu/target-i386/unicorn.c
@@ -77,6 +77,7 @@ void x86_reg_reset(struct uc_struct *uc)
     env->eip = 0;
     env->eflags = 0;
     env->eflags0 = 0;
+    env->cc_op = CC_OP_EFLAGS;
 
     env->fpstt = 0; /* top of stack index */
     env->fpus = 0;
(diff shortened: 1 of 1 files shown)


qemu/target-i386/unicorn.c
```diff
@@ -77,6 +77,7 @@ void x86_reg_reset(struct uc_struct *uc)
     env->eip = 0;
     env->eflags = 0;
     env->eflags0 = 0;
+    env->cc_op = CC_OP_EFLAGS;
```
aquynh: without this reset, what is wrong?

rhelmot:
Qemu uses a system of lazy flags calculation where instead of calculating
the processor flags at each step, it stores the last operation necessary to
compute the flags and the operands to that operation. One of these ops, the
one we're using here, just directly uses a given value as the flags. Here,
we set the operation to this value in order to actually use the null flags
we're setting into the state.

The other option would be to use the helper function in this part too, but
I thought this was more in the spirit of the initialization code.

aquynh:
yes i know what that flag is, but not sure it is necessary to reset it there.

can you include a testcase for this issue?

rhelmot:
I have not experienced any issues which I ever tracked down to this behavior, but I don't want to wait for them to find me.

rhelmot:
Actually, having looked into this further - that code is only ever run on `uc_open`. Without this, that field is never initialized anywhere.

So there are two cases - one is that it's genuninely uninitialized and leaving this line out is bad. The other is that it's set to zero somewhere in the depths of the qemu object management code, in which case you don't need to be setting eip, eflags, or eflags0 either.

Either way, it's good to have this line.

aquynh:
merged, thanks!

PR #879, opened 2017-08-30
Revert "Changed constatns in ruby gdt example"

felberj:
Reverts unicorn-engine/unicorn#876

Experimenting with the value seemed like that it is not correct. It might be that the documentation on osdev is wrong or the emulator implementation is different.

@sashs


aquynh:
@sashs, so what is a good fix for this?

sashs:
Hey,
apologize the delay. I will try to check that by next weekend.

lunixbochs:
Usercorn's GDT code is fully working now if you need another reference:

https://github.com/lunixbochs/usercorn/blob/master/go/arch/x86/linux.go#L127-L134
https://github.com/lunixbochs/usercorn/blob/master/go/arch/x86/linux.go#L56

wtdcode:

Issue #853, opened 2017-06-23
ARM Thumb ite instructions don't work when they get interrupted by an instruction limit


flowergrass:
Using the Unicorn instruction limit feature will give invalid emulation of ARM Thumb code when interrupting **ite** instructions (and other variants, it, itte, ittee etc).

In this example only the movne effects the register when running the entire sample code (of 4 instructions), which is the expected result. When stepping through instructions however, the moveq is executed regardless of the **ite** instruction and r2 is overwritten.

```C
#include <stdio.h>
#include <string.h>
#include <unicorn/unicorn.h>

#define CODE "\x9a\x42\x14\xbf\x68\x22\x4d\x22"
#define ADDRESS 0x10000

void run_test(bool stepping) {
    uc_engine *uc;
    uc_err err;

    int sp = 0x1234;
    int r2 = 0, r3 = 1;

    err = uc_open(UC_ARCH_ARM, UC_MODE_THUMB, &uc);
    if (err) {
        printf("Failed on uc_open() with error returned: %u (%s)\n",
                err, uc_strerror(err));
        return;
    }

    uc_mem_map(uc, ADDRESS, 2 * 1024 * 1024, UC_PROT_ALL);

    uc_mem_write(uc, ADDRESS, CODE, sizeof(CODE) - 1);

    uc_reg_write(uc, UC_ARM_REG_SP, &sp);

    uc_reg_write(uc, UC_ARM_REG_R2, &r2);
    uc_reg_write(uc, UC_ARM_REG_R3, &r3);

    if (!stepping) {
        printf("Running entire binary.\n");
        err = uc_emu_start(uc, ADDRESS | 1, ADDRESS + sizeof(CODE) - 1, 0, 0);
        if (err) {
            printf("Failed on uc_emu_start() with error returned: %u\n", err);
        }
    } else {
        int i, addr = ADDRESS;
        printf("Running binary one instruction at a time.\n");
        for (i = 0; i < sizeof(CODE) / 2; i++) {
            err = uc_emu_start(uc, addr | 1, ADDRESS + sizeof(CODE) - 1, 0, 1);
            if (err) {
                printf("Failed on uc_emu_start() with error returned: %u\n", err);
            }
            uc_reg_read(uc, UC_ARM_REG_PC, &addr);
        }
    }

    uc_reg_read(uc, UC_ARM_REG_R2, &r2);
    uc_reg_read(uc, UC_ARM_REG_R3, &r3);

[...]

PR #880, opened 2017-08-30
Fix pausing within Thumb2 ITE blocks (#853)

flowergrass:
A fix to issue #853 where restarting the emulator during ITE instructions would lead to incorrect operations.

While there seems to be code addressing the topic using condexec_bits, locating their underlying issue may be a better fix for another time. This solution ignores tracing instructions within ITE blocks (where condexec_mask should be greater than zero) and on IT instructions which prevents emulation from halting and breaking the code.


aquynh:
interesting, but this is just a stopgap solution. can you have a proper fix for this issue?

flowergrass:
A proper fix for the issue would require a greater understanding of unicorn and qemu itself. We've looked into it but don't have the resources to find another fix since the current solution addresses our issues.

aquynh:
merged, thanks!

dpgeorge:
Thanks for merging!

PR #889, opened 2017-09-13
Arm coproc registers read/write

jbcayrou:
Add ARM feature to read / write coprocessor registers, see #652 for details.

**Work :**
- Coproc registers are automatically generated from ARM XML documentation 
- Use Qemu `ARMCPRegInfo` objects with read/write handlers
- `const_generator.sh` was modified to use intermediary header file with macro expanded (to insert cpregs)


jbcayrou:
Continuous integration fails, I will check that

jbcayrou:
It seems good now, let me know if you want I rebase all commits into one.

mothran:
Whats the status of this @aquynh?  This pull solves my original issue ticket quite well and if this is merged I can move forward with releasing code that depends on it.  Thanks!

aquynh:
i am taking another look now, but this can get merged soon.

jbcayrou:
Hi @aquynh,
Did you have time to check this PR ?
Let me know if you have remarks or suggestions.

RomainQuidet:
Hi anyone still working on this PR ?

RomainQuidet:
@jbcayrou seems there are some conflicts now.

jbcayrou:
Hi Romain,
Sorry I did not work on this PR since my last comment.
I can try to fix conflicts but I am a little busy at the moment.
Did you try to build it on my branch ? Do you have feedbacks, if all work for your use case  ?

RomainQuidet:
Hi, 
I’m studying a SoC based on an arm 946. Standard unicorn fwk through an illegal instruction on the cp15 read/write to set TCM configuration... 
I’ll try to build your branch and test it! 
Do I need to run the script to generate the .h files for my processor before compiling ?

Romain

jbcayrou:
Hum if you have an illegal instruction error with your bytecode I don't think this patch will resolve it. The patch goal is to set/get a coproc register without writing assembly, with `uc_reg_read/uc_reg_write`. Maybe you have to configure the CPU before calling MSR/MRS

You do not need to generate .h files

RomainQuidet:
@jbcayrou I don't know yet, I'm learning this framework.
I've got UC_ERR_INSN_INVALID on this line:
 mcr        p15, #0x0, r1, c9, c1, #0x1

RomainQuidet:
seems that the bug is deeper in qemu armv5 handling. Thanks anyway !

derrekr:
FWIW, there is a bug in the arm_reg_read/write functions, where the CP regs are read uncondionally: https://github.com/jbcayrou/unicorn/commit/1f9f4849a7b5944c4539c0982f16cec0e61823a9#diff-b07e20947cbe5aac5dadd747a74fc6b3R222
This should have been inside the default case of the switch statement. At the moment it causes a nasty performance degradation when these functions are called frequently during emulation. ARM64 might be affected as well, I haven't checked.

wtdcode:
Closed due to inactivity.

Special note: I need @jbcayrou your help to make this PR merged into Unicorn2, especially your scripts to generates all theses registers.

wtdcode:
Re-open this and link to #1217

jbcayrou:
Hi all !

Sorry for the inactivity but it did not work on this patch since several years now.

> Special note: I need @jbcayrou your help to make this PR merged into Unicorn2, especially your scripts to generates all theses registers.

I will rework the patch to make it mergeable.

* The derrekr's comment is right : the CP registers should be fetched in a default case and this PR needs some performance improvement. Another improvement could be to replace the list (ARM/ARM64)_CP_REGS_INFO with a hash map.

* Regarding the issue #1330, X86 specials registers should be accessed with qemu x86 API and this patch only manages ARM coproc registers.

jbcayrou:
Hi !

I rework all the previous patch and I changed the way the coproc register are generated.

- The former patch generated registers with the ARM documentation and many registers were produced. However QEMU implements just some of them and most of registers written by the python script were not supported by QEMU. It was not convenient for users to not know what really work.
This new version is more pragmatic and the generation is done by getting all cpregs registered in a running CPU instance. In this way we can be sure QEMU implements them. The C util `arm_list_coprocregs.c` iterates on registred cpregs to produces a list which is parsed by the python script `gen_arm_cpregs.py`

- To improve the performance of the first patch version, the values of enumeration entries `UC_ARM_REGS_XXXX` match the internal qemu attributes "regid" (and a mask was added to make the distinction with others unicorn enum values) thus read/write access to cpregs are direct.
I removed some cpregs which were already implemented with a direct access to `env.cp15` and now supported in this patch.

Note :

This patch implements the read/write cpreg using `read_raw_cp_reg`/`write_raw_cp_reg`. Theses functions try to call the setter/getter registered with a given cpreg and can perform some processing on the value as add a mask or do some qemu updates (for instance `tlb_flush(CPU(cpu))` in the case of TTBR).
If this change produces issues we can simply change _read_raw_cp_reg_ by _raw_read_ to keep the same behavior than before this patch.

PS : If needed, I moved the old patch version https://github.com/jbcayrou/unicorn/tree/arm_cpregs_access_deprecated

wtdcode:
I really appreciate your contributions! But it looks like you are failing CI?

jbcayrou:
> But it looks like you are failing CI?
Oops sorry :/ I will check why.

wtdcode:
I implemented co processor registers read/write in 3e6665db00cf6a52eb73657d34fb6bade8797cd9 and 8bc1489210f24b2b64b505af048a10da0b25fbea inspired from your work.

Thanks for your contributions!

PR #892, opened 2017-09-16
Ruby: Support reading and writing x86 FPU stack registers

### LINKED ISSUE #891: Ruby bindings: add support for reading and writing FPU stack registers
With the current Ruby bindings, calling `#reg_read(Unicorn::UC_Unicorn::UC_X86_REG_FP0)` (and the rest of the FPU stack registers) causes a crash. I'm assuming this happens because the 80-bit floats present in the registers are larger than the type used to read the register value.

See: https://github.com/unicorn-engine/unicorn/blob/master/bindings/ruby/unicorn_gem/ext/unicorn.c#L173

Similarly, I'm assuming writing proper values to the FPU stack registers is not possible via `#reg_write` since it makes similar assumptions about sizes. It doesn't crash, but it's effectively capped at writing the `sigificand` + 1 bit.

See: https://github.com/unicorn-engine/unicorn/blob/master/bindings/ruby/unicorn_gem/ext/unicorn.c#L225

Reads could be handled with a `float80` struct, converted to a Ruby array.

For ex:

```c
typedef struct uc_x86_float80 {
    uint64_t significand   : 63;
    uint64_t normalized    : 1;
    uint16_t exponent      : 15;
    uint16_t sign          : 1;
} uc_x86_float80;
```

```c
VALUE fp_ary;
uc_x86_float80 f80;

case UC_X86_REG_FP0:
case UC_X86_REG_FP1:
case UC_X86_REG_FP2:
case UC_X86_REG_FP3:
case UC_X86_REG_FP4:
case UC_X86_REG_FP5:
case UC_X86_REG_FP6:
case UC_X86_REG_FP7:
    f80.significand = 0;
    f80.normalized = 0;
    f80.exponent = 0;
    f80.sign = 0;

    err = uc_reg_read(_uc, tmp_reg, &f80);

    if (err != UC_ERR_OK) {
        rb_raise(UcError, "%s", uc_strerror(err));
    }

    fp_ary = rb_ary_new();
    rb_ary_store(fp_ary, 0, ULL2NUM(f80.significand));
    rb_ary_store(fp_ary, 1, UINT2NUM(f80.normalized));
    rb_ary_store(fp_ary, 2, UINT2NUM(f80.exponent));
    rb_ary_store(fp_ary, 3, UINT2NUM(f80.sign));
    return fp_ary;
```

Writes could presumably be handled using the reverse.

fallenoak:
In order to reduce rounding problems from calculations, FPU stack registers for x86 architectures contain values stored in an 80-bit extended precision format.

As a result, reading and writing to these registers requires specific handling.

This update brings the Ruby bindings in line with the Python bindings by supporting reading and writing the FPU stack registers using 2-element arrays: [mantissa, exponent]

The mantissa array element contains the first 64 bits of the FPU stack register.

The exponent array element contains the last 16 bits of the FPU stack register.

Closes #891

### DIFF
--- a/bindings/ruby/unicorn_gem/ext/types.h
+++ b/bindings/ruby/unicorn_gem/ext/types.h
@@ -1,0 +1,25 @@
+/*
+
+Ruby bindings for the Unicorn Emulator Engine
+
+Copyright(c) 2016 Sascha Schirra
+
+This program is free software; you can redistribute it and/or
+modify it under the terms of the GNU General Public License
+version 2 as published by the Free Software Foundation.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program; if not, write to the Free Software
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+*/
+
+typedef struct uc_x86_float80 {
+    uint64_t mantissa;
+    uint16_t exponent;
+} uc_x86_float80;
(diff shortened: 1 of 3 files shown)


sashs:
Hi, it looks good. I will look at it more in detail later today.

sashs:
Ack!

aquynh:
merged, thanks!

PR #895, opened 2017-09-17
renamed unicorn gem to unicorn-engine

sashs:
That should prevent name conflicts with the unicorn webserver gem


aquynh:
does this change break all the existing apps using "unicorn" name?

sashs:
More or less. It is necessary to reinstall the binding.

sashs:
They could use the old version as well. There is no API change, only the name of the gem has changed. In ruby you can still use unicorn.

aquynh:
any solutions that we can keep backward compatibility?

sashs:
I see I forgot something.

fallenoak:
Even if this causes something to break, it seems like it might be worth it. Without the change, it'll never be possible to publish the bindings on `rubygems.org`.

sashs:
@aquynh
Yes it will breaks the apps. They have to replace `require 'unicorn'`  with `require 'unicornengine'`.
I do not know a possibility to ensure backward compatibility. I think this is not possible here. But doing this step is necessary, due to the unicorn webserver.

fallenoak:
@sashs Any chance it'd be possible to keep the `require` name in sync with the gem name?

Currently with this PR, there are three names required to use the bindings:

```
gem install unicorn-engine
require 'unicornengine'
include Unicorn
```

Perhaps the gem could be named `unicornengine` (so at least the gem name and the require string are in sync)?

sashs:
Hey, I thought about that. 
I renamed them now to:

`gem install unicornengine`
`require 'unicornengine'
include UnicornEngine`

aquynh:
What do you think about renaming to
unicorn_engine instead?



On Sep 18, 2017 1:54 AM, "Sascha Schirra" <notifications@github.com> wrote:

Hey, I thought about that.
I renamed them now to:

gem install unicornengine
require 'unicornengine' include UnicornEngine

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/unicorn-engine/unicorn/pull/895#issuecomment-330065775>,
or mute the thread

fallenoak:
@sashs @aquynh According to the RubyGems name style guide, `unicorn_engine` is likely the most conventional name:

http://guides.rubygems.org/name-your-gem/

Gem name | Require statement | Main class or module
-- | -- | --
ruby_parser | require 'ruby_parser' | RubyParser

sashs:
Sure, why not. 
`gem install unicorn_engine`
```
require 'unicorn_engine' 
include UnicornEngine
```

fallenoak:
Fantastic!

aquynh:
merged, thanks!

PR #897, opened 2017-09-22
Cleanups/fixes for the library issue conglomerate

rhelmot:
- I tried to use pkgconfig, but then I realized that this would seriously complicate building for distribution when the distributor has the library is installed globally (likely). I opted for the simpler environment variable approach instead.
- I disabled the distribution of the static library on linux and macos. On windows it is required because even if a DLL is present, the LIB must also be present in order to build against it. Yell at the msvc people, not me.

### DIFF
--- a/bindings/python/README.TXT
+++ b/bindings/python/README.TXT
@@ -8,6 +8,11 @@ from source.
 	This will build the core C library, package it with the python bindings, 
 	and install it to your system.
 
+    If you want to prevent the build of the native library during the python installation,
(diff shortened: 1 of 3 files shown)


aquynh:
this looks nice, but how about Cygwin static lib?

rhelmot:
same situation as with windows for the .lib file. the .a file has never been distributed with the python library.

aquynh:
merged, thanks!

would you do the same thing for Capstone?

PR #899, opened 2017-09-27
add test demonstrating x86-64 segmentation

williballenthin:
This is a test to demonstrate issue #898, segmentation under 64-bit x86.

Its based on `x86_gdt.py`, with the following changes:

```diff
tests/regress# diff x86_gdt.py x86_64_gdt.py                                                        
37c37,38                                                     
< CODE = '65330d18000000'.decode('hex') # xor ecx, dword ptr gs:[0x18]                                                     
---                                                          
> CODE = '6548330C2518000000'.decode('hex')  # x86-64: xor rcx, qword ptr gs:[0x18]                                        
>                                                            
71c72                                                        
<         uc = Uc(UC_ARCH_X86, UC_MODE_32)                   
---                                                          
>         uc = Uc(UC_ARCH_X86, UC_MODE_64)                   
79c80                                                        
<         uc.mem_write(SEGMENT_ADDR+0x18, 'AAAA')            
---                                                          
>         uc.mem_write(SEGMENT_ADDR+0x18, 'AAAAAAAA')        
81c82                                                        
<         gdt_entry = create_gdt_entry(SEGMENT_ADDR, SEGMENT_SIZE, A_PRESENT | A_DATA | A_DATA_WRITABLE | A_PRIV_3 | A_DIR_CON_BIT, F_PROT_32)
---                                                          
>         gdt_entry = create_gdt_entry(SEGMENT_ADDR, SEGMENT_SIZE, A_PRESENT | A_DATA | A_DATA_WRITABLE | A_PRIV_3 | A_DIR_CON_BIT, F_LONG)
91c92                                                        
<         self.assertEqual(uc.reg_read(UC_X86_REG_ECX), 0x41414141)                                                        
---                                                          
>         self.assertEqual(uc.reg_read(UC_X86_REG_RCX), 0x4141414141414141) 
```

currently, the test fails:
```
tests/regress# python x86_64_gdt.py 
0x18L
E
======================================================================
ERROR: test_gdt (__main__.GdtRead)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "x86_64_gdt.py", line 90, in test_gdt
    uc.emu_start(CODE_ADDR, CODE_ADDR+len(CODE))
  File "build/bdist.linux-x86_64/egg/unicorn/unicorn.py", line 315, in emu_start
    raise UcError(status)
UcError: Invalid memory read (UC_ERR_READ_UNMAPPED)

----------------------------------------------------------------------
Ran 1 test in 0.001s

[...]

Issue #898, opened 2017-09-27
x86-64: segmentation not working


williballenthin:
I am not able to emulate 64-bit x86 code that references segmented memory, even after setting up a GDT.

I would like to emulate 64-bit Windows shellcode that references data within the TEB (available in `gs`). To be clear, I do no have much experience setting up a GDT and segmentation, but after around four hours trying different approaches, I'm not having success. I'd like to help the outcome of this bug be either, 1) identify and fix a bug in unicorn, or 2) add a demonstration of segmentation under x86-64.

Regression test incoming.

Nukem9:
http://developer.amd.com/wordpress/media/2012/10/24593_APM_v21.pdf (page 33)

> There are two methods to update the contents of the FS.base and GS.base hidden descriptor fields. The first is available exclusively to privileged software (CPL = 0). The FS.base and GS.base hidden descriptor-register fields are mapped to MSRs. Privileged software can load a 64-bit base address in canonical form into FS.base or GS.base using a single WRMSR instruction. The FS.base MSR address is C000_0100h while the GS.base MSR address is C000_0101h.
>
> The second method of updating the FS and GS base fields is available to software running at any privilege level (when supported by the implementation and enabled by setting CR4[FSGSBASE]). The WRFSBASE and WRGSBASE instructions copy the contents of a GPR to the FS.base and GS.base fields respectively. When the operand size is 32 bits, the upper doubleword of the base is cleared. WRFSBASE and WRGSBASE are only supported in 64-bit mode.

I'm not sure if unicorn is doing any other magic behind the scenes.

williballenthin:
awesome, thanks @Nukem9 

since this looks like user error, i'll close the issue and re-open if there's actually any bug.

redogwu:
Hello, I find the instruction "WRGSBASE" can't work on windows 7(64 bit), could you mind to introduce another way？

Thank you very very very much.

PR #901, opened 2017-09-28
add 64-bit test demonstrating setting MSRs and FS/GS segments

williballenthin:
now that i've learned from #898 and #899  (thanks again @Nukem9), i've put together minimal test cases showing how to set fs and gs on x86-64. this can serve as documentation/example code for future users of unicorn.


aquynh:
merged, thanks!

PR #904, opened 2017-10-06
only build brew HEAD on master and rm trusty jobs

stephengroat:
brew HEAD pulls from master, so it only makes sense to build on master

trusty is now the travis default, explicit jobs are no longer required


aquynh:
merged, thanks!

PR #908, opened 2017-10-12
Note about GNU binutils dependency on Mac OS in *Nix readme

JeremyAgost:
The unit test makefile tries to use "gobjcopy" when assembling tests. This binary isn't present on stock Mac OS X, since it refers to the GNU version of objcopy. It's trivial to install via homebrew (`brew install binutils`) but it isn't mentioned in the Readme for building on *Nix.

I added a note about installing it, as well as cmocka using Homebrew. I also added a note that tests will fail if their architecture support wasn't compiled.

### DIFF
--- a/docs/COMPILE-NIX.md
+++ b/docs/COMPILE-NIX.md
@@ -153,8 +153,14 @@ of corresponding languages.
 
 [8] Unit tests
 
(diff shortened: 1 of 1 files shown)


aquynh:
merged, thanks!

PR #911, opened 2017-10-16
changed gem name to unicorn-engine

sashs:
Hi, 

I changed the name of the gem to unicorn-engine. This name is used when you install the gem from rubygems. 

regards


aquynh:
merged, thanks!

PR #912, opened 2017-10-17
Removed forcing nb_regs to 0 in tcg_reg_alloc_call

jief666:
Forgotten hack ?

### DIFF
--- a/qemu/tcg/aarch64/tcg-target.c
+++ b/qemu/tcg/aarch64/tcg-target.c
@@ -45,10 +45,12 @@ static const int tcg_target_reg_alloc_order[] = {
     /* X30 reserved as temporary */
 };
 
-static const int tcg_target_call_iarg_regs[8] = {
+static const int tcg_target_call_iarg_regs[] = {
     TCG_REG_X0, TCG_REG_X1, TCG_REG_X2, TCG_REG_X3,
     TCG_REG_X4, TCG_REG_X5, TCG_REG_X6, TCG_REG_X7
 };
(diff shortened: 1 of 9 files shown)


4577:
@jief666 Maybe it would be right to add a check for "defined(_MSC_VER)", in addition to "TCG_TARGET_REG_BITS == 32" in the case it would allow tests to succeed?

jief666:
Maybe a generic test about the size of the array ? A host ARM/Windows might not exist yet, but it might soon...

4577:
You have also "defined(_M_ARM)" and "defined(_M_ARM64)" for this

jief666:
Here is the proper way, I think, to resolve the problem of the dummy value in tcg_target_call_iarg_regs array.
I created a var nb_tcg_target_call_iarg_regs, set to 0 in the same case (TCG_TARGET_REG_BITS is not 64). That way tcg.c doesn't have to make any assumption about dummy value or not.

wtdcode:

jief666:
In more than 2 years, you never took my PR, even though aquynh asked me to do it.
And now you're just closing it...
Sorry to say, that's not very encouraging for contributing in the future.

Still, best of luck.

wtdcode:
@jief666 Hello. Thanks for your contribution anyway.

I'm closing PRs since we need to refresh the community to push our new version Unicorn2. You may test your problem on our new version and if the problem still exists, I would have a look and take your PR.

Thanks for your cooperation.

PR #922, opened 2017-12-13
arm64: IP0 is r16 and IP1 is r17

gunyarakun:
According to the resource following, register aliases UC_ARM64_REG_IP0 and UC_ARM64_REG_IP1 are inversely defined.
http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.den0024a/ch09s01s01.html


aquynh:
merged, thanks!

PR #936, opened 2018-01-05
Fix Java Makefile on macOS

jeremysears:
* Fixes the logic for Dylib extensions on macOS.  The logic would previously use the .dll extension instead of .dylib
* Adds an uninstall task and documentation.

### DIFF
--- a/bindings/java/Makefile
+++ b/bindings/java/Makefile
@@ -15,6 +15,9 @@ jar:
 install: lib jar
 	$(MAKE) -f Makefile.build install
 
+uninstall:
+		$(MAKE) -f Makefile.build uninstall
+
 gen_const:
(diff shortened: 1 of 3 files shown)


jeremysears:
Just some minor changes to help out.  The Java bindings build didn't work for macOS.  It would generate a dll instead of a dylib.  While I was at it, I created an uninstall, because it was handy when working out what was going on.  I think it's also generally useful.  I put in some associated doc too.  Since this is all build stuff, I didn't add any tests.  I presume that's reasonable.  Let me know if you'd prefer for me to make any changes.  Great work on this lib.  I've already found it handy.

aquynh:
@cseagle, can you ack?

aquynh:
merged, thanks!

PR #944, opened 2018-02-13
New API to set a code exec barrier

pgarba:
Usefull when you execute code inside a valid image range

uc_err uc_code_barrier(uc_engine *uc, uint64_t barrier_start, uint64_t barrier_end)

barrier_start: Start address of the barrier
barrier_end: End address of the barrier

if barrier_start and barrier_end == 0 then 'No barrier is set'


wtdcode:
Closed due to unicorn2 beta release.

PR #949, opened 2018-03-04
Fix tests when using go 1.10

comawill:
Starting with go 1.10 `go test` also runs `go vet` which leads to failing tests. (Use of `Fatal` instead of `Fatalf` in line 73).
This PR resolves this problem.
Also it removes an unnecessary use of `fmt.Errorf`.


aquynh:
@lunixbochs, can you please ack this?

lunixbochs:
Looks fine.

aquynh:
merged, thanks!

PR #964, opened 2018-05-11
x86_16: set cs register

mf85:
When `begin` address is greater than 64KB, emulation starts at wrong address because only low word is copied.

Example code.
[main.zip](https://github.com/unicorn-engine/unicorn/files/1996247/main.zip)

### DIFF
--- a/uc.c
+++ b/uc.c
@@ -549,8 +549,13 @@ uc_err uc_emu_start(uc_engine* uc, uint64_t begin, uint64_t until, uint64_t time
                 default:
                     break;
                 case UC_MODE_16:
-                    uc_reg_write(uc, UC_X86_REG_IP, &begin);
+				{
+					uint16_t cs = begin / 16;
+					uint16_t ip = begin % 16;
+					uc_reg_write(uc, UC_X86_REG_CS, &cs);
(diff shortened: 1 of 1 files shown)


MoochMcGee:
Oh, so THAT'S why my IBM PC emulator, when using unicorn, didn't work at all. Thanks for this fix!

MoochMcGee:
This is... kinda inaccurate though. You can't fully determine where the code segment begins that you want to start from from just a 32-bit address. TBH, this needs an API change to be fully accurate. If I want to start from FFFF:0000 or F000:FFF0, unicorn has no idea with this patch.

aquynh:
for this scenario, you should setup CS & EIP separately, rather than setting up EIP only, so i dont think this pull req is correct.

MoochMcGee:
why not just change the overall unicorn api to be more generic?

aquynh:
How do you want to change the api?

MoochMcGee:
I think we should add support for things like segmentation in the API.

aquynh:
What is wrong with setting CS register, like what we are doing now?

mf85:
@aquynh
Hi,
"so i dont think this pull req is correct."
should I close it then?

aquynh:
yes, let me close this for you, thanks!

PR #974, opened 2018-07-06
Added a continue function that can be used to continue the emulation …

hansonry:
…after it has been stopped

Hey, I noticed there is no easy way to run say 10 cycles at a time and then go do other stuff. Seems like you would have to call uc_emu_start and keep track of the current address register every time though the loop. I know I could use hooks to stop every 10 cycles, but I would like to run multiple different instances of unicorn simultaneously and asynchronously.

I know it isn't a complete change, I probably need to write tests and update the other language bindings (I have no idea how to do that). But before I jumped in with that stuff I wanted your thoughts.

### DIFF
--- a/include/unicorn/unicorn.h
+++ b/include/unicorn/unicorn.h
@@ -515,6 +515,23 @@ UNICORN_EXPORT
 uc_err uc_emu_start(uc_engine *uc, uint64_t begin, uint64_t until, uint64_t timeout, size_t count);
 
 /*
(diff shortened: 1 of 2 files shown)


aquynh:
this is interesting, but keeping track the PC at the exit of uc_emu_start() is simple enough, that i dont think we need an API for what you want.

@danghvu, what do you think?

wtdcode:

Issue #975, opened 2018-07-06
auto-generation scripts qapi-types.py and qapi-visit.py not found


Sauci:
the build fails on Windows 10 when trying to execute the python scripts *qapi-types.py* and *qapi-visit.py* from file qemu/Makefile, line 147 and 151 (default parallel builds is set to 4).
the problem seems to be the absolute path parameter provided to python.
if I change the code snippet:
```make
qapi-py = $(SRC_PATH)/scripts/qapi.py $(SRC_PATH)/scripts/ordereddict.py

qapi-modules = $(SRC_PATH)/qapi-schema.json $(SRC_PATH)/qapi/common.json

qapi-types.c qapi-types.h :\
$(qapi-modules) $(SRC_PATH)/scripts/qapi-types.py $(qapi-py)
	$(call quiet-command,$(PYTHON) $(SRC_PATH)/scripts/qapi-types.py \
		$(gen-out-type) -o "." -b -i $<, \
		"  GEN   $@")
qapi-visit.c qapi-visit.h :\
$(qapi-modules) $(SRC_PATH)/scripts/qapi-visit.py $(qapi-py)
	$(call quiet-command,$(PYTHON) $(SRC_PATH)/scripts/qapi-visit.py \
		$(gen-out-type) -o "." -b -i $<, \
		"  GEN   $@")
```
to:
```make
qapi-py = $(SRC_PATH)/scripts/qapi.py $(SRC_PATH)/scripts/ordereddict.py

qapi-modules = qapi-schema.json $(SRC_PATH)/qapi/common.json

qapi-types.c qapi-types.h :\
$(qapi-modules) $(SRC_PATH)/scripts/qapi-types.py $(qapi-py)
	$(call quiet-command,$(PYTHON) scripts/qapi-types.py \
		$(gen-out-type) -o "." -b -i $<, \
		"  GEN   $@")
qapi-visit.c qapi-visit.h :\
$(qapi-modules) $(SRC_PATH)/scripts/qapi-visit.py $(qapi-py)
	$(call quiet-command,$(PYTHON) scripts/qapi-visit.py \
		$(gen-out-type) -o "." -b -i $<, \
		"  GEN   $@")
```
the scripts are running correctly.

aquynh:
Could you please make a pull req?

PR #976, opened 2018-07-07
fix invalid script path (#975)

Sauci:


aquynh:
You said this fixes Win10 issue, but will this break other platform?

Sauci:
Not tested, I don’t have an other os right here... is it part of the CI tests performed?

aquynh:
regarding the Python script path, why we dont have this problem with SRC_PATH before?

regarding the new APIs, i am not very sure these are very helpful, as in most cases the align & page size are fixed, and known.

Sauci:
concerning the new API, it is a mistake from my side, reverted.
concerning the ```${SRC_PATH}``` issue, I found that the problem happens when the ```python2-setuptools``` package is not installed in cygwin.I guess this issue should be marked as won't fix? maybe an other one should be opened to update the readme?

aquynh:
Perhapa the best option is to add that missing dependency package to
README. Could you send a pull req for that?

Sauci:
ok, the commited changes have been reverted, and the markdown has been updated.

aquynh:
merged, thanks!

PR #981, opened 2018-07-22
Use python2 rather than python

flyn-org:
I maintain the Unicorn package for Fedora, and I recently found that the package failed to build. This appears due to Fedora's migration to Python 3 [1]. Namely, "Ensure that /usr/bin/python or /usr/bin/env python shebangs are replaced with either /usr/bin/python2 or /usr/bin/python3." PEP 394 [2] describes the over-arching Python plans which are driving this.

This pull request fixes the Fedora build by replacing "python" with "python2" in the top-level Makefile and making use of "--python=python2" when running QEMU's configure script. It also changes the shebangs used throughout Unicorn from "python" to "python2."

[1] https://fedoraproject.org/wiki/FinalizingFedoraSwitchtoPython3
[2] https://www.python.org/dev/peps/pep-0394/

Signed-off-by: W. Michael Petullo <mike@flyn.org>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -215,7 +215,7 @@ all: unicorn
 
 qemu/config-host.h-timestamp:
 	cd qemu && \
-	./configure --cc="${CC}" --extra-cflags="$(UNICORN_CFLAGS)" --target-list="$(UNICORN_TARGETS)" ${UNICORN_QEMU_FLAGS}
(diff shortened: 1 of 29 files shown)


aquynh:
Nice, but this will break all the systems that do not have /usr/bin/python2
linking to /usr/bin/python?

XVilka:
I think this can be safely closed with the demise of Python2.

aquynh:
there are few things:

- we totally removed Python in build process (Makefile). this will be available in the next official release soon.
- sample code run well for both Python2 & Python3
- header generation is only for developers, not for users.

so yes, this PR is not needed anymore. i will close, thanks.

flyn-org:
Thanks!

PR #986, opened 2018-07-28
added x86Msr Functions

felberj:
Added X86MsrWrite and X86MsrRead to the golang bindings.


aquynh:
@lunixbochs, please ack.

lunixbochs:
These should be in the [x86.go](https://github.com/unicorn-engine/unicorn/blob/fb4dc10fe96067b3338cf362d3a27eb1199e0308/bindings/go/unicorn/x86.go) file, and match the other register IO functions (RegWrite*).

Probably like this:
```
RegReadX86Msr(reg uint64) (uint64, error)
RegWriteX86Msr(reg uint64, val uint64) error```

felberj:
Done, please take another look

aquynh:
merged, thanks!

Issue #990, opened 2018-08-02
UC_HOOK_MEM_WRITE not firing on cmpxchg8b


toshiMSFT:
Sample code, tested against the master branch and unicorn 1.0.1 (Windows):

```c
#include <unicorn/unicorn.h>
#include <assert.h>

int read_happened = 0;
int write_happened = 0;

static void hook_mem(uc_engine *uc, uc_mem_type type,
	uint64_t address, int size, int64_t value, void *user_data)
{
	switch (type) {
	default: break;
	case UC_MEM_READ:
		read_happened = 1;
		break;
	case UC_MEM_WRITE:
		write_happened = 1;
		break;
	}
}

#define CODE "\x0F\xC7\x0D\xE0\xBE\xAD\xDE"	// cmpxchg8b [0xdeadbee0]
#define CODE_ADDR 0
#define DATA_ADDR 0xdeadb000

int main(int argc, char **argv, char **envp)
{
	uc_engine *uc;
	uc_hook trace1;
	uint64_t buffer;

	uc_open(UC_ARCH_X86, UC_MODE_32, &uc);
	uc_mem_map(uc, CODE_ADDR, 0x1000, UC_PROT_ALL);
	uc_mem_map(uc, DATA_ADDR, 0x1000, UC_PROT_ALL);
	uc_mem_write(uc, CODE_ADDR, CODE, sizeof(CODE) - 1);
	uc_hook_add(uc, &trace1, UC_HOOK_MEM_READ | UC_HOOK_MEM_WRITE, hook_mem, NULL, 1, 0);

	// memory is initially zero
	uc_mem_read(uc, 0xdeadbee0, &buffer, sizeof(buffer));
	assert(buffer == 0x0000000000000000);

	// mov edx:eax, 0x0000000000000000
	// mov ecx:ebx, 0x4141414141414141
	int zero = 0x00000000;
	int AAAA = 0x41414141;
	uc_reg_write(uc, UC_X86_REG_EDX, &zero);
	uc_reg_write(uc, UC_X86_REG_EAX, &zero);
	uc_reg_write(uc, UC_X86_REG_ECX, &AAAA);
	uc_reg_write(uc, UC_X86_REG_EBX, &AAAA);
	uc_emu_start(uc, CODE_ADDR, CODE_ADDR + sizeof(CODE) - 1, 0, 0);

	// memory was written to at 0xdeadbee0, but no write hook fired!
[...]

PR #992, opened 2018-08-03
UC_HOOK_MEM_WRITE not firing on cmpxchg8b

toshiMSFT:
Ensures that UC_HOOK_MEM_WRITE fires on cmpxchg8b by disabling MMU caching
at a level below where unicorn performs its hooking logic.

Adds a regression test.

Fixes unicorn-engine/unicorn#990

### DIFF
--- a/qemu/include/exec/cpu_ldst_template.h
+++ b/qemu/include/exec/cpu_ldst_template.h
@@ -115,20 +115,12 @@ static inline void
 glue(glue(cpu_st, SUFFIX), MEMSUFFIX)(CPUArchState *env, target_ulong ptr,
                                       RES_TYPE v)
 {
-    int page_index;
     target_ulong addr;
     int mmu_idx;
 
     addr = ptr;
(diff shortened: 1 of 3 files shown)


nanoric:
nice,  how about cmpxchg16b?

toshiMSFT:
@nanoric it is my understanding that both cmpxchg8b and cmpxchg16b call the `glue(glue(cpu_st, SUFFIX), MEMSUFFIX)` function; in fact pretty much every helper function that writes memory seems to call this function.

So my change should work for cmpxchg16b as well

toshiMSFT:
Actually, that reminds me--the corresponding ldq function should also be fixed to not go through the mmu as well (`glue(glue(cpu_ld, USUFFIX), MEMSUFFIX)`). My gut instinct is to remove the mmu accesses from all the functions in this file, but I'm worried that might be a tad excessive...

wtdcode:
To be investigated in uc2. #1217 #1449

wtdcode:
This is already fixed in the dev branch.

PR #994, opened 2018-08-08
Update java bindings for hook bug, multiple closing

EcmaXp:
This pr will fix the hook error and close bug in Java bindings.

## hook bug
Even if you add a hook only to the specified memory area, the hook is called for all other areas of memory. So the assert fails.
```java
uc.hook_add((uc, address, size, value, user) ->
{
    assert 0x40000000 <= address && address <= 0x40000000 + 0x10000;
}, 0x40000000, 0x40000000 + 0x10000, null);
```

This patch integrates all hooks into a single hookMap instead of having multiple hook lists in the Unicorn class. The `invokeXXXCallbacks` function also determines which hooks should be called so that the correct hooks can be executed.

## close bug
In the previous method, there is no way to manually close the Unicorn. If close is set to public, calling it may cause a segmentation error. So, I've added a close field to the Unicorn class to make it work correctly in a single close call. (Still calling the close call multiple times will cause a segmentation error.)

### DIFF
--- a/bindings/java/unicorn/Unicorn.java
+++ b/bindings/java/unicorn/Unicorn.java
@@ -26,50 +26,26 @@ import java.util.*;
 public class Unicorn implements UnicornConst, ArmConst, Arm64Const, M68kConst, SparcConst, MipsConst, X86Const {
 
    private long eng;
+   private boolean closed = false;
    private int arch;
    private int mode;
 
-   private long blockHandle = 0;
-   private long interruptHandle = 0;
-   private long codeHandle = 0;
-
-   private Hashtable<Integer, Long> eventMemHandles = new Hashtable<Integer, Long>();
-   private long readInvalidHandle = 0;
-   private long writeInvalidHandle = 0;
-   private long fetchProtHandle = 0;
-   private long readProtHandle = 0;
-   private long writeProtHandle = 0;
-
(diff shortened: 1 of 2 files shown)


EcmaXp:
After integration testing, I decided to create a new Java binding instead of using the existing Java binding.

aquynh:
@cseagle, can you please ack?

EcmaXp:
Please check https://github.com/EcmaXp/junicorn

- Most of them work well except for register reads / writes.
- Sadly there is no documentation and testing.
- At least in my use case it works fine but I do not know if this implementation is correct due to lack of experience in JNI.

image

aquynh:
do you want to send a pull req to include your binding into https://github.com/unicorn-engine/unicorn/blob/master/bindings/README?

EcmaXp:
There is no plan yet. Compared to previous Java bindings, stability is good, but there is no documentation and there are some features that are not implemented.

cseagle:
@aqyunh This change is really too substantial for me to comment on. If @EcmaXp wants to create a new Java binding, I really have no problem with that. I would suggest that register reads and writes are an essential thing to have working, but that may not be ready to go yet based on what I read above.

wtdcode:
Closed due to inactivity.

Issue #995, opened 2018-08-08
SYSENTER can no longer be hooked


toshiMSFT:
It's not currently possible to hook the SYSENTER instruction. We can verify this by running the SYSENTER regression test in the repo:

```
$ LD_LIBRARY_PATH=. ./tests/regress/sysenter_hook_x86
Cannot hook SYSENTER instruction
```

This is a regression on master, as unicorn-1.0.1 supported this functionality.

PR #996, opened 2018-08-08
Makes SYSENTER hookable on x86

toshiMSFT:
Adds SYSENTER to the whitelist of supported hookable instructions in unicorn
as well as fixes up the existing sysenter_hook_x86 regression test which was
previously failing

Fixes unicorn-engine/unicorn#995


aquynh:
merged, thanks!

PR #1000, opened 2018-08-16
Fuzz

catenacyber:
Integration with oss-fuzz

- One fuzz target per architecture (buffer is pure x86 or ARM, not with options as done in capstone)
- Only fuzzes `uc_emu_start` (seemed relevant)
- Some minor modification to samples/Makefile to make it work


catenacyber:
Looks ok now.
There is a lot of undefined behaviors...
Any idea on how to generate a corpus ?

aquynh:
great, thanks for doing this again!

any change on this, compared with the pull req to OSSFuzz?

aquynh:
what do you think about moving fuzz/ into tests/?

catenacyber:
Good idea :-)

aquynh:
merged, thanks!

please add your name to https://github.com/unicorn-engine/unicorn/blob/master/CREDITS.TXT, too.

PR #1008, opened 2018-09-05
Provide path to good python to qemu configure

Grazfather:
If a system has python on the path point to python3 then the qemu build will fail to configure. This change configures qemu with an explicit path to python2.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -224,7 +224,7 @@ all: unicorn
 
 qemu/config-host.h-timestamp:
 	cd qemu && \
-	./configure --cc="${CC}" --extra-cflags="$(UNICORN_CFLAGS)" --target-list="$(UNICORN_TARGETS)" ${UNICORN_QEMU_FLAGS}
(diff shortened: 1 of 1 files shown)


aquynh:
what happens if python2 is not existent?

btw, do you know that we can always build with customized Python with below command (this is documented in https://github.com/unicorn-engine/unicorn/blob/master/docs/COMPILE-NIX.md

```
$ UNICORN_QEMU_FLAGS="--python=/path/to/python2" ./make.sh
```

Grazfather:
I could change this to `which python2 || which python`. That should make it robust against that case, though I haven't seen that on a modern system.

I am making the change here because your `make.sh` calls Make. I wouldn't be able to automatically install this (which I depend on in a docker file) without manually doing the install.

stephengroat:
i know that i'm basically a broken record on this, but any interest in finally moving away from `make.sh` and to using GNU make to build?

i think that the iOS build is still broken (https://github.com/unicorn-engine/unicorn/issues/574#issuecomment-332688565) and the rest of the logic in `make.sh` is very minor and easy to port over to the makefile

no issues if not, just seems to be more of an issue that it's worth right now

aquynh:
yes if you can do GNU make, please do pull req.

PR #1011, opened 2018-09-14
Mips undefined shift fix

catenacyber:
Found by oss-fuzz

Explicit cast to unsigned int is required in order to avoid undefined behavior on left shift hitting the most significant bit


aquynh:
merged, thanks!

PR #1012, opened 2018-09-14
Prevents abort with m68K

catenacyber:
Raises exception instead

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=10331


catenacyber:
And https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=10332

catenacyber:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=10336

aquynh:
merged, thanks!

PR #1014, opened 2018-09-14
Avoids abort with ARM

catenacyber:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=10337

Is this the right way to avoid abort ? (generate an exception)


catenacyber:
Hmm... this is not good

wtdcode:

PR #1016, opened 2018-09-14
Sparc increase ttl number

catenacyber:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=10335

Testcase is "fa7c01ff"


aquynh:
merged, thanks!

PR #1021, opened 2018-09-21
Uses latest qemu arm thumb load store stuff

catenacyber:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=10420

@aquynh :
Should there be a more global rebase on latest qemu code ?
And the other way around, should we push our changes to qemu ?

### DIFF
--- a/qemu/target-arm/translate.c
+++ b/qemu/target-arm/translate.c
@@ -8532,34 +8532,29 @@ static void disas_arm_insn(DisasContext *s, unsigned int insn)  // qq
                 }
             } else {
                 int address_offset;
-                int load;
+                int load = insn & (1 << 20);
+                int wbit = insn & (1 << 21);
+                int pbit = insn & (1 << 24);
+                int doubleword = 0;
                 /* Misc load/store */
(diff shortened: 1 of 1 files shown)


catenacyber:
> did you port from the latest Qemu v3.0?

Hmm, I just looked at the code 
https://github.com/qemu/qemu/blob/master/target/arm/translate.c
So that looks like master above 3.0 

> yes, to both.

So what is the best way to do it ?
I guess that we should first import emu's fixes before submitting the ones found by oss-fuzz
I guest it would be to `git rebase` unicorn on top of qemu 3.0
But how can we do that ?

aquynh:
i am fine with any way. we can report bugs to Qemu independently.

aquynh:
for the record, could you put here the commit of Qemu that you ported code from?

catenacyber:
Well, I used version https://github.com/qemu/qemu/commit/d00584b7cfab3e38f8264bebfcb6fdb02c887ade but this is not the commit which fixed the bug (there were many as I understand)

aquynh:
merged, thanks!

PR #1024, opened 2018-09-26
Use latest qemu to avoid abort

catenacyber:
As in #1021 this patch is inspired by current qemu version
This one avoids a later abort found by
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=10459

### DIFF
--- a/qemu/target-arm/translate-a64.c
+++ b/qemu/target-arm/translate-a64.c
@@ -2619,9 +2619,9 @@ static void disas_ldst_single_struct(DisasContext *s, uint32_t insn)
         } else {
             /* Load/store one element per register */
             if (is_load) {
-                do_vec_ld(s, rt, index, tcg_addr, MO_TE + scale);
+                do_vec_ld(s, rt, index, tcg_addr, scale);
             } else {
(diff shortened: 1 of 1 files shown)


aquynh:
is this from the same commit https://github.com/qemu/qemu/commit/d00584b7cfab3e38f8264bebfcb6fdb02c887ade ?

qemu/target-arm/translate-a64.c
```diff
@@ -2619,9 +2619,9 @@ static void disas_ldst_single_struct(DisasContext *s, uint32_t insn)
         } else {
             /* Load/store one element per register */
             if (is_load) {
-                do_vec_ld(s, rt, index, tcg_addr, MO_TE + scale);
```
aquynh: i suppose that since MO_TE is removed, something elsewhere must be modified too, or this port is incomplete?

catenacyber:
The qemu commit looks to be just that change : https://github.com/qemu/qemu/commit/0a97c40f8e7172ac3d8db97fb22d0ef3025de307

wtdcode:

Issue #1028, opened 2018-10-02
uc_mem_protect duplicating data when dealing with uc_mem_map_ptr memory region


gbmaster:
I am experiencing a bug with uc_mem_protect while trying to change the protection to a subset of a memory area I allocated with uc_mem_map_ptr:

|--------- whole mem_map_ptr area ---------|
      |--- change prot area ----|

In fact, when calling split_region, the content of the subset of the memory region that is supposed to change protection is duplicated and is not pointing anymore to the originally mapped area.

While this is perfectly legit in a normally mapped area scenario, I think that the newly created MemoryRegion should still point to the old area of data.

The patch should be quite easy to implement, but, while I was trying to organize one, I wasn't able to find a simple way to understand, from uc.c, if the MemoryRegion is associated to existing host memory. Also,
it does not seem easy to me to obtain the host memory address without going through QEMU's code.

Any suggestion?

gbmaster:
A minimal example that reproduces this issue is:

```
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

#include "unicorn/unicorn.h"
#include "unicorn/x86.h"

int main(void)
{
	static const size_t buffer_size = 4 * 4 * 1024;
	static const size_t map_address = 0x400000;
	static const size_t read_byte_here = 0x401000;
	
	unsigned char *buffer;
	unsigned char single_byte;
	uc_engine *uc;
	uc_err err;

	buffer = (unsigned char *)malloc(buffer_size);
	memset(buffer, 0, buffer_size);

	err = uc_open(UC_ARCH_X86, UC_MODE_32, &uc);
	if (err != UC_ERR_OK)
	{
		printf("Unable to initialize the engine: %u\n", err);
		return -1;
	}

	err = uc_mem_map_ptr(uc, map_address, buffer_size, UC_PROT_READ | UC_PROT_WRITE, buffer);
	if (err != UC_ERR_OK)
	{
		printf("Unable to map the buffer: %u\n", err);
		return -1;
	}

	printf("Before changing protection: 0x%X\n", (uint16_t)buffer[read_byte_here - map_address]);
	uc_mem_protect(uc, read_byte_here, (buffer_size >> 1), UC_PROT_READ | UC_PROT_WRITE | UC_PROT_EXEC);

	err = uc_mem_write(uc, read_byte_here, "A", 1);
	if (err != UC_ERR_OK)
	{
		printf("Unable to read from memory: %u\n", err);
		return -1;
	}
[...]

PR #1030, opened 2018-10-02
Don't duplicate a uc_mem_map_ptr'd memory region when splitting it (#1028)

gbmaster:
I prevent the duplication of data when splitting memory regions by checking the ram block flag associated to the MemoryRegion struct.

I actually don't like the fact that I need to loop on the uc->ram_list.blocks to find the required block, but I was not able to find any other way to do this.


aquynh:
can you provide an example triggering this issue?

gbmaster:
I updated the issue #1028 with an example reproducing the issue. Let me know if you need some more information about it.

aquynh:
merged, thanks!

PR #1034, opened 2018-10-05
Haskell bindings: use ExceptT instead of deprecated EitherT

puffnfresh:


aquynh:
@adrianherrera, can you please ack?

adrianherrera:
Looks good to me!

aquynh:
merged, thanks!

PR #1037, opened 2018-10-23
Add support for virtual memory access, with Python bindings

stevielavern:
The existing `uc_mem_read` and `uc_mem_write` allow to access the emulator's physical memory. For simple emulated code where the MMU is not setup, this also corresponds to virtual memory. However, there is currently no easy way to access virtual memory for complex code that enables pagination.

I propose to add new APIs to allow users to access virtual memory: `uc_virtual_mem_read` and `uc_virtual_mem_write`. Their prototype is the same than the physical variants. I also added the corresponding Python bindings.


aquynh:
this breaks CI, would you please look at that: https://travis-ci.org/unicorn-engine/unicorn/jobs/444999394?

is there any applications for these APIs yet?

wtdcode:

PR #1047, opened 2018-11-29
i386: set MSR IA32_EFER to correct value at init for IA32e Mode

BrunoPujos:
When in IA-32e mode (x86 64) the value of the MSR  IA32_EFER (0xC0000080) should have LMA and LME bits set. The following code allow to test the issue:

```python
from unicorn import *

CODE = b'\xb9\x80\x00\x00\xc0\x0f2' # mov rcx, 0xC0000080 ; rdmsr

mu.mem_map(0x2000, 0x1000)
mu.mem_write(0x2000, CODE)
mu.emu_start(0x2000, 0x2000 + len(CODE))

print("IA32_EFER = 0x{:X}".format(mu.reg_read(UC_X86_REG_RAX)))
```


aquynh:
merged, thanks!

PR #1049, opened 2018-12-19
Abort on unsupported tlb_flush

aut0:
Added assert(false) when unsupported aarch64 instructions are used. This could have saved me a quite some time.


aquynh:
but this makes your app exit, which is not suitable for a framework.

we need to have a better approach, that gracefully returns from uc_emu_start(), and reports error instead.

aut0:
I understand your concern. I still think this is an improvement over just doing nothing and leaving the user in the dark. The commit that changed this behavior looks like the commented code was just copy pasted in these locations, since the same code does not make sense in all these locations. I think this was just an oversight. I could put another TODO there that leads to this discussion and it could be further improved in the future. From my own experience it really takes a lot of time to debug the behavior that can be introduced in an application run on the emulator that thinks it flushed the tlb but really did not.

wtdcode:

Issue #1053, opened 2019-01-29
tcg_gen_movcond problem


dmarxn:
tcg_gen_movcond (i32, i64) doesn't have the right checks for ALWAYS and NEVER conditions.
tcg_gen_setcond and tcg_gen_brcond does.

This code:
```
66 83 c8 0a          or     ax,0xa
0f 43 c8             cmovae ecx,eax
```

Causes an abort in tcg_out_brcond2():

> unicorn\qemu\tcg\i386\tcg-target.c:1037: tcg fatal error

Fixed at pull request : #1054

PR #1054, opened 2019-01-29
Fix for tcg_gen_movcond

dmarxn:
Fixed conditional move tcg bug (tcg_gen_movcond), which generated an  ALWAYS / NEVER condition despite QEMU not supporting those conditions in the tcg_out part

#1053


aquynh:
this is interesting. how does the latest Qemu handle this case?

dmarxn:
It handles it the same way.
https://github.com/qemu/qemu/blob/b4fbe1f65a4769c09e6bf2d79fc84360f840f40e/tcg/tcg-op.c#L812

aquynh:
merged, thanks!

Issue #1005, opened 2018-08-29
X86_16: STI and POPF don't set IF


sards3:
In X86 16-bit mode, if I run code like
`sti`
or
```
mov ax, 0x200 ;0x200 is the interrupt flag
push ax
popf
```

The interrupt flag does not get set. This should work in 16 bit real mode unless I'm missing something. Am I missing something?

Issue #252, opened 2015-11-07
How does Unicorn see our code as? Userland/Kernel/None?


farmdve:
The reason is because I want to use it to emulate x86 code in Windows and get 1:1 behaviour, but there are a few things that prevent this.

One thing is that Windows has what is known as Structured Exception Handling and Vectored Exception Handling. Basically when anything happens that isn't allowed, division by zero or execution of privileged instructions(like hlt or mov dr0, eax) an exception handler at the top of the seh frame is executed, if it doesn't handle this exception, the other one is called and so forth. 

This however does require us to be able to change EIP from within hooked code.

Also, the EFLAGS behaviour is different under Unicorn than an OS, https://ideone.com/WQAvk1

When attempting to set all flags(except trap flag) we should get a value of eflags that is 0x00244ED7, in Unicorn it's just 0x00000ED7. However, if the trap flag is also set, in Windows, an exception is generated, and if there is an Exception Handler installed, it will be called.

aquynh:
it is ring0, but actually this does not really matter as you have to emulate everything.

in the case you are asking, SEH can be handled by your callback handler: whenever something that you are interested happen (such as div by zero), your handler will get executed and emulate whatever you want.

for the problem with EFLAGS, can you provide a testcase exposing the issue?

farmdve:
For eflags, I've made a pull request on the test case.

As for the SEH handler, it cannot, sadly. The handler will also need to run in Unicorn, not as callback, but as emulated instructions. E.g env->eip needs to be set to the handler.

aquynh:
i think this problem can be solved simply by allowing to change EIP inside the callback?

@cseagle, do you also have to deal with this issue in your IDA plugin?

farmdve:
it should. There is a more...difficult way, restarting Unicorn, copying over the old memory to the new instance, and changing EIP to the handler, executing handler.

However we also need to be able to hook more privileged/illegal instructions. I personally want to hook CPUID,RDTSC,UD2, and when my callback is called, being able to set EIP to some handler.

aquynh:
to hook CPUID & RDTSC, you can do that now: just hook every instruction, then use disassembler (such as Capstone) to check for the instruction ID, then act accordingly.

i still need to check UD2, as i am not quite sure how Unicorn deals with it now.

farmdve:
Wouldn't hooking all instructions be kind of slow?

Right now unicorn deals with UD2 by exiting Unicorn


[...]

PR #1057, opened 2019-02-04
changed cpu_compute_eflags to use the updated eflags variable.

dmarxn:
There seem to be several issues with eflags.
#1005 Trying to set eflags with sti may fail. eflags0, which will not be updated by sti, rewrites eflags when cpu_compute_eflags uses it.  
This also happens with popfl and that's one reason why regress/eflags_noset.c doesn't pass (see issue #252).

This is similar to the fix proposed in issue #313 by @farmdve , d773548 solves the problem he was facing.


farmdve:
I really thought this was fixed after so many years.

aquynh:
merged, thanks!

PR #1059, opened 2019-02-08
Added MXCSR register, fixed writing to FPUCW.

### LINKED ISSUE #1058: X86: writing to fpucw doesn't change floating point rounding mode
Using uc_reg_write with UC_X86_REG_FPCW changes the register's value but doesn't affect rounding.
The reason is there is no subsequent call to set_float_rounding_mode().

The correct way is using cpu_set_fpuc() instead of changing the value directly.

dmarxn:
Added support for reading from and writing to the MXCSR register.
Added the correct way to write to FPUCW
Fixes #1058

Will add tests in the future.


aquynh:
merged, thanks!

PR #1064, opened 2019-02-20
Fixed the decoding of opcodes after getting vex2 using 0xc5

### LINKED ISSUE #1062: vmovdqu xmm0, xmmword ptr [edi] reads xmm word at esi
Hi, I recently find the instruction "vmovdqu xmm0, xmmword ptr [edi]" (bytes 0xc5fa6f07), will wrongly read an xmm word at esi (instead of edi, where it should). When the error occurs, the command causes an error UC_ERR_READ_UNMAPPED. edi is 0x56557020 at the moment. However, I checked the address it is reading and find it is 0x5be0cd19. 

At first, I have no idea where is this value from. It is only after I check all the registers' values did I find it is exactly the value of esi. I think this is a bug of Unicorn because the instruction bytes and disassembly agree with each other. If you have any difficulty reproducing this, please let me know and I can make a PoC.

dmarxn:
Fixes #1062.
Problem was that when decoding an instruction with the VEX prefix, the extension byte was not added.
The fix appears in upstream qemu.

### DIFF
--- a/qemu/target-i386/translate.c
+++ b/qemu/target-i386/translate.c
@@ -5145,7 +5145,7 @@ static target_ulong disas_insn(CPUX86State *env, DisasContext *s,
             rex_r = (~vex2 >> 4) & 8;
             if (b == 0xc5) {
                 vex3 = vex2;
-                b = cpu_ldub_code(env, s->pc++);
+                b = cpu_ldub_code(env, s->pc++) | 0x100;
             } else {
 #ifdef TARGET_X86_64
                 s->rex_x = (~vex2 >> 3) & 8;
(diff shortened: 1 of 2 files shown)


dmarxn:
It seems the cygwin enviornment in AppVeyor has an ImportError when trying to import pkg_resources.
Not sure if it's an enviornment issue.

aquynh:
Please can you provide a testcase?

dmarxn:
Again, AppVeyor has "ImportError: No module named pkg_resources" in the Cygwin enviornment.

aquynh:
merged, thanks!

Waterman178:
thanks

PR #1067, opened 2019-02-27
Add support for the ARM IPSR register.

### LINKED ISSUE #1065: Access to ARM IPSR
Hello,

I'm working on an emulator which includes an NVIC interrupt controller as part of the system to be emulated.  I have a sufficient implementation of NVIC for my project, but it seems unicorn lacks a public API for manipulating the IPSR register.

I'm not super familiar with the internal workings of unicorn, and I'd like my emulator to not have to depend on private APIs or structures.

I've attached a patch with a solution which works for me.
[unicorn.patch.txt](https://github.com/unicorn-engine/unicorn/files/2890667/unicorn.patch.txt)

cfrantz:
1. Create an enum name for the IPSR register.
2. Implement read and write of the IPSR via the xpsr helper functions.

Fixes #1065

### DIFF
--- a/include/unicorn/arm.h
+++ b/include/unicorn/arm.h
@@ -133,6 +133,7 @@ typedef enum uc_arm_reg {
     UC_ARM_REG_C13_C0_2,
     UC_ARM_REG_C13_C0_3,
 
+    UC_ARM_REG_IPSR,
     UC_ARM_REG_ENDING,		// <-- mark the end of the list or registers
 
     //> alias registers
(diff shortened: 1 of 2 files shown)


aquynh:
@stephengroat what do you think about Cygwin build that fails now?

stephengroat:
@aquynh looks like `setuptools` might be missing, i think i fixed in #1068

aquynh:
merged, thanks!

PR #1078, opened 2019-03-31
Update help description for make.sh

lzutao:


make.sh
```diff
@@ -136,6 +157,6 @@ case "$1" in
   "linux32" ) build_linux32;;
   "msvc_update_genfiles" ) msvc_update_genfiles;;
   * )
-    echo "Usage: $0 ["`grep '^  "' $0 | cut -d '"' -f 2 | tr "\\n" "|"`"]"
```
aquynh: the old solution is nice because it is quite future proof, so i feel that we better keep that approach?

lzutao: I think keeping description for each build option in `usage` is quite easy.
If this PR is merged, everyone would take a look in `usage` function first, then 
edit `usage` description appropriate.

make.sh
```diff
@@ -3,21 +3,42 @@
...
+    cat 1>&2 <<EOF
+make.sh - The build script for unicorn engine
+USAGE:
+    $ ./make.sh [OPTIONS]
+OPTIONS:
+                            Build the project
+    asan                    Build for ASan
+    install                 Install the project
+    uninstall               Uninstall the project
+    macos-universal         Build universal binaries on macOS
```
lzutao: I don't know if this description is correct or not.

make.sh
```diff
@@ -136,6 +157,6 @@ case "$1" in
   "linux32" ) build_linux32;;
   "msvc_update_genfiles" ) msvc_update_genfiles;;
```
lzutao: This one I have no idea about its description. Could you give a suggestion?

aquynh: this is to support MSVC build, which cannot automate this copying step itself.

make.sh
aquynh: yes this is correct

make.sh
```diff
@@ -3,21 +3,42 @@
...
-  UNICORN_ASAN=yes
-  ${MAKE} V=1
+  env UNICORN_DEBUG=yes UNICORN_ASAN=yes "${MAKE}" V=1
 }
 
 # build iOS lib for all iDevices, or only specific device
 build_iOS() {
-  IOS_SDK=`xcrun --sdk iphoneos --show-sdk-path`
-  IOS_CC=`xcrun --sdk iphoneos -f clang`
+  IOS_SDK=$(xcrun --sdk iphoneos --show-sdk-path)
```
aquynh: @radare what do you think, i am unsure if this is a good fix, that is portable across all shells?

lzutao:
Also a little bike-shed question: What keeps us from using python3 to build unicorn instead of python2?
Is it QEMU or LLVM part? If not, do you have a wish to convert to python3?

make.sh
aquynh: agreed, let see if there is any other comments on Shell coding.

make.sh
lzutao: I think it is portable. http://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#tag_02_06_03

aquynh:
Qemu part. You are welcome to send in PR to support Python3!

make.sh
aquynh: bash? but we do not depend on a specific shell, so please make sure this works with other non-bash shells too.

lzutao: I think all POSIX shell should follow [the opengroup](https://en.wikipedia.org/wiki/The_Open_Group) specification above.

aquynh:
merged, thanks!

PR #1079, opened 2019-03-31
support for YMM registers ymm8-ymm15

Lukas-Dresel:
Accessing the ymm8-ymm15 registers so far only returned 0 because they were not handled.

Should I add the following registers the same way? 

- [ ] xmm8-15
- [ ] ymm-registers
- [ ] zmm-registers


aquynh:
merged, thanks!

yes, feel free to send PRs to support other registers.

PR #1086, opened 2019-05-12
add CMakeLists.txt

firodj:
I'am trying to make unicorn to be able to build with cmake. 

Current configuration is unicorn-static and for x86_64 arch only.


aquynh:
Wow this is awesome!

Quick question:

- does this work on Windows now?
- how much more work to support all architectures?

Thanks!

firodj:
Yes it works on Windows, I'am using on VS2017, and VS2019, so actually the CMakeLists.txt based on exsiting unicorn.sln. For other architecture I think the effort is same, but need to refactor to be able generate the config-host and config-client.

aquynh:
this is a very good start! do you plan to continue working on this to support the remaining architectures?

firodj:
yes, I'am planning to support all platform if no one trying to do. :D

aquynh:
looking fwd to that!

aquynh:
any progress?

i do hope that you continue this work, or somebody pick up from here, so we can build on Windows, without having to use Visual Studio project.

aquynh:
see https://github.com/unicorn-engine/unicorn/pull/1134, which seems to be more complete

aquynh:
close this because we already have Cmake support now, thanks!

PR #1091, opened 2019-06-04
msvc: Use default platform toolset

LukeUsher:
Updates the MSVC projects to compile with the Default Platform tool set for the installed compiler, rather than a hard requirement for v110_xp.

This allows Unicorn to compile with current versions of Visual Studio, without requiring additional installation of an older tool set.

### DIFF
--- a/msvc/samples/mem_apis/mem_apis.vcxproj
+++ b/msvc/samples/mem_apis/mem_apis.vcxproj
@@ -27,26 +27,26 @@
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="Configuration">
     <ConfigurationType>Application</ConfigurationType>
     <UseDebugLibraries>true</UseDebugLibraries>
-    <PlatformToolset>v110_xp</PlatformToolset>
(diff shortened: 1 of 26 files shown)


LukeUsher:
Any feedback on this?
The appveyor fail was caused by a broken mingw/msys installation on the appveyor box, not a result of the changes in this PR.

rhelmot:
Hey - I'm trying to build with this and even though it builds the samples crash very badly. Is this happening for you as well?

rhelmot:
Actually - this is fascinating. When I build for Win32, everything works fine, but when I build for x64, things crash the first time we try to do a longjmp to exit the cpu loop. Do you have any idea what could be up here?

stephengroat:
@LukeUsher if you push an empty commit, the other appveyor issue has been fixed and you should get a ✅

aquynh:
merged, thanks!

LukeUsher:
> Actually - this is fascinating. When I build for Win32, everything works fine, but when I build for x64, things crash the first time we try to do a longjmp to exit the cpu loop. Do you have any idea what could be up here?

As others found, this was a problem with an incorrect compile flag, it was not directly related to this PR and existed in builds before this PR was created.

PR #1098, opened 2019-06-27
Removed hardcoded CP0C3_ULRI

xwings:
also added
- UC_MIPS_REG_CP0_CONFIG3
- UC_MIPS_REG_CP0_USERLOCAL


xwings:
Also fix a brk PC return issue, After the first brk and completed the second brk, it will always return to previous brk's next pc. We added in PC+4 to force after completion of second brk. code provied by w1tcher from Dubhe.

qemu/target-mips/op_helper.c
```diff
@@ -47,6 +47,10 @@ static inline void QEMU_NORETURN do_raise_exception_err(CPUMIPSState *env,
         cpu_restore_state(cs, pc);
     }
 
+    if(exception == 0x11){
```
aquynh: nice work, just a bit fix here about coding style: we need a space between `if` and `(`, and another space between `)` and `{`.

aquynh:
merged, thanks!

PR #1099, opened 2019-06-28
Add ARM64 Android support in make.sh

Cyuubi:
I noticed that building for Android ARM64 was not an option in make.sh, so here it is.


aquynh:
merged, thanks!

Issue #1066, opened 2019-02-26
Release ?


Summus6:
Hello :)

The last release (1.0.1) was commited near 2 years ago.
When can we expect a tag for a (quite) stable version ?
Maybe the current master is stable enough ?

Thank you a lot for the work you're doing !

anatol:
+1

And please sync the latest qemu version from upstream before the unicorn release.

rhelmot:
I can do a 1.0.1.post1 release to pip based on the current master. This would at the very least make unicorn pip-installable for macos mojave users. @aquynh?

tiffanyb:
I really hope you could release a new version on PyPI since the current version doesn't work on Mac Mojave due to i386 deprecation. It's been fixed in github, and all you need is to update it to PyPI.

aquynh:
Anyone having Visual Studio can confirm that MSVC still compiles?

yhql:
@aquynh I had to update the toolchain in the project files to v141 (MSVC 2017) because the v110_xp is not available anymore. It did compile correctly, but I didn't test it a lot.

Should I make a PR ?

Summus6:
Hello, how it is ?

disconnect3d:
Bump.

Summus6:
Hello

domenukk:
Any news on this? At least master is definitely more mature than ye olde version for my use case. Even a prerelease on pip might help.

domenukk:
This looks good https://pypi.org/project/unicorn/1.0.2rc1/

PR #1107, opened 2019-07-30
Specify python for building QEMU on macOS

twizmwazin:
This is necessary for use in python3 virtualenvs. macOS doesn't have a python2 binary, so we have to hardcode the path here to circumvent the virtualenv.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -172,6 +172,8 @@ LIBRARY_SYMLINK = lib$(LIBNAME).$(EXT)
 endif
 endif
 
+UNICORN_QEMU_FLAGS += --python=$(shell which python2 || which /usr/bin/python || which python)
+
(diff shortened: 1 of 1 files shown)


twizmwazin:
Alright, @rhelmot and I have been discussing the best way to approach it, so this is actually much more generalized, and should cover most *nix systems, Linux, macOS, and presumably most others.

aquynh:
merged, thanks!

rhelmot:
@aquynh if you send me a copy of the windows prebuilt libraries based on the current master I can make an alpha-release on pypi which will satisfy the needs of a lot of users, as per https://github.com/unicorn-engine/unicorn/issues/1066 and https://github.com/angr/angr/issues/1683

aquynh:
yes, emailed you.

Issue #1041, opened 2018-11-10
multiple UC_HOOK_MEM callbacks for unaligned access


RYDB3RG:
It seems to fire multiple read/write callbacks for unaligned access. Is this expected?

```
void hook_mem(uc_engine *uc, uc_mem_type type,
    uint64_t address, int size, int64_t value, void *user_data)
{
    uint64_t rip;
    uc_reg_read(uc, UC_X86_REG_RIP, &rip);
    printf("%p %s %p %d\n", rip, (type == UC_MEM_WRITE) ? "write" : "read ", address, size);
}

int main(int argc, char **argv)
{
    const uint8_t code[] = {
        "\x48\x8b\x00"  //  mov         rax,qword ptr [rax]  
        "\x48\x89\x01"  //   mov         qword ptr [rcx],rax  
        "\xcc"
    };

    uc_engine *uc;
    uc_open(UC_ARCH_X86, UC_MODE_64, &uc);
    uc_mem_map(uc, 0, 0x10000, UC_PROT_ALL);
    uc_mem_write(uc, 0, code, sizeof(code) - 1);

    uc_hook hh;
    uc_hook_add(uc, &hh, UC_HOOK_MEM_READ | UC_HOOK_MEM_WRITE, hook_mem, nullptr, 1, 0);

    const uint64_t addr = 0x1fff;
    uc_reg_write(uc, UC_X86_REG_RAX, &addr);
    uc_reg_write(uc, UC_X86_REG_RCX, &addr);
    uc_emu_start(uc, 0, sizeof(code) - 1, 0, 2);

    return 0;
}

```

produces:

```
0000000000000000 read  0000000000001FFF 8
0000000000000000 read  0000000000001FF8 8
0000000000000000 read  0000000000002000 8
0000000000000003 write 0000000000001FFF 8
0000000000000003 write 0000000000002006 1
0000000000000003 write 0000000000002005 1
0000000000000003 write 0000000000002004 1
0000000000000003 write 0000000000002003 1
0000000000000003 write 0000000000002002 1
0000000000000003 write 0000000000002001 1
0000000000000003 write 0000000000002000 1
0000000000000003 write 0000000000001FFF 1
```

firodj:
it seems the hook called for read mem by the helper_le_ld_name in softmmu_template.h before the unaligned and also reread again by spanning into aligned address. (3 calls)

I think it shouldnot called the hook when there is no actual read on mem, (the 1st call).

[...]

PR #1113, opened 2019-08-02
fix recursive UC_HOOK_MEM callbacks for cross pages access

BrunoPujos:
Here is a PR for #1041 (multiple UC_HOOK_MEM callbacks for unaligned access) which concerns recursive calls to memory access functions, when an access is made between two pages:
* A `size_recur_mem` field is added to the `uc_struct` for keeping the size already treated in the previous hook.
* Fix the recursive calls for the `UC_HOOK_MEM_READ` , `UC_HOOK_MEM_READ_AFTER` and `UC_HOOK_MEM_WRITE` by calling them only the first time.
* When a read memory access is made (`helper_le_ld_name` and `helper_be_ld_name`): the other hook (unmap, prot) are still triggered but with the updated size in the recursive call.
* When a write memory access is made (`helper_le_st_name` and `helper_be_st_name`): the other hook (unmap, prot) are still triggered but with a size of one (because one recursive call is made for each byte).


Waterman178:
thanks

chenhuitao:
@BrunoPujos It's a nice patch. Could you resolve the conflict?

aquynh:
i resolved the conflicts

aquynh:
merged, thanks!

PR #1124, opened 2019-08-22
check arguments, return error instead of raising exceptions.

### LINKED ISSUE #1117: Segementation fault when setting FS register to any value above 0x3
When initializing the FS segment register to anything above 0x3, unicorn segfaults. This doesn't happen when unicorn is initialized with `UC_MODE_64`. 

```
Python 3.7.3 (default, Jun 24 2019, 04:54:02) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.6.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from unicorn import *                                                                                                           
In [2]: from unicorn.x86_const import *                                                                                                 
In [3]: uc = Uc(UC_ARCH_X86, UC_MODE_32)                                                                                                
In [4]: uc.reg_write(UC_X86_REG_FS, 0x3)                                                                                                
In [5]: uc.reg_write(UC_X86_REG_FS, 0x4)                                                                                                
```

chenhuitao:
close #1117.
Writing the segment registers directly probably cause the emulator crash, and python can not catch the exceptions which runs in C.
Add _uc_check_cpu_x86_load_seg() before run cpu_x86_load_seg(), check the arguments , return error instead of raising exceptions. Python will check the return value, raise local language exception on error.

### DIFF
--- a/qemu/target-i386/cpu.h
+++ b/qemu/target-i386/cpu.h
@@ -1137,6 +1137,10 @@ void cpu_x86_load_seg(CPUX86State *s, int seg_reg, int selector);
 void cpu_x86_fsave(CPUX86State *s, target_ulong ptr, int data32);
 void cpu_x86_frstor(CPUX86State *s, target_ulong ptr, int data32);
 
+/*  the binding language can not catch the exceptions.
+    check the arguments, return error instead of raise exceptions. */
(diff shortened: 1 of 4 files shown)


Waterman178:
thanks

PR #1125, opened 2019-08-22
check arguments, return error instead of raising exceptions. 

### LINKED ISSUE #1117: Segementation fault when setting FS register to any value above 0x3

chenhuitao:
close #1117.
Writing the segment registers directly probably cause the emulator crash, and python can not catch the exceptions which runs in C.
Add _uc_check_cpu_x86_load_seg() before run cpu_x86_load_seg(), check the arguments , return error instead of raising exceptions. Python will check the return value, raise local language exception on error.

### DIFF
--- a/qemu/target-i386/cpu.h
+++ b/qemu/target-i386/cpu.h
@@ -1137,6 +1137,10 @@ void cpu_x86_load_seg(CPUX86State *s, int seg_reg, int selector);
 void cpu_x86_fsave(CPUX86State *s, target_ulong ptr, int data32);
 void cpu_x86_frstor(CPUX86State *s, target_ulong ptr, int data32);
 
+/*  the binding language can not catch the exceptions.
+    check the arguments, return error instead of raise exceptions. */
+int _uc_check_cpu_x86_load_seg(CPUX86State *env, int seg_reg, int sel);
+
 /* you can call this signal handler from your SIGBUS and SIGSEGV
    signal handlers to inform the virtual CPU of exceptions. non zero
    is returned if the signal was handled by the virtual CPU.  */
--- a/qemu/target-i386/cpu.h
+++ b/qemu/target-i386/cpu.h
@@ -1139,7 +1139,7 @@ void cpu_x86_frstor(CPUX86State *s, target_ulong ptr, int data32);
 
 /*  the binding language can not catch the exceptions.
     check the arguments, return error instead of raise exceptions. */
-int _uc_check_cpu_x86_load_seg(CPUX86State *env, int seg_reg, int sel);
+int uc_check_cpu_x86_load_seg(CPUX86State *env, int seg_reg, int sel);
 
 /* you can call this signal handler from your SIGBUS and SIGSEGV
    signal handlers to inform the virtual CPU of exceptions. non zero
--- a/qemu/target-i386/seg_helper.c
+++ b/qemu/target-i386/seg_helper.c
@@ -1540,7 +1540,7 @@ void helper_ltr(CPUX86State *env, int selector)
 }
 
 // Unicorn: check the arguments before run cpu_x86_load_seg().
-int _uc_check_cpu_x86_load_seg(CPUX86State *env, int seg_reg, int sel)
+int uc_check_cpu_x86_load_seg(CPUX86State *env, int seg_reg, int sel)
 {
     int selector;
     uint32_t e2;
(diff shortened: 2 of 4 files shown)


chenhuitao:
close and re-open for AppVeyor checking.

Waterman178:
thanks

qemu/target-i386/seg_helper.c
```diff
@@ -1539,6 +1539,87 @@ void helper_ltr(CPUX86State *env, int selector)
...
+    uint32_t e2;
+    int cpl, dpl, rpl;
+    SegmentCache *dt;
+    int index;
+    target_ulong ptr;
+
+    if (!(env->cr[0] & CR0_PE_MASK) || (env->eflags & VM_MASK)) {
+        return 0;
+    } else {
+
```
aquynh: please remove this blank line

qemu/target-i386/seg_helper.c
```diff
@@ -1539,6 +1539,87 @@ void helper_ltr(CPUX86State *env, int selector)
...
+            if (seg_reg == R_SS
+#ifdef TARGET_X86_64
+                && (!(env->hflags & HF_CS64_MASK) || cpl == 3)
+#endif
+                ) {
+                return UC_ERR_EXCEPTION;
+            }
+            return 0;
+        } else {
+
```
aquynh: and this blank line

qemu/target-i386/seg_helper.c
```diff
@@ -1539,6 +1539,87 @@ void helper_ltr(CPUX86State *env, int selector)
...
+            }
+
+            if (!(e2 & DESC_P_MASK)) {
+                if (seg_reg == R_SS) {
+                    return UC_ERR_EXCEPTION;
+                } else {
+                    return UC_ERR_EXCEPTION;
+                }
+            }
+
```
aquynh: and this blank line

qemu/target-i386/cpu.h
```diff
@@ -1137,6 +1137,10 @@ void cpu_x86_load_seg(CPUX86State *s, int seg_reg, int selector);
 void cpu_x86_fsave(CPUX86State *s, target_ulong ptr, int data32);
 void cpu_x86_frstor(CPUX86State *s, target_ulong ptr, int data32);
 
+/*  the binding language can not catch the exceptions.
+    check the arguments, return error instead of raise exceptions. */
+int _uc_check_cpu_x86_load_seg(CPUX86State *env, int seg_reg, int sel);
```
aquynh: how about removing the prefix _ in function name?

uc.c
```diff
@@ -371,12 +371,13 @@ uc_err uc_reg_read_batch(uc_engine *uc, int *ids, void **vals, int count)
 UNICORN_EXPORT
 uc_err uc_reg_write_batch(uc_engine *uc, int *ids, void *const *vals, int count)
 {
+    int ret = UC_ERR_OK;
     if (uc->reg_write)
-        uc->reg_write(uc, (unsigned int *)ids, vals, count);
+        ret = uc->reg_write(uc, (unsigned int *)ids, vals, count);
     else
-        return -1;  // FIXME: need a proper uc_err
+        return UC_ERR_EXCEPTION;  // FIXME: need a proper uc_err
```
aquynh: we have UC_ERR_ARG, but that may not be a good choice. 

what can be a better error code for this case?

qemu/target-i386/cpu.h
chenhuitao: Function name with underscore prefix means it is reserved, not public. Keep it is better.

aquynh: i know, but this is definitely not C convention ;-)

we never use this convention anywhere in Unicorn code, either.

uc.c
chenhuitao: In current definition of error code, I think the UC_ERR_EXCEPTION is the best. If this function failed, you have to catch the exception or terminate program.

qemu/target-i386/cpu.h
chenhuitao: O.K. Remove the underscore.

aquynh:
merged, thanks!

PR #1129, opened 2019-08-28
Handle serialization of cpu context save

cbayet:
The `context_save`, `context_update` and `context_restore` API of Unicorn are using directly C pointers in the bindings, that makes the result of a `context_save` unserializable.

The goal here is to be able to save the context of a cpu, write this state on the disk, and load this cpu context in another emulator of the same architecture and mode.
In simple words, be able to have a CPU snapshot across emulators.

This PR is a fix for the Python bindings. It doesn't break the API for other bindings.

Since the cpu context is just a big buffer, we can let the Python allocate the buffer, and pass it to `uc_context_save` function that will fill it with the cpu context data.
To be able to do this, we need to expose the `cpu_context_size` function to the bindings.

We can then let the Python handle the destruction of this object and we don't need to free a C pointer manually.

The context returned by the `context_save` API is a Python string that is serializable.

With this PR, the `uc_context_alloc` function is not needed anymore in Python bindings.

Tested on X86 emulation.


aquynh:
merged, thanks!

Issue #1127, opened 2019-08-26
Crash when deleting a hook in the hook callback


cbayet:
Hi,

The following sample will make crash unicorn:

```
#!/usr/bin/python

from unicorn import Uc, UC_HOOK_CODE,UC_ARCH_X86, UC_MODE_64
# from unicorn.x86_const import 

code = "\x90"*10
addr = 0x1000

mu = Uc(UC_ARCH_X86, UC_MODE_64)

mu.mem_map(0x1000, 0x1000)
mu.mem_write(addr, code)


class MyHook(object):
    def __init__(self, mu, addr):

        self.mu = mu
        self.hook = mu.hook_add(UC_HOOK_CODE, lambda *args: self.my_hook(*args), begin=addr, end=addr)

    def my_hook(self, *args):
        print("Inside hook ! Removing...")
        self.mu.hook_del(self.hook)
        print("Hook deleted !")


h = MyHook(mu, 0x1001)

print("Starting emulation...")
mu.emu_start(addr, addr+len(code))
print("Emulation done !")
```

The problem come from the `uc_hook_del`method (https://github.com/unicorn-engine/unicorn/blob/master/uc.c#L1148) that doesn't have any ref count check or lock check on the elements of the list `uc->hooks`.

The `list_remove` is done anyways, freeing the list element. The element is still accessed in the  hook loop in `helper_uc_tracecode`, and is dereferenced to fetch the next hook, inducing a Use After Free situation and making the emulator crash.

First, the refs counter on the hooks should probably be incremented before calling its callback and decremented after.

Then, I guess we should have a list for hooks to remove that is called at a safe moment, and that will not remove the hooks that still have refs.

aquynh:
if you can fix, please PR

cbayet:
Fixed in #1130

PR #1130, opened 2019-08-29
Fix UAF when deleting hook while in hook callback #1127

cbayet:
Fix for the issue #1127 

An attribute "to_delete" has been added to the `uc_hook` structure, to delay the modification of the list of hooks.

A check on the attribute `to_delete` has been added in every `HOOK_FOREACH` loop, and in the `helper_uc_tracecode` function
This is also possible by turning the macro `HOOK_FOREACH` into the following code:

```
#define HOOK_FOREACH(uc, hh, idx)                         \
    for (                                                 \
        ({cur = (uc)->hook[idx##_IDX].head; while(cur != NULL && cur->data && ((struct hook *)cur->data)->to_delete) {cur = cur->next;} cur;});                 \
        cur != NULL && ((hh) = (struct hook *)cur->data)  \
            /* stop excuting callbacks on stop request */ \
            && !uc->stop_request;                         \
        ({ do {cur = cur->next;} while(cur != NULL && cur->data && ((struct hook *)cur->data)->to_delete); }))
```
This is not very elegant so I added the check manually everywhere it's needed instead.

A list of hooks to delete has been added to the `uc_struct`.
The deleted hooks are actually freed and removed from the list when the emulation stop.
This is done in the function `clear_deleted_hooks`.

Maybe you can point me a better place to safely remove the hooks ? Do it only at the end of the emulation works fine, but can decrease performance in a run when a lot of hooks are added and removed.

The following testcase doesn't crash anymore and is working as intented:

```
#!/usr/bin/python

from unicorn import Uc, UC_HOOK_CODE, UC_HOOK_MEM_READ, UC_HOOK_MEM_WRITE, UC_HOOK_MEM_FETCH, UC_ARCH_X86, UC_MODE_64
from unicorn.x86_const import UC_X86_REG_RAX

code = "\x90"*10 + "\x48\x8b\x08" + "\x90"*10

CODE_ADDR = 0x1000
DATA_ADDR = 0x3000

mu.mem_map(CODE_ADDR, 0x1000)

mu.mem_map(DATA_ADDR, 0x1000)
mu.mem_write(DATA_ADDR, "A"*0x1000)

mu.mem_write(CODE_ADDR, code)

mu.reg_write(UC_X86_REG_RAX, DATA_ADDR)

code_hook = None
rw_hook = None
count = 0

def code_callback(uc, *args):
    global count
    
    print("Inside hook ! count is {}".format(count))
    if count == 1:
        mu.hook_del(code_hook)
        print("Hook deleted !")

    count += 1
    
def fetch_callback(uc, *args):
    print("any mem access callback !")

def read_and_write(uc, *args):
    print("hook read and write !!")
    mu.hook_del(rw_hook)

mu.hook_add(UC_HOOK_MEM_FETCH | UC_HOOK_MEM_READ | UC_HOOK_MEM_WRITE, fetch_callback)

rw_hook = mu.hook_add(UC_HOOK_MEM_READ | UC_HOOK_MEM_WRITE, read_and_write)

code_hook = mu.hook_add(UC_HOOK_CODE, code_callback)

print("Starting emulation...")
mu.emu_start(CODE_ADDR, CODE_ADDR+len(code))
print("Emulation done !")
```


cbayet:
I've noticed there would be a collision with #1132 .
Since this PR adds a HOOK_FOREACH loop, it would need to add the check the "to_delete" attribute of the hook before callling it.

The problem would be resolve if instead of adding the check everywhere, we would change the macro HOOK_FOREACH as shown in the first comment, but as I said I don't find the code very elegant.

What do you think ?

aquynh:
i merged #1132, which conflicts with this PR, as you pointed out.

lets see how to update this PR.

cbayet:
I merged and added the proper check on the new `HOOK_FOREACH`

Everything seems good :)

aquynh:
merged, thanks!

PR #1131, opened 2019-08-30
brew ignores CC variable

stephengroat:
remove brew check with gcc and don't use `compiler` option in homebrew test, it isn't respected


Waterman178:
thanks

aquynh:
merged, thanks!

PR #1132, opened 2019-08-30
Added an invalid instruction hook

Azertinv:
I needed a way to hook invalid instructions in Unicorn to emulate them so I added it.
This patch should not modify any previous behavior with interrupt hooks and I used the same technique previously used to detect invalid instructions exceptions.
I made the python bindings but I don't know how to support the rest of the bindings.

Here is a small example in python for my implementation:
```python
from unicorn import *

uc = Uc(UC_ARCH_X86, UC_MODE_64)

"""
    nop
    nop
    xsaves  byte ptr [rcx]
    <garbage>
"""

sc = "90900FC729FFFFFF".decode("hex")
uc.mem_map(0, 0x1000)
uc.mem_write(0, sc)

def hook0(uc, user_data):
    addr = uc.reg_read(UC_X86_REG_RIP)
    print("hook0 at 0x{:x}".format(addr))
    return False

def hook1(uc, user_data):
    addr = uc.reg_read(UC_X86_REG_RIP)
    print("hook1 at 0x{:x}".format(addr))
    uc.reg_write(UC_X86_REG_RIP, addr+3)
    return True

uc.hook_add(UC_HOOK_INSN_INVALID, hook0)
uc.hook_add(UC_HOOK_INSN_INVALID, hook1)
uc.emu_start(0, 20, 5000)
```

result:
```
hook0 at 0x2
hook1 at 0x2
hook0 at 0x5
hook1 at 0x5
```


Waterman178:
mark

aquynh:
looks nice! can you please add some sample code for this into samples/?

aquynh:
merged, thanks!

PR #1136, opened 2019-09-03
Two fixes to get unicorn build better with MSVC.

ltfish:
- typename, class, and class are all keywords for MSVC. Adding a suffix to all of them to avoid potential conflicts.
- Switching program database options for DEBUG builds from /ZI to /Zi to avoid crashes at longjmps (see [explanation](https://developercommunity.visualstudio.com/content/problem/282418/unicorn-library-crashes-when-calling-longjmp-and-c.html)).

Successfully compiled and tested under VC 2017 (15.9) after upgrading the build target to v141_xp. Confirmed with @rhelmot and got her approval.

### DIFF
--- a/msvc/samples/sample_x86/sample_x86.vcxproj
+++ b/msvc/samples/sample_x86/sample_x86.vcxproj
@@ -116,6 +116,7 @@
       <RuntimeLibrary>MultiThreadedDebug</RuntimeLibrary>
       <AdditionalIncludeDirectories>../../../include</AdditionalIncludeDirectories>
       <ExceptionHandling>false</ExceptionHandling>
+      <DebugInformationFormat>ProgramDatabase</DebugInformationFormat>
     </ClCompile>
     <Link>
       <SubSystem>Console</SubSystem>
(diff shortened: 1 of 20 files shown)


aquynh:
@rhelmot used to have some issues with building on Windows. can you build & run sample_x86.exe well for both 32bit & 64bit?

rhelmot:
I can confirm this fixes the issues we were seeing. You can also merge #1091 now.

aquynh:
merged, thanks!

chenhuitao:
By default, the Visual C++ compiler treats all files that end in .c as C source code, and all files that end in .cpp as C++ source code. There is no C++ source file, and old source works fine, so that I think adding the suffix underline is odd.

rhelmot:
Correct. However, on fish’s machine it did not work without the change
while on mine it did. There is clearly some sort of MSVC global
configuration which causes this problem to arise and it would be best to
nip it now. It’s pretty common in C only codebases (mono comes to mind) to
treat C++ keywords as though they were C keywords.

chenhuitao:
There is no need to add the suffix underline.
I suggest to revert.
I think using macro is better on bad MSVC machine temporarily.
#ifdef _MSC_VER
#define class class_
#define typename typename_
#endif

PR #1139, opened 2019-09-04
Final tweaks to MSVC build

rhelmot:
- Quote filepaths in post-build scripts so it can work for dumbasses like me with spaces in their usernames
- Update README with more maintenance advice

After this we can release 1.0.2.

### DIFF
--- a/msvc/README.TXT
+++ b/msvc/README.TXT
@@ -5,15 +5,33 @@ Zak Escano  -  January 2017
 These notes are to help myself and others with the upkeep of the msvc native port
 of unicorn-engine.
 
+
+:: Command line build instructions
+
+  msbuild -m -p:Configuration=Release -p:Platform=Win32
+  msbuild -m -p:Configuration=Release -p:Platform=x64
+
+Then bundle and release the folder "distro".
+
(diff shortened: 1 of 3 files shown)


aquynh:
merged, thanks!

@stephengroat, can we run "msbuild" in Appveyor, and even better, keep its artifacts?

stephengroat:
@aquynh looks likes it's already being run

do you know what artifacts you want to keep? i'm currently keeping unicorn.dll https://github.com/unicorn-engine/unicorn/blob/master/.appveyor.yml#L65

aquynh:
so this unicorn.dll already natively built with msvc. 

the goal is that users can download & use the artifacts to build their own tools on Windows (without having to build latest unicorn themselves).

do we already provide the link to the artifacts in our docs yet?

stephengroat:
pr specific or based off of last successful master build?

https://ci.appveyor.com/api/projects/aquynh/unicorn/artifacts/unicorn.dll?branch=master&job=CC%3Dmsvc

is a permalink to the unicorn dll for the last successful job on master for msvc

PR #1140, opened 2019-09-06
Remove warnings

chenhuitao:
Tested on Windows with VS2019.
Tested on Linux with gcc-7.4.0.
Includes the changes of "include/unicorn/platform.h" and "qemu/include/qemu/compiler.h" which are on the cmake branch.


aquynh:
merged, thanks!

Waterman178:
mark

PR #1141, opened 2019-09-09
add cmake

stephengroat:


aquynh:
we still have some issues with Appveyor?

biot:
The second and third patch here are wrong: they add cmake as a C compiler, causing appveyor to invoke it with a C compiler's arguments (-o and -c) which cmake doesn't know.

No idea how to fix it, sorry -- I don't have a Windows machine.

aquynh:
we already had this, thanks!

PR #1144, opened 2019-09-14
Fixes #1143

### LINKED ISSUE #1143: Installing python bindings fails when cross compiling for win64
I compiled unicorn from source using the MSYS2 method described in https://github.com/unicorn-engine/unicorn/blob/master/docs/COMPILE-WINDOWS.md for 64 bits Windows.
The resulting files are
- libgcc_s_seh-1.dll
- libwinpthread-1.dll
- unicorn.dll

I copy these to bindings/python/prebuilt.

However the [python bindings setup.py](https://github.com/unicorn-engine/unicorn/blob/0551b56633f658ec760eac54c14712d712b746d7/bindings/python/setup.py#L71) expects a `unicorn.lib`, which is not produced by the MSYS2 method. Further in the script [it says](https://github.com/unicorn-engine/unicorn/blob/0551b56633f658ec760eac54c14712d712b746d7/bindings/python/setup.py#L185) that `unicorn.lib` isn't even required.

AeonLucid:
https://github.com/unicorn-engine/unicorn/issues/1143 for more information.

### DIFF
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -152,11 +152,9 @@ def build_libraries():
 
     # check if a prebuilt library exists
     # if so, use it instead of building
-    if os.path.exists(os.path.join(ROOT_DIR, 'prebuilt', LIBRARY_FILE)) \
-            and (STATIC_LIBRARY_FILE is None \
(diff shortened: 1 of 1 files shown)


aquynh:
@rhelmot if you dont have any comments, i will merge this.

rhelmot:
Approved

aquynh:
merged, thanks!

PR #1145, opened 2019-09-19
compatible with python2 and python3

chenhuitao:
compatible with python2 and python3.
tested on Ubuntu-18.04-amd64 with pyhton-2.7.15+ and python-3.6.8.


aquynh:
merged, thanks!

PR #1146, opened 2019-09-22
Update Makefile: move default python2 search to lower priority

xwings:
move default python2 search to lower priority


aquynh:
merged, thanks!

PR #1147, opened 2019-09-24
fix a mem-leak

chenhuitao:
fix oss-fuzz issue 10334: unicorn/fuzz_emu_x86_32: Direct-leak in g_malloc.


aquynh:
this breaks both Travis & Appveyor, can you take a look?

Waterman178:
thanks

qemu/translate-all.c
```diff
@@ -135,24 +135,55 @@ static void cpu_gen_init(struct uc_struct *uc)
...
+    int i, x;
+    void **p;
+
+    if (uc) {
+        if (uc->l1_map) {
+            x = V_L1_SHIFT / V_L2_BITS;
+            if (x <= 1) {
+                for (i = 0; i < V_L1_SIZE; i++) {
+                    p = uc->l1_map[i];
+                    g_free(p);
```
aquynh: do we need to check if p is NULL here? 
(at lest for consistency, because the code above does that)

chenhuitao: Not be necessary,  g_free() accepts NULL pointer. But you are right, it should be checked. I will add it.

aquynh:
merged, thanks!

PR #1149, opened 2019-10-06
fix oss-fuzz issue 10334

chenhuitao:
fix oss-fuzz issue 10334: unicorn/fuzz_emu_x86_32: Direct-leak in g_malloc.


Waterman178:
thanks

aquynh:
merged, thank!

PR #1151, opened 2019-10-09
Python packaging logic for msvc build

rhelmot:
Allows `python setup.py bdist_wheel` on windows to properly build and distribute its own binaries under the new configuration.


aquynh:
it seems this does not handle binaries put under prebuilt/ anymore?

rhelmot:
It should support the prebuilt stuff, I just removed the part where the
libraries we don’t need anymore are copied. Immediately below is the part
where it copies the unicorn dll and static library. Did it not work for you?

aquynh:
merged, thanks!

PR #1159, opened 2019-10-18
fix oss-fuzz issue 10578.

chenhuitao:
fix oss-fuzz issue 10578.
Issue 10578: unicorn/fuzz_emu_x86_32: Crash in helper_pcmpestrm_xmm.


aquynh:
merged, thanks!

PR #1161, opened 2019-10-20
Makefile Updated

jcalabres:
The Java binding needs some changes in order to work correctly with the new Java versions.

### DIFF
--- a/bindings/java/Makefile.build
+++ b/bindings/java/Makefile.build
@@ -3,7 +3,7 @@
 
 JAVA_HOME := $(shell jrunscript -e 'java.lang.System.out.println(java.lang.System.getProperty("java.home"));')
 
-JAVA_INC := $(shell realpath $(JAVA_HOME)/../include)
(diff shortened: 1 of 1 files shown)


aquynh:
does this affect the older version of Java?

@cseagle please ack.

jcalabres:
Hello, I was having problems with the makefile provided. As the new versions are not using javah anymore for generating the headers files, javac with -h it's needed. In addition, in order to use extensions in new Java versions, you need to specify the classpath. I'm not sure but, it's working for me using classpath in both, javac and java. 
So, yes, this is affecting old versions of Java, but without these kinds of changes, ppl cannot compile the Java Binding (no ghidra extensions :( )
Thank you!

aquynh:
i see. is there anyway to support older version too, so we dont sacrifice compatibility?

wtdcode:
Link to #1217 #1449.

wtdcode:
Merged in 3dfec280c765384c5536126d103c6781633f29a2.

PR #1165, opened 2019-10-25
Expose different 32-bit ARM CPU models to users via UC_MODE flags

### LINKED ISSUE #117: Extend uc_open() to specify CPU model
we need to let users to specify CPU model at the time the engine is initialized. this is highlighted with a recent issue https://github.com/unicorn-engine/unicorn/issues/114#issuecomment-138341814

one way to do this is to extend `uc_open()` to contain one more param for CPU model. something like:

```
uc_err uc_open(uc_arch arch, uc_mode mode, uc_cpu model, uc_engine **uc);
```

any comments?

hosaka-corp:
A project of mine requires QEMU's ARM926 backend. Currently, `UC_MODE_MCLASS` already sets the CPU type to `cortex-m3`, so I figured it's reasonable that we'd expose some of the other 32-bit ARM CPU types this way too. 

I figure that the `UC_MODE_` flags are a simple way to implement this without changes to the existing API (i.e. by having users pass a CPU model via `uc_open()` as discussed in #117).

This addresses/closes #117. In the future, new bits for other ARM CPU models can be easily specified in  `uc_mode`.


include/unicorn/unicorn.h
```diff
@@ -107,6 +107,11 @@ typedef enum uc_mode {
     UC_MODE_THUMB = 1 << 4,       // THUMB mode (including Thumb-2)
     UC_MODE_MCLASS = 1 << 5,      // ARM's Cortex-M series (currently unsupported)
     UC_MODE_V8 = 1 << 6,          // ARMv8 A32 encodings for ARM (currently unsupported)
+
+    UC_MODE_ARM926 = 1 << 7,
```
aquynh: can you add some comments behind each mode, like we have in other modes?

qemu/hw/arm/tosa.c
```diff
@@ -19,9 +19,25 @@
 
 static int tosa_init(struct uc_struct *uc, MachineState *machine)
 {
-    if (uc->mode & UC_MODE_MCLASS) {
+
+    if (uc->mode & UC_MODE_MCLASS)
```
aquynh: for if block, please put { at the end of the same line, but not in a new line.

aquynh:
cool, i will merge this, but please see my comments.

qemu/hw/arm/tosa.c
hosaka-corp: Would omitting braces for these one-line `if` cases be preferable, or do you specifically just want braces on the same line in all cases?

hosaka-corp: Thanks, see  da417e8

qemu/hw/arm/tosa.c
```diff
@@ -21,7 +21,17 @@ static int tosa_init(struct uc_struct *uc, MachineState *machine)
 {
     if (uc->mode & UC_MODE_MCLASS) {
         uc->cpu = (CPUState *)cpu_arm_init(uc, "cortex-m3");
-    } else {
+    }
```
aquynh: else and { are on the same line, too.

it is ok to omit { for one line block.

aquynh:
merged, thanks!

kitlith:
wish I had seen this earlier, but: when selecting a model of CPU don't you only want to select a single model of CPU? it doesn't really make sense to say "i want my cpu to be both ARM926 and ARM946", so wouldn't it make more sense to define a mask of bits that selects a model (which depends on the arch selected) where, say 0000 is "don't care/default", 0001 is "ARM926", 0010 is "ARM946", 0011 is "ARM1176", etc? I suppose that might not be consistent with the current MCLASS flag, but you could always deprecate that bit and encourage use of the new model selection thing.

aquynh:
if you look at the value of these modes, you can see that we can OR them
like: UC_MODE_ARM946 | UC_MODE_ARM926.
so there is no need to change the semantics of this API to choose more than
1 mode.

however, implementation does not support this yet. instead of checking the
input mode value, we should check if the input mode can be AND with
UC_MODE_ARM946, etc.

feel free to send in PR to do this, thanks.

PR #1166, opened 2019-10-25
X64 base regs

domenukk:
This commit adds X86_64 fs and gs_base registers to Unicorn.
I've adopted the regs from `./test` branch for this, but removed support from 32 and 16 bit.
The base registers are, to my knowledge, only "officially" available on 64 bit mode.
The whole thing is a mess [0]
Not sure how exactly this plays together with a (potential) emulated gdt, but it allows to reset state without hacky workarounds.

[0] https://elixir.bootlin.com/linux/v4.5/source/arch/x86/kernel/process_64.c#L332


aquynh:
merged, thanks!

PR #1168, opened 2019-10-31
Tests, fixes on third platform.

devnexen:
MT linkage fix mainly.


aquynh:
What is the problem this is trying to fix?

devnexen:
The build and linkage kind.

aquynh:
merged, thanks!

Issue #1100, opened 2019-07-01
pip install problems


Zardows:
I tried to install unicorn and I get that:

Collecting unicorn
  Downloading https://files.pythonhosted.org/packages/7d/7f/47fe864fe967e91de2d57677618cffc91bee3918f0a3cdbaa6500b36855e/unicorn-1.0.1.tar.gz (1.5MB)
     |████████████████████████████████| 1.5MB 502kB/s 
Installing collected packages: unicorn
  Running setup.py install for unicorn ... error
    ERROR: Complete output from command /Library/Frameworks/Python.framework/Versions/3.7/bin/python3.7 -u -c 'import setuptools, tokenize;__file__='"'"'/private/tmp/pip-install-3si_utlr/unicorn/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' install --record /private/tmp/pip-record-sr9r53ei/install-record.txt --single-version-externally-managed --compile:
    ERROR: running install
    running build
    Building C extensions
    cd qemu && \
    	./configure --cc="cc" --extra-cflags="-DUNICORN_HAS_X86 -DUNICORN_HAS_ARM -DUNICORN_HAS_ARMEB -DUNICORN_HAS_M68K -DUNICORN_HAS_ARM64 -DUNICORN_HAS_MIPS -DUNICORN_HAS_MIPSEL -DUNICORN_HAS_MIPS64 -DUNICORN_HAS_MIPS64EL -DUNICORN_HAS_SPARC -fPIC -fvisibility=hidden -m32 -arch i386 -m64 -arch x86_64" --target-list="x86_64-softmmu, arm-softmmu, armeb-softmmu, m68k-softmmu, aarch64-softmmu, mips-softmmu, mipsel-softmmu, mips64-softmmu, mips64el-softmmu, sparc-softmmu,sparc64-softmmu,"
    
    ERROR: Cannot use 'python', Python 2.4 or later is required.
           Note that Python 3 or later is not yet supported.
           Use --python=/path/to/python to specify a supported Python.
    
    make: *** [qemu/config-host.h-timestamp] Error 1
    error: [Errno 2] No such file or directory: 'libunicorn.dylib'
    ----------------------------------------
[...]

PR #1172, opened 2019-11-19
restrict msbuild to win32 platform

rhelmot:
`msbuild`-based build doesn't actually work outside of windows. msbuild on mac and linux only works for C# programs I'm pretty sure.

see #1100


aquynh:
merged, thanks!

@stephengroat please could you look at the error reported by Appveyor?

PR #1173, opened 2019-11-24
Timeout error

ZehMatt:
I was working on a project where I had to measure generated code and noticed that I had no way of telling if the code runs for infinite time so I had to implement it myself, this PR is a side effect of my needs and maybe it provides to be useful in the main code base. This PR implements a new error code UC_ERR_TIMEOUT, to do that I had to add a new state variable which is reset in uc_emu_start and set in the timeout thread function. I also adjusted the sample to reflect the new change.

### DIFF
--- a/include/uc_priv.h
+++ b/include/uc_priv.h
@@ -227,6 +227,7 @@ struct uc_struct {
     bool stop_request;  // request to immediately stop emulation - for uc_emu_stop()
     bool quit_request;  // request to quit the current TB, but continue to emulate - for uc_mem_protect()
     bool emulation_done;  // emulation is done by uc_emu_start()
+    bool timed_out;     // emulation timed out, uc_emu_start() will result in EC_ERR_TIMEOUT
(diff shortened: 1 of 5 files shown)


aquynh:
merged, thanks!

aquynh:
looking back at this, there is one thing i dont like: this breaks compatibility, since old code can assume that `uc_emu_start()` returns 0 even when timeout happens. so maybe we should revert this, and find another approach that reserves compatibility.

ZehMatt:
Could maybe add a function to get the state instead but this did seem like the right approach.

aquynh:
i reverted most of this pull req, to ensure that timeout is not considered error: https://github.com/unicorn-engine/unicorn/commit/fbef45b18f2f59d9699af54c6ad6ac73eaba3b29

now you need to query if engine stops due to timeout with `uc_query()`, using `UC_QUERY_TIMEOUT` type.

thanks.

PR #1179, opened 2019-12-31
Ensure that PC is not fixed up when code tracing or timing.

gerph:
Under some circumstances, the PC is not fixed up properly when
returning from the execution of a block in cpu_tb_exec. This appears
to be caused by the resetting of the PC from the tb.

This change removes the additional fixup in the cases where there
is code tracing or timing active. Either of these cases would result
in the wrong PC being reported.

Closes unicorn-engine#1105.


aquynh:
merged, thanks!

PR #1184, opened 2020-01-07
fix some oss-fuzz

chenhuitao:
fix some oss-fuzz: 10598, 11112, 11186, 11589, 10674, 10878, 11655, 19610, 19848, 19849, 19850, 19851, 19852.

### DIFF
--- a/qemu/fpu/softfloat.c
+++ b/qemu/fpu/softfloat.c
@@ -4606,7 +4606,7 @@ int32 floatx80_to_int32( floatx80 a STATUS_PARAM )
 
     if (floatx80_invalid_encoding(a)) {
         float_raise(float_flag_invalid STATUS_VAR);
-        return 1 << 31;
+        return (int32)(1U << 31);
     }
     aSig = extractFloatx80Frac( a );
     aExp = extractFloatx80Exp( a );
@@ -4638,7 +4638,7 @@ int32 floatx80_to_int32_round_to_zero( floatx80 a STATUS_PARAM )
 
     if (floatx80_invalid_encoding(a)) {
         float_raise(float_flag_invalid STATUS_VAR);
-        return 1 << 31;
+        return (int32)(1U << 31);
     }
     aSig = extractFloatx80Frac( a );
     aExp = extractFloatx80Exp( a );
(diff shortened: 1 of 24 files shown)


qemu/fpu/softfloat.c
```diff
@@ -130,7 +130,7 @@ static int32 roundAndPackInt32( flag zSign, uint64_t absZ STATUS_PARAM)
     absZ = ( absZ + roundIncrement )>>7;
     absZ &= ~ ( ( ( roundBits ^ 0x40 ) == 0 ) & roundNearestEven );
     z = (int32_t)absZ;
-    if ( zSign ) z = - z;
+    if ( zSign && (z != 0x80000000)) z = - z;
```
aquynh: where is this 0x80000000 from?

chenhuitao: 0x80000000 is MIN_INT32.

qemu/target-arm/translate.c
```diff
@@ -807,7 +807,9 @@ void arm_gen_test_cc(TCGContext *tcg_ctx, int cc, int label)
         break;
     default:
         fprintf(stderr, "Bad condition code 0x%x\n", cc);
```
aquynh: pls comment out this fprintf line

qemu/target-i386/int_helper.c
```diff
@@ -352,7 +352,7 @@ static int idiv64(uint64_t *plow, uint64_t *phigh, int64_t b)
         neg128(plow, phigh);
     }
     sb = (b < 0);
-    if (sb) {
+    if (sb && (b != 0x8000000000000000LL)) {
```
aquynh: this looks like a trick here. sorry but i feel it is ugly to fix a bug this way. any ideas why oss-fuzzer reports this issue?

qemu/target-arm/translate.c
chenhuitao: O.K.

qemu/target-i386/int_helper.c
chenhuitao: The min int8_t is -128(0x80), the max is 127(0x7f), the abs value of min is bigger than max and overflowed. In most machine and compiler, -(-128) = ~(-128) + 1 = -128. Same ruler for other int types. I think the oss-fuzz just report this bug is not good compatible, just like left shift the negative int. This code change keeps the original logic and becomes compatible.

aquynh:
merged, thanks!

PR #1186, opened 2020-01-09
Migration for Travis-CI

chfl4gs:
make test disabled for msys2/MinGW for now.

### DIFF
--- a/qemu/target-arm/neon_helper.c
+++ b/qemu/target-arm/neon_helper.c
@@ -608,7 +608,7 @@ NEON_POP(pmax_u16, neon_u16, 2)
 #undef NEON_FN
 
 #define NEON_FN(dest, src1, src2) \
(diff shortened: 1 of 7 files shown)


aquynh:
this makes Travis support Windows, so we may remove Appveyor CI later. @stephengroat do you have any comments?

stephengroat:
this seems to remove msys2 support

is that the idea?

chfl4gs:
msys2 support is still there. However make test is not implemented as there is some issue with shared library loading.

chfl4gs:
Test run results are here.
https://travis-ci.com/chfl4gs/unicorn/builds/143668320

chfl4gs:
make test produces error as shown in https://travis-ci.com/chfl4gs/unicorn/builds/143214405#L1138

aquynh:
nice, will we have msys support soon?

PR #1189, opened 2020-01-14
fix some oss-fuzz

chenhuitao:
fix some oss-fuzz bugs: 20069, 20071, 20073, 20075, 20076, 20078, 20080.


aquynh:
merged, thanks!

PR #1192, opened 2020-01-17
fix some oss-fuzz

chenhuitao:
fix some oss-fuzz bug: 20166, 14042.


qemu/target-mips/helper.c
```diff
@@ -341,10 +341,15 @@ int mips_cpu_handle_mmu_fault(CPUState *cs, vaddr address, int rw,
              " prot %d\n",
              __func__, address, ret, physical, prot);
     if (ret == TLBRET_MATCH) {
+      if (mmu_idx < 0 || mmu_idx >= NB_MMU_MODES) {
+        raise_mmu_exception(env, address, rw, ret);
```
aquynh: this part is tricky. do you look at the latest Qemu (4.2) to see if they changed this code?

chenhuitao: I don't look at qemu-4.2. When the mmu_idx array index is out of bounds, it overwrite other members of env struct. So I add bound check.

aquynh:
merged, thanks!

Issue #1193, opened 2020-01-18
Block hooks called twice immediately following an emulation stop


gerph:
I'm seeing problem with the block hook being called twice (with no intervening code execution) when the emulation was stopped in the previous emulation cycle. That seems rather difficult to explain, so I'll give a sequence that causes it:

1. Register a block hook.
2. Register an exception hook (UC_HOOK_INTR)
3. Start execution of code (emu_start)
4. Our block hook is called to say that we're executing our code
5. When an exception instruction is executed, the hook issues an emu_stop to request that the system return.
6. We return from emu_start, and do whatever we need to do to handle the exception.
7. Start execution of code again.
8. Our block hook is called to say that we execute the instructions after the exception instruction.
9. Our block hook is called AGAIN, with the same block.
10. Execution continues.

That's the sequence that happens in the failure case, and if we insert hooks for code, we see that between the block hook being called at step 8 and step 9, no code is executed (or at least, no code hook is executed, and by constructing the code in known ways, it is clear that the code is not executed.

Actual example code that exhibits this issue:

```python
# Demonstrate the repeated block hooks being called.

from capstone import *
import capstone.arm_const

from unicorn import *
import unicorn.arm_const as arm

# The bit pattern that is used here is intended to expose execution issues:
#  * If an instruction is executed twice, there's a gap in the bit pattern that
#    will be filled (ie odd bits should never be set in normal execution)
#  * If a block is executed twice, multiple bits will be affected.
armcode = """
    8000 : .... : E3A00902 : MOV     r0,#&8000
    8004 : .... : E2800001 : ADD     r0,r0,#&1 ; = 1
    8008 : .... : E2800004 : ADD     r0,r0,#&4 ; = 4
    800C : .... : EF000001 : SWI     1
    8010 : .... : E2800010 : ADD     r0,r0,#&10 ; = 16
    8014 : @... : E2800040 : ADD     r0,r0,#&40 ; = 64
    8018 : .... : EF000002 : SWI     2
    801C : .... : E2800C01 : ADD     r0,r0,#&100 ; = 256
    8020 : .... : E3A0F000 : MOV     pc,#&0
"""

md = Cs(CS_ARCH_ARM, CS_MODE_ARM)
md.detail = True
md.syntax = CS_OPT_SYNTAX_NOREGNAME

code = []
lines = armcode.splitlines()
for line in lines:
    if line:
        # Change the little-endian string to a set of characters
        inst = line[18:26]
[...]

PR #1194, opened 2020-01-18
Fix block hooks being called twice after an early exit from execution.

gerph:
Ensure the TCG exit flag is cleared at the end of cpu_exec. This ensures
that subsequent calls are not polluted by the prior call to request an
early exit. The symptoms of the problem being addressed here are that
after a cpu_exit call triggered within a hook there may still be a
pending tcg_exit_req flag set. This then causes a block to start its
execution and then be aborted (from which point it'll continue because
there's no other condition to service). The start of the execution
causes the block hook to be called, no actual code to be run, and then
the block started again with another block hook call.

This change is discussed in ticket 1193:
    https://github.com/unicorn-engine/unicorn/issues/1193

### DIFF
--- a/qemu/cpu-exec.c
+++ b/qemu/cpu-exec.c
@@ -291,6 +291,9 @@ int cpu_exec(struct uc_struct *uc, CPUArchState *env)   // qq
         }
     } /* for(;;) */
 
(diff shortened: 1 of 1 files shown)


aquynh:
merged, thanks!

PR #1195, opened 2020-01-22
Make it easier to link against libunicorn on macOS

saagarjha:
libunicorn.dylib is currently installed on macOS as unversioned but its install name is versioned and lacking an @rpath, making it difficult to use from clients without modifying it with `install_name_tool`. This change makes it easier to link against because now binaries that want to do so can adapt *their* `LC_RPATH` to match Unicorn's install location and the actual library does not have to be modified. See https://github.com/angr/angr/pull/1932 for how this is being used.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -112,7 +112,7 @@ API_MAJOR=$(shell echo `grep -e UC_API_MAJOR include/unicorn/unicorn.h | grep -v
 ifeq ($(UNAME_S),Darwin)
 EXT = dylib
 VERSION_EXT = $(API_MAJOR).$(EXT)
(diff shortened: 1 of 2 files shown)


aquynh:
@rhelmot please ack.

saagarjha:
The build failure also looks relevant; what platform is that exactly?

aquynh:
do you mean those MingW build on Appveyor? you can ignore that.

rhelmot:
comments from the angr PR also apply here

also why are we producing files which match libunicorn*.dylib but aren't libunicorn.dylib?

rhelmot:
okay I understand why there's libunicorn.1.dylib (hand-waving make.sh) but is it possible that we could _just_ have there be libunicorn.dylib? There aren't really versioning semantics in this project which make that sort of filename-level designation make sense

saagarjha:
> okay I understand why there's libunicorn.1.dylib (hand-waving make.sh)

CMake does this, I believe, if you set `SOVERSION`: https://github.com/unicorn-engine/unicorn/blob/60896de9f4418a32c6e4dde94b0eb9d4f8fa1b3b/CMakeLists.txt#L849

aquynh:
merged, thanks!

PR #1201, opened 2020-02-08
Python 3.8 compatibility

HarukaMa:
Python 3.8 moved some types from `collections` to `collections.abc`.


aquynh:
Cool, how to reproduce this issue?

HarukaMa:
Just `./make.sh` when the default python in the system is python 3.8 should be enough I think.

Actually, I was using a version from master which is called 3.9 at the moment, but the document says 3.8 is when it changed.

aquynh:
merged, thanks!

PR #1204, opened 2020-02-10
Expose uc_vmem_{read,write} to users

hosaka-corp:
Uses `cpu_memory_rw_debug()` to resolve a virtual address and perform some read/write. 
I've only tested this on 32-bit ARM - don't know if this particular path works for other guest architectures.

### DIFF
--- a/include/uc_priv.h
+++ b/include/uc_priv.h
@@ -54,6 +54,9 @@ typedef bool (*uc_write_mem_t)(AddressSpace *as, hwaddr addr, const uint8_t *buf
 
 typedef bool (*uc_read_mem_t)(AddressSpace *as, hwaddr addr, uint8_t *buf, int len);
 
+typedef bool (*uc_read_vmem_t)(CPUState *cpu, hwaddr addr, uint8_t *buf, int len);
+typedef bool (*uc_write_vmem_t)(CPUState *cpu, hwaddr addr, const uint8_t *buf, int len);
+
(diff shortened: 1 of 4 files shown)


aquynh:
Why do you need this?

uc_mem_xxx are typically used for virtual address access, so what is your scenario?

hosaka-corp:
I am hosting a bunch of not-entirely-userspace code (a kernel) that configures the ARM MMU, loads other binaries, etc.

Things behave fine until the MMU is configured. Afterwards, when using `uc_mem_read`,  I get `UC_ERR_READ_UNMAPPED` on virtual addresses that I have _not_ mapped in (because I cannot know all the valid mappings a priori at any given instant). 

So, I've been using `uc_mem_map_ptr` to map memory to _physical addresses_ and relying on this commit when I need to resolve virtual addresses in my hooks and such. It seems to work fine.

Although, this confuses me because it indicates that Unicorn's abstraction for memory regions are slightly ambiguous: they do not _easily_ behave like virtual memory, and seem like they are actually more like physical memory (because they always require some backing on the host and cannot overlap each other). 

If I wanted to use `uc_mem_map_ptr` to implement this, wouldn't I would need to have knowledge of all the mappings all the time? Am I supposed to emulate the MMU outside of the actual CPU in QEMU by managing all of the regions with `uc_mem_map_ptr` by myself or something? Unless maybe I'm doing something else fundamentally wrong here?

Issue #1203, opened 2020-02-09
cross build unicorn use NDK, helper_* function error


SmailCatLog:
https://github.com/unicorn-engine/unicorn/issues/1106#issue-473980982

https://github.com/unicorn-engine/unicorn/blob/63ebcf41a838bb9e259ba309f2025b1da819e323/qemu/tcg/tcg.c#L2378-L2381

```c++
#if TCG_TARGET_REG_BITS == 32 && !defined(__ARM_ARCH)
    // do this because msvc cannot have arrays with 0 entries. 
    nb_regs = 0; 
#endif

hexcoder-:
s/define/defined/

hexcoder-:
The fix tested ok with Raspberry Pi 4 (on 32-bit raspbian). Please commit! Thanks.

aquynh:
Please pull req.

PR #1205, opened 2020-02-10
fixes #1183 as per #1203

### LINKED ISSUE #1106: calling convention
I try to cross build unicorn use NDK, but i found an error when i run samples.
The  function 'helper_uc_tracecode' gets arguments from r0 ~ r3, but actually, the caller pass arguments by stack... how can i solve this problem?

I try to add  inline asm code to solve this problem(file: uc.c)

```c
// TCG helper
void __attribute__((naked)) helper_uc_tracecode(int32_t size, uc_hook_type type, void *handle, int64_t address);
void __attribute__((naked)) helper_uc_tracecode(int32_t size, uc_hook_type type, void *handle, int64_t address){
        asm(
    "LDR R0,[SP] \n\t"       
    "LDR R1,[SP,4] \n\t"
    "LDR R2,[SP,8] \n\t"
    "LDR R3,[SP,0xC] \n\t"
    "b helper_uc_tracecode_real"
    );

}
void helper_uc_tracecode_real(int32_t size, uc_hook_type type, void *handle, int64_t address)
{
.....
}
```

then helper_uc_tracecode runs correctly.however, i found many functions which name starts with ‘helper_’，such as 'helper_wfi_arm' had this problem.

all of their caller  are out of area of .text, maybe are dynamic gen.

### LINKED ISSUE #1183: Segmentation Fault emulating MIPS on ARM
When running [the simple test harness of afl unicorn](https://github.com/Battelle/afl-unicorn/blob/master/unicorn_mode/samples/simple/simple_test_harness.py)  (without any afl patches!) using Unicorn 1.0.2rc1 on the Raspberri PI, the following line crashes with segmentation fault:
https://github.com/unicorn-engine/unicorn/blob/24f55a7973278f20f0de21b904851d99d4716263/qemu/target-mips/op_helper.c#L2277

The stack is completely bogus when it crashes. No clue what's going on, sadly. It works fine on x86...

Command line:
```bash
➜  simple git:(master) python simple_test_harness.py sample_inputs/sample1.bin
Loading data input from sample_inputs/sample1.bin
Starting the AFL forkserver by executing 1 instruction
[1]    20476 segmentation fault  python simple_test_harness.py sample_inputs/sample1.bin
```

domenukk:
This should fix #1183 on ARM and possibly other archs, as per #1203

### DIFF
--- a/qemu/tcg/tcg.c
+++ b/qemu/tcg/tcg.c
@@ -2378,7 +2378,7 @@ static int tcg_reg_alloc_call(TCGContext *s, const TCGOpDef *def,
     flags = args[nb_oargs + nb_iargs + 1];
 
     nb_regs = ARRAY_SIZE(tcg_target_call_iarg_regs);
-#if TCG_TARGET_REG_BITS == 32
+#if TCG_TARGET_REG_BITS == 32 && defined(_M_IX86)
     // do this because msvc cannot have arrays with 0 entries.
     nb_regs = 0;
 #endif
--- a/qemu/tcg/tcg.c
+++ b/qemu/tcg/tcg.c
@@ -2378,7 +2378,7 @@ static int tcg_reg_alloc_call(TCGContext *s, const TCGOpDef *def,
     flags = args[nb_oargs + nb_iargs + 1];
 
     nb_regs = ARRAY_SIZE(tcg_target_call_iarg_regs);
-#if TCG_TARGET_REG_BITS == 32 && defined(_M_IX86)
+#if TCG_TARGET_REG_BITS == 32 && defined(__ARM_ARCH)
     // do this because msvc cannot have arrays with 0 entries.
     nb_regs = 0;
 #endif


domenukk:
Now it works :) Could still have the same issue on mips or risc-v in case that is used somewhere..

qemu/tcg/tcg.c
```diff
@@ -2375,7 +2375,7 @@ static int tcg_reg_alloc_call(TCGContext *s, const TCGOpDef *def,
     flags = args[nb_oargs + nb_iargs + 1];
 
     nb_regs = ARRAY_SIZE(tcg_target_call_iarg_regs);
-#if TCG_TARGET_REG_BITS == 32
+#if TCG_TARGET_REG_BITS == 32 && defined(__ARM_ARCH)
```
aquynh: can you explain why this is only for ARM?

domenukk: I didn't look into the bug, but doing it for all None-VS x86 as in  https://github.com/unicorn-engine/unicorn/pull/1205/commits/15d941e847f2cdf714209bc4238083cf6d0bc1b5 the x86 test-cases failed - so instead I played it safe.
Maybe @SmailCatLog can give some insights?

SmailCatLog: Because I don't know the macro definition of x86

SmailCatLog: Because I don't know the macro definition of x86

aquynh:
That check for ARM makes little sense. Can you improve it?

qemu/tcg/tcg.c
```diff
@@ -2375,7 +2375,7 @@ static int tcg_reg_alloc_call(TCGContext *s, const TCGOpDef *def,
     flags = args[nb_oargs + nb_iargs + 1];
 
     nb_regs = ARRAY_SIZE(tcg_target_call_iarg_regs);
-#if TCG_TARGET_REG_BITS == 32
+#if TCG_TARGET_REG_BITS == 32 && defined(_M_IX86)
```
domenukk: @aquynh How about this? If the comment is correct, only microsoft compilers should be affected, so this might be an okay fix?

aquynh: i still dont get it: can you explain why this only affects X86 32bit?

domenukk: That's the commit in question:
https://github.com/unicorn-engine/unicorn/commit/002151874a78c991f41fbe9d16a9f4d6e97ee959
The comment by @xorstream states `// do this because msvc cannot have arrays with 0 entries.`
I didn't look any deeper than that, sorry.

qemu/tcg/tcg.c
aquynh: any follow up?

aquynh:
any follow up on this PR?

domenukk:
In the meantime, we went back to `&& defined(__ARM_ARCH)` as it still segfaulted for 32 bit otherwise (not only for msvc). I still didn't look into the issue further and don't claim to understand what's going on here :)

domenukk:
Rebased on the current master. 
@SmailCatLog so this may still break on non-x86 like MIPS, is it?

SmailCatLog:
It looks like any 32-bit reg pass parameters will have problems.

SmailCatLog:
the TCG_TARGET_REG_BITS macro is for https://github.com/unicorn-engine/unicorn/blob/master/qemu/tcg/i386/tcg-target.c#L66

wtdcode:
Closed due to outdated.

PR #1207, opened 2020-02-13
The expression(args[i] < 0) always returns false

Waterman178:
`typedef uint64_t tcg_target_ulong;
typedef tcg_target_ulong TCGArg;
TCGArg *args`

`} else { do_reset_output: for (i = 0; i < nb_oargs; i++) { if (args[i] < 0 || args[i] >= TCG_MAX_TEMPS) {`

`if (!(def->flags & TCG_OPF_SIDE_EFFECTS) && nb_oargs != 0) { for(i = 0; i < nb_oargs; i++) { for(i = 0; i < nb_oargs; i++) { if (args[i] < 0 || args[i] >= TCG_MAX_TEMPS) {`

The expression(args[i] < 0) always returns false.So i deleted it

### DIFF
--- a/qemu/tcg/optimize.c
+++ b/qemu/tcg/optimize.c
@@ -1370,7 +1370,7 @@ static TCGArg *tcg_constant_folding(TCGContext *s, uint16_t *tcg_opc_ptr,
             } else {
         do_reset_output:
                 for (i = 0; i < nb_oargs; i++) {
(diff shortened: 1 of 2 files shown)


aquynh:
cc @chenhuitao

chenhuitao:
Yes, you are right. TCGArg is unsigned, delete them.

aquynh:
merged, thanks!

Issue #1214, opened 2020-02-27
macro/struct redefinition in msvc


HyperSine:
Hi,
I installed unicorn via `vcpkg install unicorn:x64-windows`. After creating a project in VS2019, with following simple code:

```cpp
#include <tchar.h>
#include <stdio.h>
#include <windows.h>
#include <unicorn/unicorn.h>
#include <stdexcept>

int _tmain(int argc, PTSTR argv[]) {
    uc_err err = UC_ERR_OK;
    uc_engine* vm_i386 = nullptr;

    err = uc_open(UC_ARCH_X86, UC_MODE_32, &vm_i386);
    if (err != UC_ERR_OK) {
        throw std::runtime_error(uc_strerror(err));
    }

    err = uc_close(vm_i386);
    if (err != UC_ERR_OK) {
        throw std::runtime_error(uc_strerror(err));
    }

    return 0;
}
```

I got a lot of redefinition warnings/errors:
```
1>------ Build started: Project: UnicornTest0, Configuration: Debug x64 ------
1>_tmain.cpp
1>C:\Program Files (x86)\Windows Kits\10\Include\10.0.18362.0\shared\ws2def.h(103,1): warning C4005: 'AF_IPX': macro redefinition
1>C:\Program Files (x86)\Windows Kits\10\Include\10.0.18362.0\um\winsock.h(457): message : see previous definition of 'AF_IPX'
1>C:\Program Files (x86)\Windows Kits\10\Include\10.0.18362.0\shared\ws2def.h(147,1): warning C4005: 'AF_MAX': macro redefinition
1>C:\Program Files (x86)\Windows Kits\10\Include\10.0.18362.0\um\winsock.h(476): message : see previous definition of 'AF_MAX'
1>C:\Program Files (x86)\Windows Kits\10\Include\10.0.18362.0\shared\ws2def.h(185,1): warning C4005: 'SO_DONTLINGER': macro redefinition
1>C:\Program Files (x86)\Windows Kits\10\Include\10.0.18362.0\um\winsock.h(399): message : see previous definition of 'SO_DONTLINGER'
1>C:\Program Files (x86)\Windows Kits\10\Include\10.0.18362.0\shared\ws2def.h(238,25): error C2011: 'sockaddr': 'struct' type redefinition
1>C:\Program Files (x86)\Windows Kits\10\Include\10.0.18362.0\um\winsock.h(1007): message : see declaration of 'sockaddr'
[...]

PR #1215, opened 2020-02-27
use UNICORN_INTERNAL to separate internal code

HyperSine:
It should fix issue #1214.

`<winsock2.h>` is required by the definition in `<unicorn/platform.h>` only:
```
static int gettimeofday(struct timeval* t, void* timezone) {    // `struct timeval` is defined in winsock2.h
    ...
}
```

This function is used by unicorn internally only and is not a part of unicorn apis. So such code should be placed in a header file which is used internally only. Or, at least, it should not be visible to users of unicorn. However, I cannot find a proper header file where the code is moved. And creating a new internal header file would be very troublesome for me.

So I introduce a new macro `UNICORN_INTERNAL` which makes the code I mentioned visible only when the macro is defined. The new macro is added to preprocessor definitions of all VS projects that are a part of unicorn, so that the code I mentioned works as always for internal and users of unicorn won't have `<winsock2.h>` included.

Of course, so does `static int usleep(uint32_t usec) {...}`.

### DIFF
--- a/pkgconfig.mk
+++ b/pkgconfig.mk
@@ -10,4 +10,4 @@ PKG_MINOR = 0
 PKG_EXTRA = 2
 
 # version tag. Examples: rc1, b2, post1
-PKG_TAG = rc2
+PKG_TAG = rc3


aquynh:
it is better to move this intern, but do not use this macro in the API. can you fix this?

PR #1220, opened 2020-03-03
fix finding python path which only has python3

chenhuitao:
fix finding python path which only has python3.


aquynh:
merged, thanks!

Issue #635, opened 2016-09-11
get_clock() is not monotonic (on non _WIN32)


JayFoxRox:
`get_clock` (include/qemu/timer.h) is  used in the timeout calculation (and probably in various other places too?). However, that timer is _not_ monotonic and might jump between loop iterations.
This probably makes it possible to jump back in time after `uc_emu_start` which would mean the timeout won't trigger as expected.

This is partially an upstream (documentation) issue. Unicorn trips on it.

---

Here is a rundown of the involved components:

**[gettimeofday manpage](http://man7.org/linux/man-pages/man2/gettimeofday.2.html)**

"The time returned by gettimeofday() is affected by discontinuous jumps in the system time (e.g., if the system administrator manually        changes the system time). If you need a monotonically increasing clock, see clock_gettime(2)."

**[Unicorn timer.h (get_clock_realtime)](https://github.com/unicorn-engine/unicorn/blob/master/qemu/include/qemu/timer.h#L631:)**

The comment reads `/* real time host monotonic timer */` but the function internally uses the non-monotonic `gettimeofday`.

([This documentation errror also exists upstream!](https://github.com/qemu/qemu/blob/master/include/qemu/timer.h#L800))

**[Unicorn timer.h (get_clock)](https://github.com/unicorn-engine/unicorn/blob/master/qemu/include/qemu/timer.h#L657)**

This calls `get_clock_realtime` internally which, as I've just shown, is _not_ monotonic.
So if the system time changes, the timeout calculation will fail.

(This issue does _not exist_ upstream)

**[QEMU timer.h (get_clock)](https://github.com/qemu/qemu/blob/master/include/qemu/timer.h#L826)**

Note that they are aware of the previously mentioned monotonic-issue: `/* XXX: using gettimeofday leads to problems if the date changes, so it should be avoided. */`
They have a precompiler check for `CLOCK_MONOTONIC` which resolves the issue.

---

This issue only affects posix. Windows already uses the monotonic performance counters.

aquynh:
this is interesting! can you come up with a fix?

JayFoxRox:
Why was this closed?

wtdcode:
> Why was this closed?

Thanks for your interest. You could have a try on our new version: Unicorn2 on dev branch. If the problem still exists, please re-open this issue.

wtdcode:
> Why was this closed?

Confirmed. Re-open.

wtdcode:
Manually fixed in 81942e9

Issue #1217, opened 2020-02-28
Unicorn 2 - looking for sponsors


aquynh:
Since its born in 2015, Unicorn has been inspiring so many cool research & tools, that we believe it deserves serious upgrade. Based on Qemu 2.1.2, there is a lot to catch up with the latest version of Qemu (Qemu is at v4.2 at the moment). Unfortunately, the code base of Qemu changes so much since 2015, that we believe it is faster to rewrite Unicorn from scratch.

For this reason, next major update of Unicorn (lets call it **Unicorn 2**) requires huge effort & lots of time. So we call for the support from Unicorn users!

If you are using Unicorn for your work, and **wish to sponsor the development of Unicorn 2**, please **contact us**.

korniltsev:
are you going to keep in sync with upsream after rewrite?

aquynh:
That is not easy to keep syncing with the upstream. This means we must
trade off with some sacrifices in features.

Keep posting here if you have any ideas.

Waterman178:
Support for the next version!

XVilka:
Related https://github.com/unicorn-engine/unicorn/issues/1190

minexew:
@ XVilka Due to how Unicorn is created (ripping out the heart of QEMU and fixing it up to make it nicely embeddable), it unfortunately seems that any new branch of Unicorn is doomed to ultimately fall out of sync with QEMU upstream again.

GPUCode:
Is there any update on Unicorn 2?

kzsnow:
*My two cents*: I did a bit of "Unicorn 2" prototyping with Qemu 4.2. The two biggests "lifts" I encountered, in terms of source code modifications, were compartmentalizing qemu globals (e.g. when multiple emu instances are used), and dealing with the memory implementation.

Like others, I'm cautiously optimistic there is a way to minimize Qemu source code mods to make it easier to keep up with development there, and greatly lessen the burden for updates.

A quick hack to get around the issue of qemu globals without source code mods is to load unique instances of the shared library for each "emu" instance. Of course the trade-off is increased memory consumption (however big the qemu lib is, for each instance). I've used Unicorn quite a bit, and in all my use cases (personally) this would be an acceptable trade-off.

[...]

PR #1232, opened 2020-04-12
fix macro/struct redefinition from winsock2.h

HyperSine:
A new fix for issue #1214. No new macro is introduced this time. 

I have to say it is hard to decide which internal header file `gettimeofday()` moves to. But I found `gettimeofday()` is called at only one place, so I choose to inline it manually. 

BTW, I also fixed a very old issue #635. Just merge code from upstream.

### DIFF
--- a/include/unicorn/platform.h
+++ b/include/unicorn/platform.h
@@ -155,17 +155,7 @@ typedef _W64 unsigned int  uintptr_t;
 // sys/time.h compatibility
 #if defined(_MSC_VER)
 #include <sys/timeb.h>
-#include <winsock2.h>
 #include <windows.h>
-
-static int gettimeofday(struct timeval* t, void* timezone)
-{
-    struct _timeb timebuffer;
-    _ftime( &timebuffer );
-    t->tv_sec = (long)timebuffer.time;
-    t->tv_usec = 1000*timebuffer.millitm;
-    return 0;
-}
 #else
 #include <sys/time.h>
 #endif
--- a/qemu/include/qemu/timer.h
+++ b/qemu/include/qemu/timer.h
@@ -158,6 +158,30 @@ static inline int64_t get_ticks_per_sec(void)
  */
 
 /* real time host monotonic timer */
+#ifdef _WIN32
+static inline int64_t get_clock_realtime(void)
+{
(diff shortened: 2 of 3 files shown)


chenhuitao:
@HyperSine I think that keeping the original getttimeofday() is better, _ftime() works. Just removing the line which includes winsock2.h.

HyperSine:
@chenhuitao 
No, it won't fix any problems. `gentimeofday()` requires `struct timeval` which is [defined in `<winsock.h>`](https://docs.microsoft.com/en-us/windows/win32/api/winsock/ns-winsock-timeval). 

So, if you want to keep the original `gettimeofday()` you must include `<winsock.h>` or `<winsock2.h>` in `<unicorn/platform.h>`. You can't remove `#include <winsock2.h>`. Otherwise, it would break.

chenhuitao:
@HyperSine Thanks for your reply. I think that `<windows.h>` includes the `<winsock.h>` or `<winsock2.h>`, so that we do not need the `<winsock2.h>` in `<unicorn/unicorn.h>`, just remove it to avoid the [winsock2.h-windows.h-conflict problem](https://stackoverflow.com/questions/1372480/c-redefinition-header-files-winsock2-h). If someone includes `<winsock.h>` or `<winsock2.h>`  in his code, he should resolve the conflict by himself.

HyperSine:
@chenhuitao 
Yes, `<windows.h>` includes `<winsock.h>` except `<winsock2.h>`. However, `msvc/unicorn/*/*.vcxproj` have `WIN32_LEAN_AND_MEAN` defined. So `<winsock.h>` is stripped from `<windows.h>`. If you remove `#include <winsock2.h>`, neither `<winsock.h>` nor `<winsock2.h>` will be included and unicorn would have compile failure.

chenhuitao:
@HyperSine 
You are right. I missed the macro `WIN32_LEAN_AND_MEAN`.
This [winsock2.h-windows.h-conflict problem](https://stackoverflow.com/questions/1372480/c-redefinition-header-files-winsock2-h) is MS's, not unicorn's.
BTW, users add this macro or include the `<winsock2.h>` before the `<windows.h>` in their code is another solution.

wtdcode:
Interesting. Waiting for investigation. #1217

wtdcode:
Manually fixed in 81942e9

PR #1243, opened 2020-04-27
move some operations after hook callback

chenhuitao:
move some operations after hook callback to before. so that user callback control the register before next execution.


aquynh:
merged, thanks!

heapcrash:
It appears that this broke some things for Pwntools, specifically we use Unicorn emulation to find PLT addresses in stripped ELFs.  1.0.2rc3 works fine, rc4 is broken.

aquynh:
Please provide a (minimal) sample code triggering this issue.

NyaMisty:
Well maybe it's this one:
https://github.com/Gallopsled/pwntools/issues/1538#issuecomment-648875882

heapcrash:
Here's a minimal test case from the linked Pwntools issue:

## Example Script to Demonstrate Issue

```py
#!/usr/bin/env python3
from pwn import *

# Turn on MAX debug logging, so that we can see PLT exploration
context.log_level = 1

# Load an ELF from the Pwntools test corpus of ELF files, a BE MIPS file which is not PIE
# The PLT is discovered dynamically via Unicorn emulation, so e.plt should get filled out here
e = ELF(pwnlib.data.elf.get('test-mips-big'))
```

## Working with 1.0.2.rc3

The script will print something like:

```
[DEBUG] PLT 0x400bb0 printf
[DEBUG] PLT 0x400bc0 __stack_chk_fail
[DEBUG] PLT 0x400bd0 __libc_start_main
[DEBUG] PLT 0x400be0 __isoc99_scanf
[DEBUG] PLT 0x400bf0 atoi
[*] '/Users/heapcrash/pwntools/pwnlib/data/elf/test-mips-big'
    Arch:     mips-32-big
    RELRO:    Partial RELRO
    Stack:    Canary found
    NX:       NX disabled
    PIE:      No PIE (0x400000)
    RWX:      Has RWX segments
```

## Broken with 1.0.2rc6

The script does not print any PLT debug information, just the ELF info.  This indicates we were unable to emulate instructions in the PLT to find out which GOT entries they correspond to.

```
[*] '/Users/heapcrash/pwntools/pwnlib/data/elf/test-mips-big'
    Arch:     mips-32-big
    RELRO:    Partial RELRO
    Stack:    Canary found
    NX:       NX disabled
    PIE:      No PIE (0x400000)
    RWX:      Has RWX segments
```

PR #1245, opened 2020-04-27
fix m68k func name

chenhuitao:
fix m68k func name. raise_exception() is a static func in op_helper.c, global func is helper_raise_exception().


aquynh:
merged, thanks!

Issue #980, opened 2018-07-19
wrong pc value reported in INTR hook for arm64


bx:
I have found that for arm64, reg_read reports incorrect (/unexpected) PC values during execution of a UC_HOOK_INTR hook. (For the most recent commit of the master branch, cf0bcb82.)
What I expect, and what appears to be the case in other architectures (such as MIPS and x86_64), is that the reported PC refers to the address of the currently executing svc/syscall instruction. 
I have found that _most_ (but not _all_) of the time on arm64, the PC is set to the address of the instruction following the svc (occasionally the reported PC does not appear to be anywhere near an svc instruction, unfortunately I don't have a minimal example of this).  

The following code highlights this issue
```
def intr_hook(uc, intno, expected):
    pc = uc.reg_read(UC_ARM64_REG_PC)
    if not pc == expected:
        print "Interrupt expected at %x, but pc is %x" % (expected, pc)
    else:
        print "Interrupt at %x ok" % (pc)

addr = 0x40000
code = '010000d4'.decode('hex')  #  svc     #0x0

uc = Uc(UC_ARCH_ARM64, UC_MODE_ARM)
uc.mem_map(addr, 0x1000, UC_PROT_ALL)
uc.mem_write(addr, code)
uc.hook_add(UC_HOOK_INTR, intr_hook, user_data=addr)
uc.emu_start(addr, addr+len(code))
```

When the above is executed, it prints the following:
> Interrupt expected at 40000, but pc is 40004

Issue #972, opened 2018-07-02
ARM - Wrong PC in data hook


abiondo:
In ARM, inside a memory hook, the PC is not the instruction that caused the memory access, but the start of the basic block containing that instruction.
This is probably coming from QEMU's basic block translation.
Other architectures, such as x86, are fine.

Demo code:
```python
from __future__ import print_function
from unicorn import *
from unicorn.arm_const import *

# +0: nop
# loop:
# +2: ldr r0, =0x20001000
# +4: ldr r1, [r0]
# +6: b loop
CODE = b'\x00\xbf\x01\x48\x01\x68\xfc\xe7\x00\x10\x00\x20'
ADDRESS = 0x20000000

# Alternates between True and False so that the hook
# does not hit when continuing
trigger = True


def hook_memory(uc, access, address, size, value, user_data):
    global trigger
    if trigger:
        pc = uc.reg_read(UC_ARM_REG_PC)
        print('    Inside hook, addr = 0x{:x}, pc = 0x{:x}'.format(address, pc))
        uc.emu_stop()
    trigger = not trigger


def smoke_test():
    uc = Uc(UC_ARCH_ARM, UC_MODE_THUMB)
    uc.mem_map(ADDRESS, 2 * 1024 * 1024)
    uc.mem_write(ADDRESS, CODE)
    uc.hook_add(UC_HOOK_MEM_READ, hook_memory, begin=ADDRESS+0x1000, end=ADDRESS+0x1004)

    uc.reg_write(UC_ARM_REG_PC, ADDRESS)

    for _ in range(5):
        pc = uc.reg_read(UC_ARM_REG_PC)
        print('>>> Before start, pc = 0x{:x}'.format(pc))
        uc.emu_start(pc | 1, ADDRESS + len(CODE))


if __name__ == '__main__':
    smoke_test()
```

Output:
```
>>> Before start, pc = 0x20000000
    Inside hook, addr = 0x20001000, pc = 0x20000000
>>> Before start, pc = 0x20000000
    Inside hook, addr = 0x20001000, pc = 0x20000002
>>> Before start, pc = 0x20000002
    Inside hook, addr = 0x20001000, pc = 0x20000002
>>> Before start, pc = 0x20000002
    Inside hook, addr = 0x20001000, pc = 0x20000002
>>> Before start, pc = 0x20000002
    Inside hook, addr = 0x20001000, pc = 0x20000002
```

[...]

Issue #969, opened 2018-06-06
Wrong PC on ARM after stop from hook


abiondo:
When stopping the emulation from a hook, the PC is correct within the hook, but it's then set to the beginning of the basic block when reading it outside the hook.

Here's a minimal testcase:

```python
#!/usr/bin/env python

from __future__ import print_function
from unicorn import *
from unicorn.arm_const import *

# +0: mov r0, #0
#     loop:
# +4:   add r0, r0, 1
# +8:   b loop
CODE = b'\x4f\xf0\x00\x00\x00\xf1\x01\x00\xfc\xe7'
ADDRESS = 0x10000

# Alternates between True and False so that the hook
# does not hit when continuing
trigger = True


def hook_code(uc, address, size, user_data):
    global trigger
    if trigger:
        pc = uc.reg_read(UC_ARM_REG_PC)
        print('    Inside hook, pc = 0x{:x}'.format(pc))
        uc.emu_stop()
    trigger = not trigger


def smoke_test():
    uc = Uc(UC_ARCH_ARM, UC_MODE_THUMB)
    uc.mem_map(ADDRESS, 2 * 1024 * 1024)
    uc.mem_write(ADDRESS, CODE)
    uc.hook_add(UC_HOOK_CODE, hook_code, begin=ADDRESS+8, end=ADDRESS+8)

    uc.reg_write(UC_ARM_REG_PC, ADDRESS)

    for _ in range(5):
        pc = uc.reg_read(UC_ARM_REG_PC)
        print('>>> Before start, pc = 0x{:x}'.format(pc))
        uc.emu_start(pc | 1, ADDRESS + len(CODE))
        r0 = uc.reg_read(UC_ARM_REG_R0)
        print('    After stop, r0 = {}'.format(r0))


if __name__ == '__main__':
    smoke_test()
```

I'm trying to do something like a breakpoint: stop at `0x10008` from the hook, then do something from outside, then resume emulation. The output is:

```
>>> Before start, pc = 0x10000
    Inside hook, pc = 0x10008
    After stop, r0 = 1
>>> Before start, pc = 0x10000
    Inside hook, pc = 0x10008
    After stop, r0 = 2
>>> Before start, pc = 0x10004
    Inside hook, pc = 0x10008
    After stop, r0 = 4
>>> Before start, pc = 0x10004
    Inside hook, pc = 0x10008
    After stop, r0 = 6
>>> Before start, pc = 0x10004
[...]

PR #1257, opened 2020-05-14
Fix issue with some memory hooks and PC register

elp0t0r1c0:
In the context of a mem hook, the PC register value was sometimes wrong
(maybe the following issues are related to this one: #980, #972, #969, not sure)

This is because the generated code does not track by default the PC values
(stored inside `TCGContext::gen_opc_pc`). It does when we ask unicorn
to hook code or basic block (`gen_intermediate_code_pc` is called for the
specific blocks depending on the specified hook addresses).

I'm proposing the following solution: when we detect an invalid memory access
(READ, WRITE), we compute again the tc using `gen_intermediate_code_pc` just
for the current block, and then we set PC using `uc->set_pc`.

Another solution could be to track every single ld/str inside each blocks where
an hook could be triggered, but I think this is more expensive than the
solution I propose.


elp0t0r1c0:
You can find two tests inside [tests.zip](https://github.com/unicorn-engine/unicorn/files/4630527/tests.zip) (arm64 and x86) which demonstrate the bug.

gerph:
I have believed for some time that when exceptions occur (ARM data abort, specifically), the PC is not set correctly unless there is some form of tracing in use. So I think there's probably a need to do it on all memory operations. However, I haven't got around to putting together a test case to prove this, as most of the time when my code is running it's not generating exceptions.

I will try to create some tests that show whether the PC is incorrect on aborts (which may be independant from this change, but I mention it here in case it helps in understanding).

elp0t0r1c0:
Yeah, there are basically these two solutions. I cannot know which one is better.! It depends on your emulated code I guess (if you never fail on mem access / never catch any exceptions, or at the contrary if you have multiple fails everywhere).

elp0t0r1c0:
What do you think about it, @aquynh ?

aquynh:
let me look at this

aquynh:
this is a nice idea, but it only solves invalid memory access case.

i think we should solve this problem completely, by tracking every memory events. more fine grain check can mitigate performance penalty as much as possible.

elp0t0r1c0:
> this is a nice idea, but it only solves invalid memory access case.
> 
> 
> 
> i think we should solve this problem completely, by tracking every memory events. more fine grain check can mitigate performance penalty as much as possible.

agreed.
If I understand right, the solution should be to generate blocks with PC for each mem instruction inside any memhook range ?
For example with that code:

```
MOV X1, X0          ; #1
LDP X2, X3, [X1]  ; #2 with PC
ADD X2, X2, X3   ; #3
SUB X2, X2, #1
STR X2, [SP]        ; #4 with PC 
```

we should have 4 blocks ?

wtdcode:
Closed due to inactive and partly fixed in uc2.

PR #1262, opened 2020-05-20
Oss-fuzz ideal integration

catenacyber:
This PR does multiple things :
- adds CIFuzz (GitHub workflow to test Pull Requests with fuzzers)
- Fixes one memory leak in watchpoint for ARM. This leak should be fixed in other architectures
- Adds builds to Travis, with fuzz targets and sanitizers, and run the fuzz targets on the public corpus (to be tested)


aquynh:
merged, thanks!

PR #1265, opened 2020-05-22
Oss-fuzz ideal integration

catenacyber:
Adds oss-fuzz badge to REAMDE.md
And 
Better fix for watchpoints leak (generic to all architectures and not only ARM)


aquynh:
merged, thanks!

PR #1270, opened 2020-05-27
PPC support

simigo79:
As far as I remember there were some global name conflicts between ppc32 and ppc64, so I removed ppc64 from the build to speed up ppc32 development. Most likely the conflicts can be solved by updating header_gen.py
Also only one ppc platform is implemented (e500 which is kind of "generic nxp") by explicitly initializing its types in generic ppc_cpu_register_types() function. For ppc64 one has to select a different platform (e500 is 32 bit) and replace e500 types with the ones for selected platform. Or better implement a way to select ppc platform. Also I replaced e500 mmu with "real" mmu (that is physical == virtual dummy). For the selected ppc64 platform one has to replace its mmu with "real" one as well. (I think that mmu emulation is impractical for unicorn).

I knew about unused files but I didn't remove them because some of them would be needed if someone wants to implement other flavors of ppc.

Also there is pointer_fix.py script left in qemu/target-ppc directory. It is not needed. Please remove it, it is one of the scripts I used to process qemu sources for unicorn.


aquynh:
there is a conflict. looks like you should clone the next branch, and try again.

aquynh:
do you continue to add PPC64 support?

simigo79:
It looks like qemu/target-ppc/pointer_fix.py file is the reason of the conflict. Please remove it, it is not needed. (I'll clone the next branch anyway)
I can try to add 5500 ppc64. But I'm not going to do it quickly.

aquynh:
ok i removed qemu/target-ppc/pointer_fix.py

aquynh:
thanks to @studoot, now Windows build with MSVC works.

aquynh:
@simigo79 i put your name into CREDITS.TXT: https://github.com/unicorn-engine/unicorn/commit/857b155bb843f7b2aa025ceef8dde3ead1b4944c

aquynh:
@simigo79 you may interest in this work: https://github.com/DGA-MI-SSI/unicorn-ppc/issues/1

simigo79:
Well, probably their work is much less "quick and dirty" than mine. 
My opinion is: I have some quite complicated simulations  (working with my PPC support. I can try to clone their version and try it. If it works OOTB then probably you should replace my version with theirs.

aquynh:
cool, keep me updated!

simigo79:
I've tried their version with several simulations. No error so far. Also their version seems to be somewhat (10% or so) faster, not sure if it's faster PPC emulation or some generic unicorn code.
Please note that samples directory contents is missing in their version.

minusthreesixty:
> I've tried their version with several simulations. No error so far. Also their version seems to be somewhat (10% or so) faster, not sure if it's faster PPC emulation or some generic unicorn code.
> Please note that samples directory contents is missing in their version.

Thanks for checking our project out ! The missing samples problem has been fixed in https://github.com/DGA-MI-SSI/unicorn-ppc/commit/d5464c0024de9c47c689c60000d50d3012e1afaf

simigo79:
My overall opinion at the present time is that the DGA version is (alas) better than mine.
If I find some problems in either version, I'll notify you.

wtdcode:

PR #1271, opened 2020-05-27
Make MSVC happy with the new PowerPC code

### LINKED ISSUE #1269: PowerPC fails to build with MSVC
we now support PowerPC as a new target, and the code is available in our "next" branch: https://github.com/unicorn-engine/unicorn/tree/next

however, MSVC does not like the new PowerPC code, as you can find in the output at
https://travis-ci.org/github/unicorn-engine/unicorn/jobs/691705267. 
as a result, we cannot build Unicorn natively for Windows.

if you can help to make MSVC happy, please do pull request.

studoot:
This aims to fix #1269 (PowerPC fails to build with MSVC). The problems found are:

1. Enums are 32bit when compiled by MSVC (because Windows uses LLP64, not LP64, so ints are 32 bits wide), so enum definitions that contain enumerations with values wider than 32bits have been changed to be macros instead.
2. MSVC is has different names for some functions that GCC has.
3. The MSVC preprocessor is basically broken in places, so worked around that.
4. MSVC doesn't understand GCC function attributes.

### DIFF
--- a/qemu/target-ppc/cpu.h
+++ b/qemu/target-ppc/cpu.h
@@ -1823,10 +1823,9 @@ static inline int cpu_mmu_index (CPUPPCState *env)
 
 /*****************************************************************************/
 /* PowerPC Instructions types definitions                                    */
-enum {
-    PPC_NONE           = 0x0000000000000000ULL,
+#define PPC_NONE (0x0000000000000000ULL)
     /* PowerPC base instructions set                                         */
-    PPC_INSNS_BASE     = 0x0000000000000001ULL,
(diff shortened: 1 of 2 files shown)


aquynh:
merged, thanks!

aquynh:
btw, we are still missing ppc-softmmu.vcxproj & ppc-softmmu.vcxproj.filters under https://github.com/unicorn-engine/unicorn/tree/next/msvc/unicorn/ppc-softmmu, as well as some update in unicorn.sln for MSVC build. 

if you can, please help to fix them, thanks!

studoot:
PR #1272 raised for Visual Studio project files...

PR #1272, opened 2020-05-28
Add Visual Studio project files for PowerPC support

studoot:
Added vcxproj/vcxproj.filters files for PowerPC code within Unicorn and also the PowerPC sample file.


aquynh:
merged, thanks!

please put your name into https://github.com/unicorn-engine/unicorn/blob/next/CREDITS.TXT.

PR #1280, opened 2020-06-04
adding pypi workflow

chfl4gs:
**Purpose:**
Utilizing github action to build and publish python distribution packages including source package to pypi.org  on repo [tag]. Build will still be triggered on [push].

Binary distribution packages:
sdist, win32, win-amd64, manylinux1_x86_64, manylinux_i686, macos_x86_64

**Preparation:**
1. Go to https://pypi.org/manage/account/#api-tokens and create a new API token for the Unicorn project on PyPI, limit the token scope to just this project. You can call it something like GitHub Actions CI/CD or anything that can be easily distinguishable in the token list. Don’t close the page just yet — you won’t see that token again.

2. In a separate browser tab or window, go to the Github Settings tab of unicorn-engine/unicorn repository and then click on Secrets in the left sidebar.

3. Create a new secret called pypi_pass and copy-paste the token generated from the first step.

**Build tools/Dockers**
Using dockcross's manylinux1 images for Linux build. pypa manylinux version can be found [here](https://github.com/chfl4gs/unicorn/actions/runs/126324927/workflow).

**Artifacts**
Workflow will produce zipped artifacts on each builds for download/inspection/testing. Note: unicorn-packages-macos-latest-x32 artifact is just a source package.

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -1,0 +1,41 @@
+name: Build and publish pypi package
+
+on: [push]
+
+jobs:
+  deploy:
+
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [windows-latest, ubuntu-latest, macos-latest]
+        platform: [x32, x64]
+        exclude:
+          - { platform: x32, os: macos-latest }
+          - { platform: x32, os: ubuntu-latest }  
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
+    - name: Set up MSVC
+      if: matrix.os == 'windows-latest'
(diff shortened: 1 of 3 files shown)


aquynh:
@rhelmot do you have any comments on this pull req, and also on Linux support missing?

rhelmot:
You aren't supposed to build linux binaries outside the [manylinux docker images](https://github.com/pypa/manylinux) in order to make them work cross-distribution. If you do this then you'll be able to do the i386 builds too since there's an i386 manylinux docker image.

chfl4gs:
> You aren't supposed to build linux binaries outside the [manylinux docker images](https://github.com/pypa/manylinux) in order to make them work cross-distribution. If you do this then you'll be able to do the i386 builds too since there's an i386 manylinux docker image.

Yup, switching to dockcross/manylinux1 dockers for for Linux builds. Please find the update.

rhelmot:
lgtm

aquynh:
merged, thanks!

PR #1278, opened 2020-06-03
cmake: fix static linking due to cyclic deps

Johnnynator:
linking of samples like sample_arm do fail due to cyclic references in
the archives

<details><summary>Failing log for reference</summary>
```
[1/2] : && /usr/lib/ccache/bin/clang   -rdynamic CMakeFiles/sample_arm.dir/samples/sample_arm.c.o  -o sample_arm  libunicorn.a  -lpthread  libarm-softmmu.a  libarmeb-softmmu.a  -lm && :
FAILED: sample_arm 
: && /usr/lib/ccache/bin/clang   -rdynamic CMakeFiles/sample_arm.dir/samples/sample_arm.c.o  -o sample_arm  libunicorn.a  -lpthread  libarm-softmmu.a  libarmeb-softmmu.a  -lm && :
/usr/bin/x86_64-unknown-linux-musl-ld: libarm-softmmu.a(tcg.c.o):(.data.rel.ro+0x2f88): undefined reference to `helper_div_i32'
/usr/bin/x86_64-unknown-linux-musl-ld: libarm-softmmu.a(tcg.c.o):(.data.rel.ro+0x2fa0): undefined reference to `helper_rem_i32'
/usr/bin/x86_64-unknown-linux-musl-ld: libarm-softmmu.a(tcg.c.o):(.data.rel.ro+0x2fb8): undefined reference to `helper_divu_i32'
/usr/bin/x86_64-unknown-linux-musl-ld: libarm-softmmu.a(tcg.c.o):(.data.rel.ro+0x2fd0): undefined reference to `helper_remu_i32'
/usr/bin/x86_64-unknown-linux-musl-ld: libarm-softmmu.a(tcg.c.o):(.data.rel.ro+0x2fe8): undefined reference to `helper_div_i64'
/usr/bin/x86_64-unknown-linux-musl-ld: libarm-softmmu.a(tcg.c.o):(.data.rel.ro+0x3000): undefined reference to `helper_rem_i64'
/usr/bin/x86_64-unknown-linux-musl-ld: libarm-softmmu.a(tcg.c.o):(.data.rel.ro+0x3018): undefined reference to `helper_divu_i64'
/usr/bin/x86_64-unknown-linux-musl-ld: libarm-softmmu.a(tcg.c.o):(.data.rel.ro+0x3030): undefined reference to `helper_remu_i64'
[...]

PR #1281, opened 2020-06-04
cmake: fix static linking due to cyclic deps

Johnnynator:
linking of samples like sample_arm do fail due to cyclic references in
the archives

PR for next branch as request in #1278


aquynh:
merged, thanks!

PR #1279, opened 2020-06-04
Adding Travis-CI cmake static build

chfl4gs:
cmake static build failed with error [message](https://travis-ci.com/github/chfl4gs/unicorn/jobs/343337504#L242)

Travis-CI cmake version 3.12.4

```
CMake Error at CMakeLists.txt:924 (install):

  install TARGETS given no ARCHIVE DESTINATION for static library target

  "unicorn".

-- Configuring incomplete, errors occurred!
```

PR #1282, opened 2020-06-04
cmake: add ARCHIVE rule to the install command

scribam:
Should fix the issue mentioned in https://github.com/unicorn-engine/unicorn/pull/1279


aquynh:
merged, thanks!

PR #1284, opened 2020-06-04
cmake: fix build if workdir is outside of CMAKE_BINARY_DIR

Johnnynator:
this e.g. happens if you would try to compile this like this:

cmake -B build
cmake --build build --config Release

@scribam Should fix the remaining CI failures of https://github.com/Vita3K/Vita3K/pull/727

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -186,6 +186,7 @@ else()
     execute_process(COMMAND sh ${CMAKE_CURRENT_SOURCE_DIR}/qemu/configure
         ${EXTRA_CFLAGS}
         ${TARGET_LIST}
+        WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
     )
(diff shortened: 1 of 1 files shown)


scribam:
@Johnnynator 
:thinking: Have you tried it? It does not work for me locally :confused: 
But I think instead it is necessary to set a `WORKING_DIRECTORY` for this command `COMMAND sh ${CMAKE_CURRENT_SOURCE_DIR}/qemu/configure` to fix the failures.

Johnnynator:
Mh... I did try, maybe smth of a different build leaked in. But in that case it should be needed for both.

Johnnynator:
This doesn't make any sense why the previous change did work for me. The create_config script is totally workdir independent and just takes stdin/stdout, cmake itself already works with the correct paths.

aquynh:
merged, thanks!

PR #1286, opened 2020-06-06
Fixing minor typo on PyPI distribution workflow

chfl4gs:
A minor typo on github action step for publishing to pypi.org. It should be **user** instead of **username**. My bad.

```
uses: pypa/gh-action-pypi-publish@master
with:
    user: __token__
    password: ${{ secrets.pypi_pass }}
    packages_dir: ${{ github.workspace }}/bindings/python/dist/
```


aquynh:
merged, thanks!

PR #1287, opened 2020-06-06
remove qobject, qapi, qom

chenhuitao:
It's a huge commit.
remove qobject, qapi, qom..
update make and cmake.


aquynh:
merged to "test2" branch, thanks!

PR #1288, opened 2020-06-07
remove apci, default-congfigs

chenhuitao:
remove apci, default-congfigs.
update make and cmake.
remove owner arg in memory.
fix a mem-leak bug.


aquynh:
merged, thanks!

PR #1295, opened 2020-06-11
Rust bindings

pr0me:
This PR introduces enhanced Rust bindings for Unicorn.

They are based on https://github.com/unicorn-rs/unicorn-rs/ (which is no longer maintained ) and were developed as part of https://github.com/fgsect/BaseSAFE .

There are no dedicated sample cases but example code can be found as part of tests/unicorn.rs.
These bindings also come with some example utility functionalities, such as a basic drop-in allocator
utilizing Unicorn hooks for sanitization.


aquynh:
nice work, thanks!

a quick question: do you want to add Rust support to https://github.com/unicorn-engine/unicorn/blob/master/bindings/const_generator.py, so you dont have to maintain the consts? FYI, whenever the core update anything, i just go to bindings/, and run "make" to update all the bindings.

pr0me:
Thanks for the heads-up!
I chose to maintain them manually as some require individual modifications (especially unicorn_const) which were not trivially automatable from within const_generator.py.

aquynh:
merged, thanks!

bet4it:
Why do you introduce the `data` field and all functions in `utils.rs` such as `heap_oob`?
The are only useful for your repository and nobody else know how to use them, they shouldn't be included in this official repository.

PR #1296, opened 2020-06-13
PyPI test check

chfl4gs:
Adding test pypi step to check if uploads are in order then only publish to pypi.org

**Required**
secret for test pypi token named "**testpypi_pass**" to be created.


aquynh:
merged, thanks!

PR #1300, opened 2020-06-22
fix 64-bit fstenv

liangjs:
FSTENV and FNSTENV instruction in 64-bit and 32-bit mode should be handled in the same way.

https://stackoverflow.com/a/50460464/11790783

### DIFF
--- a/qemu/target-i386/fpu_helper.c
+++ b/qemu/target-i386/fpu_helper.c
@@ -1008,9 +1008,8 @@ void helper_fstenv(CPUX86State *env, target_ulong ptr, int data32)
         }
     }
 
-    // DFLAG enum: tcg.h, case here to int
-    if ((env->hflags & HF_CS64_MASK) || data32) {
-        /* 32 or 64 bit */
+    if (data32) {
+        /* 32 bit */
         cpu_stl_data(env, ptr, env->fpuc);
         cpu_stl_data(env, ptr + 4, fpus);
         cpu_stl_data(env, ptr + 8, fptag);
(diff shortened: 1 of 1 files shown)


aquynh:
can you look at Qemu 5.0, and see if this code is the same?

liangjs:
Related Qemu code is [here](https://github.com/qemu/qemu/blob/7d3660e79830a069f1848bb4fa1cdf8f666424fb/target/i386/fpu_helper.c#L1244).
Unicorn diverges from Qemu in [this](https://github.com/unicorn-engine/unicorn/commit/b7d60313b5ccd49e3a3d87c7190e011214089c60) commit.
Actually there is no 64-bit version of FSTENV/FNSTENV instruction. So in x86-64 executions, the program also uses 32-bit version and `data32` is set.

aquynh:
> Related Qemu code is [here](https://github.com/qemu/qemu/blob/7d3660e79830a069f1848bb4fa1cdf8f666424fb/target/i386/fpu_helper.c#L1244).
> Unicorn diverges from Qemu in [this](https://github.com/unicorn-engine/unicorn/commit/b7d60313b5ccd49e3a3d87c7190e011214089c60) commit.
> Actually there is no 64-bit version of FSTENV/FNSTENV instruction. So in x86-64 executions, the program also uses 32-bit version and `data32` is set.

so the testcase in regress/fpu_ip64.py  is also broken?
can you fix that too?

liangjs:
The fpu_ip test is still correct, no need to fix.

aquynh:
merged, thanks!

PR #1305, opened 2020-07-12
Fix failing test caused by removal of UC_ERR_TIMEOUT

ss23:
Caused by fbef45b18f2f59d9699af54c6ad6ac73eaba3b29.
Now the code assumes if it completes without an error, it was successful.


aquynh:
merged, thanks!

PR #1306, opened 2020-07-13
Return 'indefinite integer value' for invalid SSE fp->int conversions

liangjs:
Merge from qemu commit 1e8a98b53867f61da9ca09f411288e2085d323c4.
https://github.com/qemu/qemu/commit/1e8a98b53867f61da9ca09f411288e2085d323c4


aquynh:
merged, thanks!

PR #1309, opened 2020-07-29
Improved Rust bindings

Nikolasel:
* Added basic block hooking
* Changed confusing struct naming. Before: Protection::All -> R,W,X, Now: Permission::All -> R,W,X
* Fixed issue with remove_hook(..). Implementation tried to remove hook from incorrect hashmap.
* Made unused private vmmap(..) public.

### DIFF
--- a/bindings/rust/README.md
+++ b/bindings/rust/README.md
@@ -6,14 +6,14 @@ An extended version for fuzzing with AFL++ support can be found in https://githu
 
 ```rust
 use unicorn::RegisterARM;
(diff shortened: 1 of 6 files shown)


aquynh:
merged, thanks!

domenukk:
Why `Permission` for protection bits? This seems wrong. Internally, uc also uses protection everywhere, for example here https://github.com/unicorn-engine/unicorn/blob/f1f59bac5542776fe85fb225a88d5cc623f89b87/qemu/translate-all.c#L1900

Issue #1313, opened 2020-08-02
Weird hang while running neon instructions with pld


sunho:
Hello! I'm from https://github.com/Vita3K/Vita3K. 

We experienced weird hang while running arm thumb2 code. The relevant code was like this.

```asm
lsrs r6, r6, #5
lsl.w r8, r6, #4
cmp r6, #0
beq.w #0x344
vst2.8 {d20, d21, d22, d23}, [lr]
vst2.8 {d16, d17, d18, d19}, [sb]
mov lr, r3
mov r0, r1
mov.w ip, #0
add.w sb, lr, #1
vld2.8 {d20, d21, d22, d23}, [lr]
add.w ip, ip, #1
pld [r0, #0x90]
vld2.8 {d16, d17, d18, d19}, [sb]
cmp ip, r6
vshll.i8 q12, d20, #8
add.w sb, r0, #0x10
vshll.i8 q10, d21, #8
add.w lr, lr, #0x20
vmovl.u8 q11, d16
vmovl.u8 q8, d17
vorr q9, q11, q12
vorr q8, q8, q10
vst1.16 {d18, d19}, [r0]
add.w r0, r0, #0x20
vst1.16 {d16, d17}, [sb]
``` 

I discovered that if I run this code in simpler setting with code hook enabled, it also hangs. Following code is what I used. I also discovered that if I remove "pld [r0, #0x90]", it can complete even with code hook. (```ARM_CODE = CODE_WITHOUT_PLD```)

```python
from __future__ import print_function
from unicorn import *
from unicorn.arm_const import *

CODE_WITHOUT_PLD = bytes(bytearray.fromhex("76094FEA0618002E00F0A2814EF90F4349F90F039E4608464FF0000C0EF101096EF90F430CF1010C69F90F03B445F2FF248300F11009F2FF25430EF1200EC8FF306AC8FF310A66EFF82160EFF40140F94F2A00F1200049F94F0A"))
CODE_WITH_PLD = bytes(bytearray.fromhex("76094FEA0618002E00F09C814EF90F4349F90F039E4608464FF0000C0EF101096EF90F430CF1010C90F890F069F90F03B445F2FF248300F11009F2FF25430EF1200EC8FF306AC8FF310A66EFF82160EFF40140F94F2A00F1200049F94F0A"))
ARM_CODE = CODE_WITH_PLD
ADDRESS    = 0x10000

def hook_code(uc, address, size, user_data):
[...]

PR #1315, opened 2020-08-11
Fix enabled hooks even after deleting them

sunho:
When you uc_emu_start with count more than 0 and then uc_emu_start with count = 0, to be deleted hooks are not processed yet (because it's processed at the end of uc_emu_start) still have effects. 

Because of this issue, code hook internally inserted by unicorn to count the instructions are not really deleted when we uc_emu_start with count = 0. This "leftover" hook inserted monitoring tcg between every instruction and resulted in the hang described in https://github.com/unicorn-engine/unicorn/issues/1313, even if we didn't add any code hook.

This PR fixes this issue by checking to_delete field in HOOK_BOUND_CHECK, so that unicorn can ignore to be deleted hooks.

### DIFF
--- a/include/uc_priv.h
+++ b/include/uc_priv.h
@@ -127,7 +127,7 @@ enum uc_hook_idx {
 // if statement to check hook bounds
 #define HOOK_BOUND_CHECK(hh, addr)                  \
     ((((addr) >= (hh)->begin && (addr) <= (hh)->end) \
-         || (hh)->begin > (hh)->end))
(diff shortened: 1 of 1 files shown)


aquynh:
do you mean you call uc_emu_start() in nested way (i.e call uc_emu_start() from inside Unicorn callback)?

sunho:
1. Call emu_start with count = 1
2. Call emu_start with count = 0

Then one code hook (count hook) added internally by unicorn still affacts unicorn in (2).

aquynh:
merged, thanks!

PR #1317, opened 2020-08-19
fix an issue with ARM APSR register handling

h01G3r:
- Q flag / GE flag were not included in APSR register (read/write)
  - UC_ARM_REG_APSR_NZCV register constant was ignored completely.
  - regression test added


aquynh:
merged, thanks!

Issue #1320, opened 2020-08-31
Wrong IP value in hook callback with UC_ARCH_X86 and UC_MODE_16


wtdcode:
The code to reproduce this bug:

```C++
#include <cstdio>
#include "unicorn/unicorn.h"

const char* cmd = "\x90\x90\x90";
int count = 1;

void read_regs(uc_engine* uc){
	short ip;
	short cs;
	short sp;
	short ss;
	uc_reg_read(uc, UC_X86_REG_SS, &ss);
	uc_reg_read(uc, UC_X86_REG_SP, &sp);
	uc_reg_read(uc, UC_X86_REG_IP, &ip);
	uc_reg_read(uc, UC_X86_REG_CS, &cs);
	printf("cs:ip=%hx:%hx ss:sp=%hx:%hx\n", cs, ip, ss, sp);
}

void cb(uc_engine* uc, uint64_t addr, uint32_t size, void* user_data){
	printf("The %d hook_code callback:\n", count);
	read_regs(uc);
	printf("addr: 0x%lx\n", addr);
	count += 1;
}

int main(){
	uc_engine* uc;
	uc_err err;
	printf("Start\n");
	err = uc_open(UC_ARCH_X86, UC_MODE_16, &uc);
	if (err != UC_ERR_OK){
		printf("Failed to open uc\n");
		return -1;
	}
	err = uc_mem_map(uc, 0x7000, 4*1024, UC_PROT_ALL);
	if (err != UC_ERR_OK){
		printf("Failed to allocate memory with %d\n", err);
		return -1;
	}
	int ip = 0x100;
	int cs = 0x75a;
	int address_to_load = cs*16 + ip;
	err = uc_mem_write(uc, address_to_load, cmd, 3);
	if (err != UC_ERR_OK){
		printf("Failed to write memory with %d\n", err);
		return -1;
	}
	err = uc_reg_write(uc, UC_X86_REG_IP, &ip);
	if (err != UC_ERR_OK){
		printf("Failed to write register with %d\n", err);
		return -1;
	}
	err = uc_reg_write(uc, UC_X86_REG_SS, &cs);
	if (err != UC_ERR_OK){
		printf("Failed to write register with %d\n", err);
		return -1;
	}
	err = uc_reg_write(uc, UC_X86_REG_SP, &ip);
	if (err != UC_ERR_OK){
		printf("Failed to write register with %d\n", err);
		return -1;
	}
[...]

PR #1321, opened 2020-08-31
Fix IP value for UC_MODE_16

wtdcode:
Fix for #1320.

I'm not sure whether it's the right place to fix. Any instruction or advice is highly welcome.


aquynh:
merged, thanks!

PR #1322, opened 2020-09-01
Fixing installation on cygwin, Travis-ci cache adjustment

chfl4gs:
1. zip_safe to False on python binding as contains binary lib. Some older OS might not able to use zipped egg with binary lib.

2. Fixing python binding build and installation under cygwin, correcting lib filename and path.

3. Adjusting travis-ci cache, and temporarily disabling unittest on cmake static build.

4. disabling unittest on cygwin clang as cmocka does not build properly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,17 @@ env:
 script:
   - |
     if [[ "$TRAVIS_OS_NAME" == "windows" ]]; then
-        choco install cygwin cyg-get && \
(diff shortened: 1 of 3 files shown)


chfl4gs:
See build result
https://travis-ci.org/github/unicorn-engine/unicorn/builds/722948913

aquynh:
merged, thanks!

PR #1326, opened 2020-09-15
Add qemu_ prefix to bitmap_set and bitmap_clear to avoid conflicts.

mogoreanu:
When unicorn and systemd are combined into a single binary the 2
libraries conflict on bitmap_set and bitmap_clear functions which breaks
unicorn.


aquynh:
merged, thanks!

PR #1327, opened 2020-09-15
Removed unused bitmap_set rename bitmap_clear to qemu_bitmap_clear.

mogoreanu:
When unicorn and systemd are included into a single binary they conflict
on the bitmap_clear function which breaks unicorn. It seems that
long-term we can remove the qemu_bitmap_clear altogether and leave only
bitmap_zero_extend.


aquynh:
does this and https://github.com/unicorn-engine/unicorn/pull/1326 do the same thing?

StalkR:
Yes just for different branches

- #1326 is for the master branch
- #1327 is for the next branch

aquynh:
merged, thanks!

PR #1331, opened 2020-09-21
Fix setjmp/longjmp on native Windows

wtdcode:
## Why do we need this wrapper?

Short answer: Windows default implementation of setjmp/longjmp is incompatible with jit generated code.
A longer answer: https://blog.lazym.io/2020/09/21/Unicorn-Devblog-setjmp-longjmp-on-Windows/.

Note that x86 setjmp/longjmp is implemented without RtlUnwind so there is no need to modify that.


wtdcode:
Now works with cmake + VS 2017 x64 / nmake, cygwin, MinGW with msys2.

@aquynh

aquynh:
merged, thanks!

PR #1335, opened 2020-09-23
Fix context saving

wtdcode:
1. Fix `uc_context` size in uc.c.
2. Make `UcContext` convertible to bytes.
3. Make `UcContext` picklable.
4. Fix `context_update` in python bindings.

### DIFF
--- a/bindings/python/sample_x86.py
+++ b/bindings/python/sample_x86.py
@@ -4,7 +4,7 @@
 from __future__ import print_function
 from unicorn import *
 from unicorn.x86_const import *
-
+import pickle
 
 X86_CODE32 = b"\x41\x4a\x66\x0f\xef\xc1" # INC ecx; DEC edx; PXOR xmm0, xmm1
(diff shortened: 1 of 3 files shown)


wtdcode:
@aquynh 
One question: Why do you decide to save `jmp_buf` in #2e0f753e6f9411901dce1e6779b0bcdd57d13c09? That makes the saved context tied to a specific `uc_engine`.

xwings:
This solutions works only if Unicorn and and Qiling in the same instance. Most of the time we are not and we might not.

We used to be able to do this in 1.0.2-rc4
1. a.py run save state to a file and stop and generate a_state.bin
2. b.py load a_state.bin and run fuzzer.

Unicorn Afl runs in different instance. So, this method will kill all the fuzzing ability.

Solutuion, saving jmp_buf is just a option and not compulsory ?

aquynh:
merged, thanks!

PR #1336, opened 2020-09-24
Implement uc_context_free

wtdcode:
This PR follows #1335.

WARNING: This PR introduces a break change: contexts allocated by `uc_context_alloc` now should be free-ed by `uc_context_free`. uc_free() the contexts will still work in some scenarios, but the result is undefined.


aquynh:
merged, thanks!

PR #1338, opened 2020-10-01
Fix compile if HAS_ARM is defined but HAS_ARM_EB isn't

bpfoley:
This allows us to build LE-only ARM unicorn variants

### DIFF
--- a/uc.c
+++ b/uc.c
@@ -193,7 +193,11 @@ uc_err uc_open(uc_arch arch, uc_mode mode, uc_engine **result)
                     return UC_ERR_MODE;
                 }
                 if (mode & UC_MODE_BIG_ENDIAN) {
+#ifdef UNICORN_HAS_ARMEB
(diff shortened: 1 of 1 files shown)


aquynh:
which command did you use to build?

bpfoley:
I did this as part of some custom build scripts to build Unicorn as a part of a library where I only wanted a Little Endian ARM support and no other platform. To do that I used -DUNICORN_HAS_ARM but not -DUNICORN_HAS_ARMEB.

This seems like a fairly reasonable thing that others may want to do given that ARMEB is relatively unusual.

aquynh:
you can put that into this PR too, so it will be useful for others?

bpfoley:
I'm not sure they're that useful. The scripts are actually Google-internal [bazel](https://bazel.build/) build rules. They're good enough for what we need, but aren't configurable enough for general purpose use.

aquynh:
merged, thanks!

PR #1347, opened 2020-10-24
Fix build with LLVM on powerpc64(le)

pkubaj:
Conflicting symbols need to be undefined before including altivec.h.

Without it, build fails with:
`error: unknown type name 'vector'`


aquynh:
merged, thanks!

PR #1351, opened 2020-11-02
wrapping cpu_exit call with mutex to prevent race condition

oblivia-simplex:
While using Unicorn in series of genetic programming experiments, my research group would occasionally run into segmentation faults. These seemed to occur only when the Unicorn emulation called the `uc_emu_stop()` function, from a watchdog thread separate from the main emulation thread. This function checks to ensure that `uc->current_cpu` is not null, and _then_ calls `cpu_exit(uc->current_cpu)`. This led us to suspect a race condition, whereby, after the check but before the call, `uc->current_cpu` was made null by events unfolding on another thread. Wrapping this critical section in a mutex lock, as done here, appears to have fixed this issue, and we're no longer seeing any segfaults.

### DIFF
--- a/uc.c
+++ b/uc.c
@@ -675,6 +675,8 @@ uc_err uc_emu_start(uc_engine* uc, uint64_t begin, uint64_t until, uint64_t time
 }
 
 
+pthread_mutex_t EMU_STOP_MUTEX;
+
 UNICORN_EXPORT
 uc_err uc_emu_stop(uc_engine *uc)
 {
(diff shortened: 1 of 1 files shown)


uc.c
```diff
@@ -674,6 +676,15 @@ uc_err uc_emu_start(uc_engine* uc, uint64_t begin, uint64_t until, uint64_t time
     return uc->invalid_error;
 }
 
+#if defined(_WIN32) || defined(_WIN64)
+int EMU_STOP_MUTEX = 0;
+#define enter_critical_section(_lock)
```
wtdcode, line 681: So Windows doesn't have the corresponding support?

wtdcode:

PR #1354, opened 2020-11-26
ARM, Fix UC_HOOK_MEM_[READ, READ_AFTER, WRITE]

bsekisser:
modified:   qemu/tcg/arm/tcg-target.c

Found hook calls would only happen once due to tlb needing to be
	filled...  Thereafter hook calls would be bypassed.  This
	patch ensures hook calls do not get bypassed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -850,7 +850,7 @@ set(UNICORN_SRCS_COMMON
     uc.c
 )
 
-if (MSVC)
+if (WIN32)
     set(UNICORN_SRCS
         ${UNICORN_SRCS_COMMON}
         qemu/util/oslib-win32.c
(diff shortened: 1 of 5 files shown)


newblaque (approved):

aquynh:
is this only for Arm? do you see this on other architectures?

aquynh:
do you have any PoC code for this bug?

bsekisser:
Arm.  May also apply to others.  X86 does something similar.  Will try to rmember to get back to you later.On Dec 23, 2020 12:38 AM, Nguyen Anh Quynh <notifications@github.com> wrote:
is this only for Arm? do you see this on other architectures?

bsekisser:
Not at this time.On Dec 23, 2020 12:42 AM, Nguyen Anh Quynh <notifications@github.com> wrote:
do you have any PoC code for this bug?

aquynh:
We need a testcase to confirm the bug, and also to avoid future regression

bsekisser:
Test is posted.  Repository on my end is borked.  Let me know how you want me to proceed.

stevielavern:
FYI this also affects ARM64. A similar patch is needed in `qemu/tcg/aarch64/tcg-target.c` so that we always take the "slow path" even in case of TLB hit.

wtdcode:

PR #1360, opened 2020-12-11
Remove deprecated javah and build with javac

iGio90:
Hi, I was having trouble building java bindings on latest ubuntu. I installed jdk8 which has javah but somehow was failing during install (missing compiled classes even if they were in place in unicorn/). I guess mixing with jdk11 pop the issue. 
Using javac from most recent jdk works very fine.


aquynh:
merged, thanks!

PR #1368, opened 2021-01-13
Fix wrong sync after UC_ERR_[READ, WRITE, FETCH]_[UNMAPPED, PROT]

### LINKED ISSUE #1323: x86_64: RIP isn't accurate during/after an unmapped memory hook.
To demonstrate this problem, I've modified the sample application as such:

```
#include <unicorn/unicorn.h>

// code to be emulated
#define X86_CODE64 "\x90\xa1\xef\xbe\xad\xde\x00\x00\x00\x00" // NOP; movabs eax,ds:0xdeadbeef

// memory address where emulation starts
#define CODE_ADDR 0x1000000

static void unmapped_handler(uc_engine *uc, uc_mem_type type,
       uint64_t address, int size, int64_t value, void *user_data)
{
  long int r_rip;

  uc_reg_read(uc, UC_X86_REG_RIP, &r_rip);
  printf(">>> RIP = 0x%x\n", r_rip);

  uc_emu_stop(uc);

  uc_reg_read(uc, UC_X86_REG_RIP, &r_rip);
  printf(">>> RIP = 0x%x\n", r_rip);
}

int main(int argc, char **argv, char **envp)
{
  uc_engine *uc;
  uc_hook unmapped_hook;
  uc_err err;
  long int r_rip;

  // Initialize emulator in X86-32bit mode
  err = uc_open(UC_ARCH_X86, UC_MODE_64, &uc);
  if (err != UC_ERR_OK) {
    printf("Failed on uc_open() with error returned: %u\n", err);
    return -1;
  }

  err = uc_hook_add(uc, &unmapped_hook, UC_HOOK_MEM_READ_UNMAPPED, unmapped_handler, NULL, 1, 0);
  if (err != UC_ERR_OK) {
    printf("Failed on uc_hook_add() with error returned: %u\n", err);
    return -1;
  }

  // map 2MB memory for this emulation
  uc_mem_map(uc, CODE_ADDR, 2 * 1024 * 1024, UC_PROT_EXEC|UC_PROT_READ);

  // write machine code to be emulated to memory
  if (uc_mem_write(uc, CODE_ADDR, X86_CODE64, sizeof(X86_CODE64) - 1)) {
    printf("Failed to write emulation code to memory, quit!\n");
    return -1;
  }

  // emulate code in infinite time & unlimited instructions
  err=uc_emu_start(uc, CODE_ADDR, CODE_ADDR + sizeof(X86_CODE64) - 1, 0, 0);
  if (err) {
    printf("Failed on uc_emu_start() with error returned %u: %s\n",
      err, uc_strerror(err));
  }

  // now print out some registers
  printf("Emulation done. Below is the CPU context\n");

  uc_reg_read(uc, UC_X86_REG_RIP, &r_rip);
  printf(">>> RIP = 0x%x\n", r_rip);

  uc_close(uc);

  return 0;
}
```

At no point am I able to see which instruction this occurs at, the output is as follows:

```
>>> RIP = 0x1000000
>>> RIP = 0x1000000
Emulation done. Below is the CPU context
>>> RIP = 0x1000000
```

wtdcode:
Note that:
1. We only guarantee the pc (and other internal states) is correct
if and only of `uc_emu_start` returns without any error (or errors
have been handled in callbacks.).
2. If memory read/write error isn't handled by hooks, the state is
undefined and the pc is probably wrong if no hook is installed.

This fixes #1323.

### DIFF
--- a/qemu/cpu-exec.c
+++ b/qemu/cpu-exec.c
@@ -327,17 +327,31 @@ static tcg_target_ulong cpu_tb_exec(CPUState *cpu, uint8_t *tb_ptr)
          * or timer mode is in effect, since these already fix the PC.
          */
         if (!HOOK_EXISTS(env->uc, UC_HOOK_CODE) && !env->uc->timeout) {
-            if (cc->synchronize_from_tb) {
-                // avoid sync twice when helper_uc_tracecode() already did this.
-                if (env->uc->emu_counter <= env->uc->emu_count &&
-                        !env->uc->stop_request && !env->uc->quit_request)
-                    cc->synchronize_from_tb(cpu, tb);
-            } else {
-                assert(cc->set_pc);
-                // avoid sync twice when helper_uc_tracecode() already did this.
-                if (env->uc->emu_counter <= env->uc->emu_count &&
-                        !env->uc->stop_request && !env->uc->quit_request)
-                    cc->set_pc(cpu, tb->pc);
+            // We should sync pc for R/W error.
+            switch (env->invalid_error) {
+                case UC_ERR_WRITE_PROT:
+                case UC_ERR_READ_PROT:
+                case UC_ERR_FETCH_PROT:
+                case UC_ERR_WRITE_UNMAPPED:
+                case UC_ERR_READ_UNMAPPED:
+                case UC_ERR_FETCH_UNMAPPED:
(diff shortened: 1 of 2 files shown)


aquynh:
can you document this issue somewhere, maybe in unicorn.h, or in the docs?

wtdcode:
Done.

Issue #785, opened 2017-03-17
ARM BE8 mode


zachriggle:
The ARM spec was written by people who hit their head too many times, and includes a "be8" mode in addition to traditional "big" and "little" endiannesses.  Specifically, BE8 uses little-endian instructions and big-endian data memory. 

If it helps any, qemu-user supports this mode.  ARMv7 either added or removed support for traditional BE32, in favor of BE8, so it's recent and relevant.  The current GCC for Ubuntu targeting ARMBE will emit BE8 binaries.

aquynh:
@xizhizhang, what do you think?

aquynh:
for reference: http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.ddi0290g/ch06s05s01.html

zachriggle:
Relevant patches: https://patchwork.ozlabs.org/patch/143758/

aquynh:
if the patch you provided above is enough, that is not a lot of work to port over. can you do that?

xizhizhang:
It seems qemu have already commit BE8's in 2012:
https://github.com/qemu/qemu/commit/d8fd2954996255ba6ad610917e7849832d0120b7
so, is also in unicorn too.

> BE-32(traditional big endian) is supported by ARM cores up to the ARM11 family (v6) (for example ARM7TDMI, ARM926EJ-S, ARM1136JF-S). BE-8 is supported by the ARM11 family and later (for example ARM1136JF-S, Cortex-R4, Cortex-A8). ARMv7 and ARMv8 cores do not support BE-32.

And, gcc toolchain already choose --be8 as the default link option.

I have just test a BE8 style THUMB2 code in big endian example.
`\xf0\x4f\x00\x37\xeb\xa2\x01\x03`
Above binary means "mov r0, #0x37; sub r1, r2, r3". and it works correctly.

If there are some bugs in ARM big endian mode's implementation, I would like to try to fix it.

Thanks~

zachriggle:
@aquynh That's the patch that adds support for QEMU.  The code is all there already, support just needs to be added to Unicorn.

aquynh:
Zach, can you send a testcase for this?

yekscorp:
I just ran into this problem with a binary I am working with. Is there a plan to add a flag to enable this mode in unicorn?

PR #1369, opened 2021-01-13
Add ARM BE8 support

w4kfu-synacktiv:
Hello,

Here is a PR that add supports to unicorn for the Arm BE8 mode, it will solve the issue https://github.com/unicorn-engine/unicorn/issues/785.

Quick reminder this mode use Big-endian data memory and Little-endian instructions.

BE8 has been already introduced in qemu : https://github.com/qemu/qemu/commit/d8fd2954996255ba6ad610917e7849832d0120b7, this PR just add a new ARM mode, and update `bswap_code` in the `CPUARMState` structure in function `arm_cpu_reset`

Here is an example of its usage

```python
from unicorn import *
from unicorn.arm_const import *

ARM_CODE = b"\x00\x4b\x01\xE0\x11\x22\x33\x44" # ldr r3, [pc, #0] ; b #8 ; dword_11223344
ADDRESS    = 0x10000

def test_armbe8():
    print("Emulate ARM BE8 code")
    mu = Uc(UC_ARCH_ARM, UC_MODE_THUMB | UC_MODE_ARMBE8)
    mu.mem_map(ADDRESS, 2 * 1024 * 1024)
    mu.mem_write(ADDRESS, ARM_CODE)
    mu.reg_write(UC_ARM_REG_R3, 0x00000000)
    mu.emu_start(ADDRESS | 1, ADDRESS + len(ARM_CODE))
    r3 = mu.reg_read(UC_ARM_REG_R3)
    print(">>> As big endian, R3 should be 0x11223344")
    print(">>> R3 = 0x{0:08X}".format(r3))

if __name__ == '__main__':
    test_armbe8()
```

Let me know what you think.

Thx,


wtdcode:
LGTM, a few questions:

1. Does qemu (as of the version unicorn forked) have this support?
2. Why would you like to force update the `bswap_code`?

Also, please add such samples to sample-arm.c. Thanks a lot!

aquynh:
merged, thanks!

PR #1373, opened 2021-02-10
Modernize CMake

mrexodia:
Did my best to modernize the CMake a bit and add options for installation/samples, which will allow users to add unicorn to their CMake projects by doing `add_subdirectory` (or FetchContent). Everything is backwards compatible and when installing unicorn I get the same outputs.

Example project that automatically downloads and compiles unicorn: https://github.com/mrexodia/unicorn_template

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,11 +21,14 @@ set(UNICORN_VERSION_MAJOR 1)
 set(UNICORN_VERSION_MINOR 0)
 set(UNICORN_VERSION_PATCH 2)
 
-option(UNICORN_BUILD_SHARED "Build shared instead of static library" ON)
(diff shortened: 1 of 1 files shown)


wtdcode:
Nice contribution. Just a question: Does it also work with `find_library`?

mrexodia:
Everything is backwards compatible. In general you probably shouldn’t use find_library though, use find_package instead (although I don’t know if that works).

wtdcode:
Oh sry, I mean find_package. Does that work?

mrexodia:
No idea. I just made everything backwards compatible. If the install is
done correctly it’ll probably work.

aquynh:
merged, thanks!

PR #1381, opened 2021-03-30
Add arm ite blocks samples from #853

wtdcode:
Add the sample from #853 to sample-arm.c.

PR #1382, opened 2021-04-02
The UC instance must be created twice to reproduce the problem

wtdcode:
This one fixes the sample submitted in #1381.


aquynh:
merged, thanks!

PR #1385, opened 2021-04-19
Fix fpip

wtdcode:
See my review here for details.

https://github.com/unicorn-engine/unicorn/pull/74/files#r614950124

The patch is submitted to qemu at https://lists.nongnu.org/archive/html/qemu-devel/2021-04/msg02868.html.

No idea the time it is merged in the upstream but I would like to merge it into Unicorn early since many users may emulate MSF generated shellcode which requires correct fpip implementation.

### DIFF
--- a/qemu/target-i386/translate.c
+++ b/qemu/target-i386/translate.c
@@ -6875,6 +6875,8 @@ static target_ulong disas_insn(CPUX86State *env, DisasContext *s,
                 goto illegal_op;
             }
         }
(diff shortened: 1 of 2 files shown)


aquynh:
merged, thanks!

wtdcode:
This fix is proceeded by https://github.com/qemu/qemu/commit/84abdd7d271c2df69a9d394be093efd885da7a4c and would be ported to uc2.

PR #1387, opened 2021-04-21
Handle exceptions raised in Python hook functions

### LINKED ISSUE #1386: Exceptions in Python hooks are ignored
Exceptions occurring in Python hooks are ignored and emulation continues to run. See the following example.

```python
from unicorn import Uc, UC_ARCH_X86, UC_MODE_64, UC_HOOK_CODE

CODE = b"\x90" * 3
CODE_ADDR = 0x1000


def bad_code_hook(uc, address, size, data):
    raise ValueError("Something went wrong")


if __name__ == "__main__":
    uc = Uc(UC_ARCH_X86, UC_MODE_64)
    uc.mem_map(CODE_ADDR, 0x1000)
    uc.mem_write(CODE_ADDR, CODE)

    uc.hook_add(UC_HOOK_CODE, bad_code_hook, begin=CODE_ADDR, end=CODE_ADDR + len(CODE))

    print("Starting emulation")
    uc.emu_start(CODE_ADDR, CODE_ADDR + len(CODE))
    print("Emulation completed successfully")
```

I think the proper behavior would be to stop emulation and raise the exception.

insane-shane:
Closes #1386.

If a hook raises an exception, a call to `emu_stop()` is issued. Before `emu_start()` returns, it will check for an outstanding exception and raise it.

### DIFF
--- a/bindings/python/unicorn/unicorn.py
+++ b/bindings/python/unicorn/unicorn.py
@@ -3,6 +3,7 @@
 import ctypes
 import ctypes.util
 import distutils.sysconfig
+from functools import wraps
 import pkg_resources
 import inspect
 import os.path
@@ -199,6 +200,27 @@ def uc_arch_supported(query):
     return _uc.uc_arch_supported(query)
 
 
+def _catch_hook_exception(func):
+    @wraps(func)
+    def wrapper(self, *args, **kwargs):
+        """Catches exceptions raised in hook functions.
+
+        If an exception is raised, it is saved to the Uc object and a call to stop
+        emulation is issued.
+        """
+        try:
+            return func(self, *args, **kwargs)
+        except Exception as e:
+            # If multiple hooks raise exceptions, just use the first one
+            if self._hook_exception is None:
+                self._hook_exception = e
+
+            self.emu_stop()
+
+    return wrapper
+
+
+
 class uc_x86_mmr(ctypes.Structure):
     """Memory-Management Register for instructions IDTR, GDTR, LDTR, TR."""
     _fields_ = [
@@ -300,6 +322,7 @@ class Uc(object):
         self._ctype_cbs = {}
         self._callback_count = 0
         self._cleanup.register(self)
+        self._hook_exception = None  # The exception raised in a hook
 
     @staticmethod
     def release_handle(uch):
@@ -316,6 +339,9 @@ class Uc(object):
         status = _uc.uc_emu_start(self._uch, begin, until, timeout, count)
         if status != uc.UC_ERR_OK:
             raise UcError(status)
+
+        if self._hook_exception is not None:
+            raise self._hook_exception
 
     # stop emulation
     def emu_stop(self):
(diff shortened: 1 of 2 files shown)


wtdcode:
1. It's users' responsibility to implement such mechanism.
2. Only the first exception is raised, which makes it useless for handling multiple exceptions.

insane-shane:
>     1. It's users' responsibility to implement such mechanism.
Agreed, but there isn't a clean way of doing it currently. The user would have to wrap all their hooks in `try..except` and essentially do what this merge does. Even then, unhandled exceptions won't propagate up because the hook was called from the C library code. The propagation would be useful when the user isn't expecting an error in their code.

>     2. Only the first exception is raised, which makes it useless for handling multiple exceptions.
There isn't a way to raise multiple exceptions at the same time in Python.

wtdcode:
> > ```
> > 1. It's users' responsibility to implement such mechanism.
> > ```
> 
> Agreed, but there isn't a clean way of doing it currently. The user would have to wrap all their hooks in `try..except` and essentially do what this merge does. Even then, unhandled exceptions won't propagate up because the hook was called from the C library code. The propagation would be useful when the user isn't expecting an error in their code.

Leaving the exception handling to users provides more flexibility here. For example, users may handle the exceptions outside the callback like yours but without really stopping the Unicorn. If your PR is merged, users have to stop the engine whatever exception is thrown even if they write their wrappers.

> 
> > ```
> > 2. Only the first exception is raised, which makes it useless for handling multiple exceptions.
> > ```
> 
> There isn't a way to raise multiple exceptions at the same time in Python.

Ah, sry that I forget the engine is stopped in this case.

insane-shane:
> users may handle the exceptions outside the callback like yours but without really stopping the Unicorn

I don't think this is possible without this mege. First, exceptions will not propagate outside the callback because they get ignored when crossing the Python-C threshold. Second, even if users catch the exception (currently this _must_ happen within the callback) and decide to handle it outside the callback, the Unicorn emulation must be stopped, otherwise the program will continue to run.

> If your PR is merged, users have to stop the engine whatever exception is thrown even if they write their wrappers.

That's simply not true. The code in my merge will only execute on _unhandled_ exceptions. The user can still catch exceptions in their callbacks and handle them appropriately within the context of the callback.

The intended use case for this merge is to raise unexpected errors to the user. Having errors be completely (and silently) ignored is [surprising](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) and problematic because an unexpected error means the program is not behaving as intended. Thus the correct course of action is to halt program execution and raise an error. With the way Unicorn is implemented, the only way to do this is to stop emulation so the call to `emu_start()` returns.

wtdcode:
> > users may handle the exceptions outside the callback like yours but without really stopping the Unicorn
> 
> I don't think this is possible without this mege. First, exceptions will not propagate outside the callback because they get ignored when crossing the Python-C threshold. Second, even if users catch the exception (currently this _must_ happen within the callback) and decide to handle it outside the callback, the Unicorn emulation must be stopped, otherwise the program will continue to run.
> 
> > If your PR is merged, users have to stop the engine whatever exception is thrown even if they write their wrappers.
> 
> That's simply not true. The code in my merge will only execute on _unhandled_ exceptions. The user can still catch exceptions in their callbacks and handle them appropriately within the context of the callback.
> 
> The intended use case for this merge is to raise unexpected errors to the user. Having errors be completely (and silently) ignored is [surprising](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) and problematic because an unexpected error means the program is not behaving as intended. Thus the correct course of action is to halt program execution and raise an error. With the way Unicorn is implemented, the only way to do this is to stop emulation so the call to `emu_start()` returns.

I did some tests with your PR and you are right indeed. This `try..catch` should become the last resort to stop the whole engine as expected.

aquynh:
merged, thanks!

PR #1397, opened 2021-05-15
compile on VS2008/VS2010

dnasdw:
Compiling with Win32 Release/MinSizeRel/RelWithDebInfo in VS2008 will cause "fatal error C1063: compiler limit: compiler stack overflow" problem.
Compiling with Win32 Debug or x64 does not have this problem.
You can easily bypass this problem by changing the optimization options /O1 and /O2 to /Od.
In VS2010, when _INTPTR is 0 or 1, UINTPTR_MAX is 0xFFFFFFFF.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -827,7 +827,15 @@ if (WIN32)
         ${CMAKE_CURRENT_SOURCE_DIR}/msvc/unicorn/qapi-visit.c
     )
     if(CMAKE_SIZEOF_VOID_P EQUAL 8)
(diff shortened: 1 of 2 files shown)


aquynh:
can you confirm that this builds on the latest VS?

dnasdw:
Yes, I am sure it can be compiled in the latest VS.
Because all modifications only affect VS2008 and VS2010.

My VS2019 version is 16.9.5.

aquynh:
merged, thanks!

PR #1414, opened 2021-06-25
Optimize memory_mapping

h33p:
Optimized `memory_mapping` function by ensuring the regions are sorted, which allows to use binary search. This makes insertion in `mem_map` `O(n)`, however, `memory_mapping` function seems to be invoked far more often than the former, and improving it from `O(n)` to `O(log n)` makes it scale much better with lots of single page mappings.

I tested it against Linux kernel's `kallsyms_lookup_name` function and total runtime went down from 1.6 seconds to 1.2.

### DIFF
--- a/uc.c
+++ b/uc.c
@@ -697,25 +697,52 @@ uc_err uc_emu_stop(uc_engine *uc)
     return UC_ERR_OK;
 }
 
+// return target index where a memory region at the address exists, or could be inserted
+//
+// address either is inside the mapping at the returned index, or is in free space before
+// the next mapping.
+//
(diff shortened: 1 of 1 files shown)


wtdcode:
Interesting! Link to #1217

wtdcode:
Also #1449

wtdcode:
Hello, I reviewed your changes and it looks fine to me. Could you resolve the conflict?

h33p:
Hey, I did resolve the conflicts, my master is now based on dev, since gh did not allow to change source branch.

PR #1443, opened 2021-09-13
Support to read ST registers in rust bindings

bet4it:
And the reading of `GDTR` and `IDTR` seems not correct.
Do the size of`uc_x86_mmr` is 10 bytes?


wtdcode:

bet4it:
Why you not provide information???

wtdcode:
Re-open as requested.

bet4it:
If nobody know what's the size of `uc_x86_mmr`, we could just merge this PR. I'm sure what this PR does is correct, and we can deal with `GDTR` and `IDTR` in a future PR.

wtdcode:
> uc_x86_mmr

It's define in x86.h:

```C
typedef struct uc_x86_mmr {
    uint16_t selector;  /* not used by GDTR and IDTR */
    uint64_t base;      /* handle 32 or 64 bit CPUs */
    uint32_t limit;
    uint32_t flags;     /* not used by GDTR and IDTR */
} uc_x86_mmr;
```

bet4it:
Do I need to consider structure alignment? What's the value of `sizeof(struct uc_x86_mmr)`?

wtdcode:
> Do I need to consider structure alignment? What's the value of `sizeof(struct uc_x86_mmr)`?

It's 24 on x86_64 platform, you may add this support next time.

PR #1444, opened 2021-09-26
Rust binding: call cmake in build script

Hakuyume:
To use unicorn via crates.io or git dependency, libunicorn.a should be built in build.rs.


Lani698 (changes requested):
Request

makotoshimazu (approved):
LGTM

wtdcode:
Link to #1217 #1449

aquynh:
please submit pull req to add Rust binding to the "dev" branch of Unicorn2.

Hakuyume:
> please submit pull req to add Rust binding to the "dev" branch of Unicorn2.

OK. I rebased this PR on the dev branch

bet4it:
Could you reuse the library if it's already installed on the system? We can just do a dynamic link on it.

_Originally posted by @bet4it in https://github.com/unicorn-engine/unicorn/issues/1447#issuecomment-932656437_

Hakuyume:
@bet4it 

I updated the build script to check `LIBUNICORN` environment variable.
If the variable is set, the build script uses it instead of building own static library.
You can specify both system-wide and local dynamic library via the variable.

Since I think the behavior of build script should be controlled explicitly, I omitted automatic fallback.

wtdcode:
@Hakuyume If you are still confused, just have a try like this:

```bash
/t/t/build $ cat ../CMakeLists.txt
project(test)

add_library(lib1 STATIC lib1.c)

add_library(lib2 STATIC lib2.c)

target_link_libraries(lib2 lib1)

add_executable(a a.c)
target_link_libraries(a lib2)
/t/t/build $ cmake..
# Some log
/t/t/build $ make VERBOSE=1
# Some logs
[100%] Linking C executable a
/usr/local/Cellar/cmake/3.19.1/bin/cmake -E cmake_link_script CMakeFiles/a.dir/link.txt --verbose=1
/Library/Developer/CommandLineTools/usr/bin/cc  -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX11.3.sdk -Wl,-search_paths_first -Wl,-headerpad_max_install_names CMakeFiles/a.dir/a.c.o -o a  liblib2.a liblib1.a
# Other logs
```

As you can see, the executable a is linked against both `liblib2.a` and `liblib1.a`.

Hakuyume:
> That's not available on Windows and you have to handle that.

Since the build script before my PR does not support windows (as you mentioned in https://github.com/unicorn-engine/unicorn/pull/1444#issuecomment-943486387), Ignoring windows is not degradation.
Can I go without support of such systems? Developing for windows and macOS is hard for me because I don't have windows and macOS environment.
Someone who has windows can add windows support on my PR.

PR #1460, opened 2021-10-13
Fix rust bindings not building

moonshadow565:


wtdcode:
You are doing a dynamic link. AFAIK, there is some rpath problems in rust ecosystem. I would add a rust CI to detect this problem and review this PR again.

moonshadow565:
> You are doing a dynamic link. AFAIK, there is some rpath problems in rust ecosystem. I would add a rust CI to detect this problem and review this PR again.

Yeah i noticed that current rust bindings aren't even passing tests.

I build it dynamically because static cmake builds are broken on dev branch.
cmake *-softmmu targets should be `OBJECT`s

wtdcode:
Fixed in 6d0d0897f8123a5c8a1c8d18e09dcb1e619e5f3f and CI.

PR #1467, opened 2021-10-20
fxsave / fxsave64 should store the floating point instruction pointer (fpip)

dglynos:
fxsave / fxsave64 are sometimes used in GetPC type of code in exploits and exploit encoders. This patch provides support for storing the Floating Point Instruction Pointer (fpip) in fxsave / fxsave64, thus enabling such exploit code to be emulated.


aquynh:
if this really fixes the bug in Unicorn1, we can merge it.

git-census-assess:
@aquynh yes, it fixes the bug; have a look at the unit tests if you like.

aquynh:
merged, thanks!

dglynos:
Thanks!

PR #1469, opened 2021-10-22
RV64 D extension

QDucasse:
Hello, 
I am trying to use RISC-V with its D extension (double precision) and I don't really know how to check or initialize for extensions. I added three tests, two (for rv32 and rv64) that perform a `fmv.d`, move between two floating-point registers, available on both RV32D and RV64D and they pass! 
However, when trying `fmv.d.x` (copy from a general register to a floating-point register) that is only available on RV64D, I get a `UC_ERR_EXCEPTION`.

Is there any way to check for the present extensions? Have I setup the engine incorrectly?

### DIFF
--- a/tests/unit/test_riscv.c
+++ b/tests/unit/test_riscv.c
@@ -203,6 +203,105 @@ static void test_riscv64_3steps_pc_update() {
   OK(uc_close(uc));
 }
 
+static void test_riscv32_fp_move(void) {
+  uc_engine *uc;
+  char code[] = "\xd3\x81\x10\x22"; // fmv.d f3, f1
+
+  uint32_t r_f1 = 0x1234;
+  uint32_t r_f3 = 0x5678;
+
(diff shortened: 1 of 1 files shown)


wtdcode:
Fixed in 09aa0f944f595b9e534ba5ab11c8c024f92ac14d and it's really painful to handcraft riscv assembly.

wtdcode:
Additional note: You have to enable mstatus.fs manually, please see how I modify your tests.

QDucasse:
Thanks a lot for your inputs! I used clang and llvm-objdump to generate the assembly, I don't know if this might help you!

QDucasse:
Hello, as for `mstatus`, I was wondering why this register is not exposed directly through Unicorn the same way `cpacr_el1` is for ARM?

wtdcode:
> Hello, as for `mstatus`, I was wondering why this register is not exposed directly through Unicorn the same way `cpacr_el1` is for ARM?

That's a problem indeed, you may open another issue

PR #1421, opened 2021-07-04
Allow Rust memory hooks to return a bool

alureon:
Unicorn memory hooks return a boolean to indicate if the exception was handled and emulation should continue, or if emulation should halt.

This adds this functionality to the Rust bindings, allowing memory hooks to return a bool.

PR #1480, opened 2021-11-08
Rust bindings improvements

domenukk:
These bindings from unicornafl allow us to
- use a data ptr (if this one is needed is debatable, what do you think?)
- have closure that don't live for`'static` anymore -> easier to capture things in
- have non-boxed closures (perf++, hopefully)
- no_std (potentially smaller)
- no longer need to cast regs to i32 manually

### DIFF
--- a/bindings/rust/src/lib.rs
+++ b/bindings/rust/src/lib.rs
@@ -9,7 +9,7 @@
 //! use unicorn_engine::RegisterARM;
 //! use unicorn_engine::unicorn_const::{Arch, Mode, Permission, SECOND_SCALE};
 //!
-//! fn main() {
+//! fn emulate() {
 //!     let arm_code32: Vec<u8> = vec![0x17, 0x00, 0x40, 0xe2]; // sub r0, #23
 //!
 //!     let mut unicorn = unicorn_engine::Unicorn::new(Arch::ARM, Mode::LITTLE_ENDIAN).expect("failed to initialize Unicorn instance");
@@ -35,17 +35,15 @@ mod arm64;
 mod m68k;
 mod mips;
 mod ppc;
-mod riscv;
 mod sparc;
 mod x86;
-pub use crate::{arm::*, arm64::*, m68k::*, mips::*, ppc::*, riscv::*, sparc::*, x86::*};
+use std::{marker::PhantomData, ptr};
+
+pub use crate::{arm::*, arm64::*, m68k::*, mips::*, ppc::*, sparc::*, x86::*};
 
 use ffi::uc_handle;
-use std::collections::HashMap;
-use std::ffi::c_void;
-use std::marker::PhantomPinned;
-use std::pin::Pin;
-use unicorn_const::*;
+use libc::c_void;
+use unicorn_const::{uc_error, Arch, HookType, MemRegion, MemType, Mode, Permission, Query};
 
 #[derive(Debug)]
 pub struct Context {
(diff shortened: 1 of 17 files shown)


bindings/rust/tests/unicorn.rs
```diff
@@ -133,15 +153,12 @@ pub static X86_REGISTERS: [RegisterX86; 125] = [
...
 ];
 
-type Unicorn<'a> = unicorn_engine::UnicornHandle<'a>;
-
 #[test]
 fn emulate_x86() {
     let x86_code32: Vec<u8> = vec![0x41, 0x4a]; // INC ecx; DEC edx
 
-    let mut unicorn = unicorn_engine::Unicorn::new(Arch::X86, Mode::MODE_32)
+    let mut emu = unicorn_emulator::Unicorn::new(Arch::X86, Mode::MODE_32, 0)
```
wtdcode, line 160: You are failing CI due to the different crate names.

uc.c
```diff
@@ -686,6 +686,27 @@ static void clear_deleted_hooks(uc_engine *uc)
     list_clear(&uc->hooks_to_del);
 }
 
+// set a data ptr (for use in bindings)
+UNICORN_EXPORT
+uc_err uc_emu_set_data_ptr(uc_engine *uc, void *data)
```
wtdcode, line 691: Do we really need this API? I mean, why don't you save this data in the rust bindings?

domenukk: I need to know the pointer to the object somehow in the callbacks, if I store it earlier it may have been moved.
Right now it's solved in a way where the callback gets a different type than the surrounding object which also seems wrong. I can think about it again

bindings/rust/src/lib.rs
```diff
@@ -306,7 +319,7 @@ impl<'a> UnicornHandle<'a> {
 
     /// Read 128, 256 or 512 bit register value into heap allocated byte array.
     ///
-    /// This adds safe support for registers >64 bit (GDTR/IDTR, XMM, YMM, ZMM, ST (x86); Q, V (arm64)).
+    /// This adds safe support for registers >64 bit (GDTR/IDTR, XMM, YMM, ZMM (x86); Q, V (arm64)).
```
domenukk, line 322: Didn't mean to change this part

bet4it: Why you revert some of my changes...🙃

domenukk: Sorry didn't mean to :)

domenukk:
@wtdcode got rid of the additional API

wtdcode:
Merged, thanks. ;)

domenukk:
@wtdcode one thing, in case you autogenerate the register ids, make sure you add

```rust
impl From<RegisterARM> for i32 {
    fn from(r: RegisterARM) -> Self {
        r as i32
    }
}
```
to each file :)

wtdcode:
> @wtdcode one thing, in case you autogenerate the register ids, make sure you add
> 
> ```rust
> impl From<RegisterARM> for i32 {
>     fn from(r: RegisterARM) -> Self {
>         r as i32
>     }
> }
> ```
> 
> to each file :)

Rust bindings constants are not generated automatically unfortunately and that's on my TODO plan.

bet4it:
There are two types of mem callback:
https://github.com/unicorn-engine/unicorn/blob/52f90cda023abaca510d59f021c88629270ad6c0/include/unicorn/unicorn.h#L283-L319
`uc_cb_eventmem_t` has a return value but `uc_cb_hookmem_t` doesn't.
You may need to add a new `add_mem_invalid_hook` function.


`uc_cb_hookinsn_invalid_t` and `uc_cb_insn_in_t` also have a return value:
https://github.com/unicorn-engine/unicorn/blob/52f90cda023abaca510d59f021c88629270ad6c0/include/unicorn/unicorn.h#L190-L206
You may change them in this PR.

_Originally posted by @bet4it in https://github.com/unicorn-engine/unicorn/issues/1421#issuecomment-875193524_


`mem_hook_proxy` shouldn't always return a bool.

domenukk:
> > @wtdcode one thing, in case you autogenerate the register ids, make sure you add
> > ```rust
> > impl From<RegisterARM> for i32 {
> >     fn from(r: RegisterARM) -> Self {
> >         r as i32
> >     }
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > to each file :)
> 
> Rust bindings constants are not generated automatically unfortunately and that's on my TODO plan.

I had it running at one point, but probably needs some adaption to newer changes, take a look at:
https://github.com/AFLplusplus/unicornafl/blob/6c9ed25bfb8ff8650921fc27be74ba4e6b7282a4/bindings/const_generator.py

domenukk:
> There are two types of mem callback:
> 
> https://github.com/unicorn-engine/unicorn/blob/52f90cda023abaca510d59f021c88629270ad6c0/include/unicorn/unicorn.h#L283-L319
> 
> `uc_cb_eventmem_t` has a return value but `uc_cb_hookmem_t` doesn't.
> You may need to add a new `add_mem_invalid_hook` function.
> 
> `uc_cb_hookinsn_invalid_t` and `uc_cb_insn_in_t` also have a return value:
> 
> https://github.com/unicorn-engine/unicorn/blob/52f90cda023abaca510d59f021c88629270ad6c0/include/unicorn/unicorn.h#L190-L206
> 
> You may change them in this PR.
> 
> _Originally posted by @bet4it in [#1421 (comment)](https://github.com/unicorn-engine/unicorn/pull/1421#issuecomment-875193524)_
> 
> `mem_hook_proxy` shouldn't always return a bool.

Ah, missed that. Will take a look at it later and open another pr (unless someone else wants to)

domenukk:
@bet4it what do you recommend? Drop the bool return again? Add a hook?

bet4it:
Add `add_mem_invalid_hook` and `add_insn_in_invalid_hook` function.

PR #1481, opened 2021-11-09
Restore cp15 registers

bet4it:


wtdcode:
What are these registers used for? Looks good to me in general but I hope some unit tests about the use cases should be added.

bet4it:
I'm not sure about others, I only know `tpidr_el0` relates with thread operations and I can only add tests for it😅

wtdcode:
> I'm not sure about others, I only know `tpidr_el0` relates with thread operations and I can only add tests for it😅

Ah I see. Never mind, it's pretty fine.

PR #1483, opened 2021-11-09
No redownload in rust bindings

domenukk:
With uc2, the bindings no longer use the surrounding unicorn checkout, but download the unicorn instance.
Sadly, this happens with every build, so changing a single line in your script takes 5 minutes to build...
This PR changes this to only download once.

### DIFF
--- a/bindings/rust/build.rs
+++ b/bindings/rust/build.rs
@@ -1,11 +1,11 @@
 use bytes::Buf;
 use flate2::read::GzDecoder;
 use reqwest::header::USER_AGENT;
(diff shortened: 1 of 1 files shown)


wtdcode:
Your version seems old and you may check this commit ec02eccb85bc44b7547c7a3422b668cd27af8b1d

domenukk:
Now it should look better :)
Maybe rename the `CI` env to `NO_DOWNLOAD` or something? It's also useful during bindings and uc development

wtdcode:
Merged, thanks!

PR #1486, opened 2021-11-13
Accomodate for setting a different CMAKE_GENERATOR in the environment

mid-kid:


bindings/python/setup.py
```diff
@@ -134,7 +134,7 @@ def build_libraries():
         subprocess.check_call(["cmake", '-B', BUILD_DIR, "-DCMAKE_BUILD_TYPE=" + conf])
         os.chdir(BUILD_DIR)
         threads = os.getenv("THREADS", "4")
-        subprocess.check_call(["make", "-j" + threads])
+        subprocess.check_call(["cmake", "--build", ".", "-j" + threads])
```
wtdcode, line 137: Does `cmake --build` support `-j`?

wtdcode: Okay seems so.

wtdcode:
Merged, thanks!

PR #1504, opened 2021-11-30
rust: Add RISCV CSR registers

Kritzefitz:
The addition of these registers in the C base caused the rust values
for all floating point registers and the PC to point to some of the
CSR registers instead.


bindings/rust/src/riscv.rs
```diff
@@ -41,41 +41,168 @@ pub enum RegisterRISCV {
...
-    F23 = 56,
-    F24 = 57,
-    F25 = 58,
-    F26 = 59,
-    F27 = 60,
-    F28 = 61,
-    F29 = 62,
-    F30 = 63,
-    F31 = 64,
-    PC = 65,
```
wtdcode, line 77: Why not add the variables after `PC=65`?

Kritzefitz: The conversion of `RegisterRISCV` in rust to `uc_riscv_reg` in C goes via an intermediate `int`. This means the conversion of `PC` should looks something like this:

```
RegisterRISCV::PC -> 190 -> UC_RISCV_REG_PC
```

However, this assumes that the mapping between `RegisterRISCV` and ints int rust (which is defined explicitly) matches that between `uc_riscv_reg` and ints in C (which is implicitly defined by the order of the enum). When the CSR registers were added to `uc_riscv_reg` the mapping in C became out of sync with the mapping in rust, so the conversion of `PC` from rust to C currently looks something like this:

```
RegisterRISCV::PC -> 65 -> UC_RISCV_REG_HPMCOUNTER21
```

As you see, the C side thinks we want to read HPMCOUNTER21, when we actually want to read the PC. This is obviously not good.

To fix this, we need to change the mappings in rust for all registers that have been shifted by the addition of the CSR registers in the C code.

As a side note: keeping the mappings between registers in C and rust synced by hand seems pretty error prone to me. Unfortunately I don't know a better alternative, because rust's FFI to my knowledge doesn't provide a mechanism to access definitions of enums in C.

wtdcode: Sorry I realize it's my mistake [here](https://github.com/unicorn-engine/unicorn/commit/c6fdbb373506adc3c725c4e97f93c6953e53aca3). The variables are in the right place.

wtdcode:
Merged, thanks.

Issue #1502, opened 2021-11-26
Unicorn 2 fails to emulate multiple M68K instructions


zznop:
Hit CPU unhandled exception with these M68K instructions (using 2.0.0-rc4):

```
0641 0100                                         addi.w  #$100,d1
```

```
46FC 2700                                         move    #$2700,sr
```

```
51C8 FFF2                                         dbf     d0,loc_220
```

```
0A47 0000                                         eori.w  #0,d7
```

wtdcode:
Thanks for your report and I would have a look ASAP.

zznop:
Did some digging into the C source code. I found that in `qemu/target/m68k/cpu.c` that there are different instructions enabled for different M68K CPU models. By setting `uc->cpu_model` to `0`, it sets the CPU to M68000 and it looks like it emulates most of the instructions mentioned above correctly. The only issue is with the `move` instruction into `sr`. I haven't traced down the code responsible for this yet, but I plan to dig deeper when I have more time. For now, here's the patch I applied to force it to set the CPU to M68000:

```
diff --git a/qemu/target/m68k/unicorn.c b/qemu/target/m68k/unicorn.c
index 3930bc56..8709ba5c 100644
--- a/qemu/target/m68k/unicorn.c
+++ b/qemu/target/m68k/unicorn.c
@@ -151,6 +151,7 @@ static int m68k_cpus_init(struct uc_struct *uc, const char *cpu_model)
 {
     M68kCPU *cpu;
 
+    uc->cpu_model = 0;
     cpu = cpu_m68k_init(uc);
     if (cpu == NULL) {
         return -1;
```

Obviously, this isn't a solution. I think we want the go bindings (and all other bindings) to expose the `cpu_model` in the unicorn interface/struct so we can set the processor model

aquynh:
obviously, you are one of few people who seriously play with m68k.

do you have a similar issue with Unicorn 1?

wtdcode:
> @wtdcode Did some digging into the C source code. I found that in `qemu/target/m68k/cpu.c` that there are different instructions enabled for different M68K CPU models. By setting `uc->cpu_model` to `0`, it sets the CPU to M68000 and it looks like it emulates most of the instructions mentioned above correctly. The only issue is with the `move` instruction into `sr`. I haven't traced down the code responsible for this yet, but I plan to dig deeper when I have more time. For now, here's the patch I applied to force it to set the CPU to M68000:
> 
> ```
[...]

PR #1506, opened 2021-12-02
SetCPUModel go binding for setting the CPU model

zznop:
I added a `SetCPUModel` method to the unicorn interface in the go bindings to address #1502. CGO does not support calling C functions with varargs so I also added a `uc_ctl_set_cpu_model_helper` function to `uc.c` to call `uc_ctl`

Usage:

`
emu, _ := uc.NewUnicorn(uc.ARCH_M68K, uc.MODE_BIG_ENDIAN)
emu.SetCPUModel(uc.CPU_M68000_CPU)
`

### DIFF
--- a/bindings/go/unicorn/uc.c
+++ b/bindings/go/unicorn/uc.c
@@ -23,3 +23,7 @@ uc_err uc_reg_write_batch_helper(uc_engine *handle, int *regs, uint64_t *val_in,
     free(val_ref);
     return ret;
 }
+
+uc_err uc_ctl_set_cpu_model_helper(uc_engine *handle, int model) {
+    return uc_ctl_set_cpu_model(handle, model);
+}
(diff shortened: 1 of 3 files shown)


bindings/go/unicorn/uc.c
```diff
@@ -23,3 +23,7 @@ uc_err uc_reg_write_batch_helper(uc_engine *handle, int *regs, uint64_t *val_in,
     free(val_ref);
     return ret;
 }
+
+void uc_ctl_set_cpu_model_helper(uc_engine *handle, int model) {
+    uc_ctl(handle, UC_CTL_CPU_MODEL, model);
```
wtdcode, line 28: This is wrong, you have to use `uc_ctl_set_cpu_model` macro.

bindings/go/unicorn/uc.h
```diff
@@ -1,2 +1,3 @@
 uc_err uc_reg_read_batch_helper(uc_engine *handle, int *regs, uint64_t *val_out, int count);
 uc_err uc_reg_write_batch_helper(uc_engine *handle, int *regs, uint64_t *val_in, int count);
+void uc_ctl_set_cpu_model_helper(uc_engine *handle, int model);
```
wtdcode, line 3: This is also wrong, the type of the return value is uc_err.

zznop: Fixed

bindings/go/unicorn/uc.c
zznop: Fixed

wtdcode:
Merged, thanks!

PR #1509, opened 2021-12-06
Bug fix for LUI instruction (MIPS)

FedorNiskov:
This little patch fixes a bug in the QEMU of Unicorn - implementation of the LUI instruction (MIPS) is not correct: the instruction must perform signed extension of the immediate operand, but it is unsigned in the code (probably a misprint); I have fixed it. I also attach the corresponding fragment of the MIPS instruction set reference which describes this instruction.

lui_mips

### DIFF
--- a/qemu/target-mips/translate.c
+++ b/qemu/target-mips/translate.c
@@ -2551,7 +2551,7 @@ static void gen_logic_imm(DisasContext *ctx, uint32_t opc,
             tcg_gen_ext32s_tl(tcg_ctx, *cpu_gpr[rt], *cpu_gpr[rt]);
             MIPS_DEBUG("aui %s, %s, %04x", regnames[rt], regnames[rs], imm);
         } else {
-            tcg_gen_movi_tl(tcg_ctx, *cpu_gpr[rt], uimm << 16);
+            tcg_gen_movi_tl(tcg_ctx, *cpu_gpr[rt], imm << 16);
             MIPS_DEBUG("lui %s, " TARGET_FMT_lx, regnames[rt], uimm);
         }
         break;


qemu/target-mips/translate.c
```diff
@@ -2551,7 +2551,7 @@ static void gen_logic_imm(DisasContext *ctx, uint32_t opc,
             tcg_gen_ext32s_tl(tcg_ctx, *cpu_gpr[rt], *cpu_gpr[rt]);
             MIPS_DEBUG("aui %s, %s, %04x", regnames[rt], regnames[rs], imm);
         } else {
-            tcg_gen_movi_tl(tcg_ctx, *cpu_gpr[rt], uimm << 16);
+            tcg_gen_movi_tl(tcg_ctx, *cpu_gpr[rt], imm << 16);
```
wtdcode, line 2554: As previous `(rs != 0 && (ctx->insn_flags & ISA_MIPS32R6))` infers, only `MIPS32R6` instruction set requires sign extension and the logic here implements it correcly.

FedorNiskov:
> please confirm this issue in Unicorn2, in the "dev" branch.

@aquynh , it seems that Unicorn 2 uses more modern version of QEMU where this bug has already been fixed; however, while Unicorn 2 is being developed and many people are still using Unicorn 1, maybe we should fix it here too.

wtdcode:
> > please confirm this issue in Unicorn2, in the "dev" branch.
> 
> @aquynh , it seems that Unicorn 2 uses more modern version of QEMU where this bug has already been fixed; however, while Unicorn 2 is being developed and many people are still using Unicorn 1, maybe we should fix it here too.

https://github.com/qemu/qemu/blob/2f8eb086732ad1875003101f5324f01c47d7408c/target/mips/tcg/translate.c#L2502

Even the latest QEMU is using exactly the same code. The difference you are observing is probably that we changed the default CPU model for MIPS emulation (and thus the actual instruction set).

FedorNiskov:
> > > please confirm this issue in Unicorn2, in the "dev" branch.
> > 
> > 
> > @aquynh , it seems that Unicorn 2 uses more modern version of QEMU where this bug has already been fixed; however, while Unicorn 2 is being developed and many people are still using Unicorn 1, maybe we should fix it here too.
> 
> https://github.com/qemu/qemu/blob/2f8eb086732ad1875003101f5324f01c47d7408c/target/mips/tcg/translate.c#L2502
> 
> Even the latest QEMU is using exactly the same code. The difference you are observing is probably that we changed the default CPU model for MIPS emulation (and thus the actual instruction set).

@wtdcode , I have also checked the new QEMU, and it uses 'imm' instead of 'uimm', so I suggest to do the same fix in our QEMU - to replace 'uimm' -> 'imm'

qemu/target-mips/translate.c
FedorNiskov: @wtdcode , this instruction should always perform sign extension - you can check it in the modern QEMU: it uses 'imm', not 'uimm', in these lines; the problem of 'uimm' is that 'uimm << 16' is an unsigned 32-bit value, and it is cast (unsigned-extended) to 64-bit, but this is incorrect and causes errors during emulation.

wtdcode: Sorry I did a double-check and you are right indeed.

wtdcode:
> > > > please confirm this issue in Unicorn2, in the "dev" branch.
> > > 
> > > 
> > > @aquynh , it seems that Unicorn 2 uses more modern version of QEMU where this bug has already been fixed; however, while Unicorn 2 is being developed and many people are still using Unicorn 1, maybe we should fix it here too.
> > 
> > 
> > https://github.com/qemu/qemu/blob/2f8eb086732ad1875003101f5324f01c47d7408c/target/mips/tcg/translate.c#L2502
> > Even the latest QEMU is using exactly the same code. The difference you are observing is probably that we changed the default CPU model for MIPS emulation (and thus the actual instruction set).
> 
> @wtdcode , I have also checked the new QEMU, and it uses 'imm' instead of 'uimm', so I suggest to do the same fix in our QEMU - to replace 'uimm' -> 'imm'

Yes, sorry I was wrong.

wtdcode:
For this PR, I suggest you upgrading to unicorn2 since there is no more support & release for Unicorn1.

@aquynh I could confirm it's an old QEMU bug that is already fixed in Unicorn2. Maybe we can merge it?

aquynh:
Yes we accept bugfixes for Unicorn1, for that reason: people still use it.

FedorNiskov:
@aquynh @wtdcode Thank you!

PR #1511, opened 2021-12-10
Add use_system_unicorn feature in rust bindings

bet4it:
Add a new `use_system_unicorn` feature to support link unicorn rust bindings against the unicorn library installed on the computer.

Some code are borrowed from https://github.com/keystone-engine/keystone/blob/master/bindings/rust/keystone-sys/build.rs.


bet4it:
@Hakuyume What do you think of it?

wtdcode:
Merged, thanks!.

Sorry for the delay.

PR #1512, opened 2021-12-10
Issues with count for code patching 

QDucasse:
Hello, when rewriting (what I called code patching) the code memory to change an instruction, the previous instruction will be executed instead. 
In the tests, I wrote an `add 1` to a register, emulated to see the result, then overwrote the instruction with `add 7ff` to the same register and re-emulated. When using the `start_address`/`end_address` method, the execution of the new instruction proceeds however when using a count, the previous instruction is executed instead.

### DIFF
--- a/tests/unit/test_arm64.c
+++ b/tests/unit/test_arm64.c
@@ -50,4 +50,64 @@ static void test_arm64_until()
     OK(uc_close(uc));
 }
 
-TEST_LIST = {{"test_arm64_until", test_arm64_until}, {NULL, NULL}};
+
+static void test_arm64_code_patching() {
+    uc_engine *uc;
+    char code[] = "\x00\x04\x00\x11"; // add w0, w0, 0x1
+    uc_common_setup(&uc, UC_ARCH_ARM64, UC_MODE_ARM, code, sizeof(code) - 1);
+    // zero out x0
+    uint64_t r_x0 = 0x0;
+    OK(uc_reg_write(uc, UC_ARM64_REG_X0, &r_x0));
+    // emulate the instruction
+    OK(uc_emu_start(uc, code_start, code_start + sizeof(code) -1, 0, 0));
+    // check value
+    OK(uc_reg_read(uc, UC_ARM64_REG_X0, &r_x0));
+    TEST_CHECK(r_x0 == 0x1);
+    // patch instruction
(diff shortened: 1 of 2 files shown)


wtdcode:
I believe it should be addressed by this unit test: https://github.com/unicorn-engine/unicorn/blob/1923c123150bc2ec2cd259658fe7d730287ef9de/tests/unit/test_x86.c#L633

Really weird but a workaround is `uc_ctl_remove_cache` on the instruction before emulation.

guillep:
> I believe it should be addressed by this unit test:
> 
> https://github.com/unicorn-engine/unicorn/blob/1923c123150bc2ec2cd259658fe7d730287ef9de/tests/unit/test_x86.c#L633
> 
> Really weird but a workaround is `uc_ctl_remove_cache` on the instruction before emulation.

Do you mean that we need to flush some cache using `uc_ctl_remove_cache` after patching code? I think that's a good solution. I don't expect unicorn to realize that we are overwriting code...

wtdcode:
> > I believe it should be addressed by this unit test:
> > https://github.com/unicorn-engine/unicorn/blob/1923c123150bc2ec2cd259658fe7d730287ef9de/tests/unit/test_x86.c#L633
> > 
> > Really weird but a workaround is `uc_ctl_remove_cache` on the instruction before emulation.
> 
> Do you mean that we need to flush some cache using `uc_ctl_remove_cache` after patching code? I think that's a good solution. I don't expect unicorn to realize that we are overwriting code...

Not exactly. Unicorn tracks all code page and when you are overwriting code, the cache should be flushed accordingly.

wtdcode:
Okay, I confirm it's a bug. The code cache is not flushed by `uc_mem_write`.

wtdcode:
After some investigation, I'm aware that it's not a bug. It's somewhat subtle for this case. Internally, `uc_mem_write` would treat the address as the physical address while the translation cached is based on the virtual address. Unfortunately, the conversion from virtual address to physical address is straightforward and easy while the vice versa is almost impossible (consider: 1. many virtual addresses are mapped to the same physical address 2. some virtual address is not backed by virtual address). Thus, @guillep is correct: We don't know whether the user is overwriting some code or data during `uc_mem_write` and therefore it's users' responsibility to flush cache when they are patching some code.

Btw, SMC code, i.e. the code that patching themselves during emulation works fine and doesn't need extra care.

wtdcode:
Merged, thanks.

PR #1528, opened 2022-01-08
python: look for libunicorn.so.1 rather than libunicorn.so

flyn-org:
Many linux distributions place the ".so" form of their shared libraries
in an optional development package, preferring the ".so.N" form. Modify
the Python package to look for the latter rather than the former.

Signed-off-by: W. Michael Petullo <mike@flyn.org>

I maintain the Fedora unicorn package. I am not sure how this fits into the larger Unicorn project, but it seemed to make sense for Fedora.

### DIFF
--- a/bindings/python/unicorn/unicorn.py
+++ b/bindings/python/unicorn/unicorn.py
@@ -23,7 +23,7 @@ if _python2:
 _lib = { 'darwin': 'libunicorn.dylib',
          'win32': 'unicorn.dll',
          'cygwin': 'cygunicorn.dll',
-         'linux': 'libunicorn.so',
+         'linux': 'libunicorn.so.1',
          'linux2': 'libunicorn.so' }
 
 


wtdcode:
I'm afraid the change here may break the python bindings on Linux totally.

btw, CI fails for now and I'm fixing it.

flyn-org:
Most applications look for the libX.so.N form, because N represents the interface version (https://tldp.org/HOWTO/Program-Library-HOWTO/shared-libraries.html). I guess I don't understand why unicorn does otherwise.

Here is an example of a C program that I compiled against unicorn using GCC:
```
$ ldd /tmp/foo
	linux-vdso.so.1 (0x00007ffc5a5a3000)
	libunicorn.so.1 => /lib64/libunicorn.so.1 (0x00007f82cc7e0000)
	libc.so.6 => /lib64/libc.so.6 (0x00007f82cc5d6000)
	libm.so.6 => /lib64/libm.so.6 (0x00007f82cc4fa000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f82cd289000)
```

wtdcode:
> Most applications look for the libX.so.N form, because N represents the interface version (https://tldp.org/HOWTO/Program-Library-HOWTO/shared-libraries.html). I guess I don't understand why unicorn does otherwise.
> 
> Here is an example of a C program that I compiled against unicorn using GCC:
> 
> ```
> $ ldd /tmp/foo
> 	linux-vdso.so.1 (0x00007ffc5a5a3000)
> 	libunicorn.so.1 => /lib64/libunicorn.so.1 (0x00007f82cc7e0000)
> 	libc.so.6 => /lib64/libc.so.6 (0x00007f82cc5d6000)
> 	libm.so.6 => /lib64/libm.so.6 (0x00007f82cc4fa000)
> 	/lib64/ld-linux-x86-64.so.2 (0x00007f82cd289000)
> ```

Because `libunicorn.so` usually is a link to `libunicorn.so.1` or `libunicorn.so.2`. That shall work fine, right?

flyn-org:
> Because `libunicorn.so` usually is a link to `libunicorn.so.1` or `libunicorn.so.2`. That shall work fine, right?

Yes, but if you rely on libunicorn.so, then you do not know which interface version you will get, between .so.N, .so.M, and so on. Thus using the .so link will break applications who rely on libraries that follow the normal versioning conventions. The .so link is for build tools that create programs that then rely on the proper .so.N, depending on the build environment installed on the system. Almost any application I look at with `ldd` behaves this way.

As I pointed out earlier, the .so links are normally part of a development package (i.e., `-devel` on Fedora or `-dev` on Ubuntu), and development packages should not be a runtime requirement in most cases.

flyn-org:
> Because `libunicorn.so` usually is a link to `libunicorn.so.1` or `libunicorn.so.2`. That shall work fine, right?

Yes, but if you rely on libunicorn.so, then you do not know which interface version you will get, between .so.N, .so.M, and so on. Thus using the .so link will break applications who rely on libraries that follow the normal versioning conventions. The .so link is for build tools that create programs that then rely on the proper .so.N, depending on the build environment installed on the system. Almost any application I look at with `ldd` behaves this way.

As I pointed out earlier, the .so links are normally part of a development package (i.e., `-devel` on Fedora or `-dev` on Ubuntu), and development packages should not be a runtime requirement in most cases.

wtdcode:
> > Because `libunicorn.so` usually is a link to `libunicorn.so.1` or `libunicorn.so.2`. That shall work fine, right?
> 
> Yes, but if you rely on libunicorn.so, then you do not know which interface version you will get, between .so.N, .so.M, and so on. Thus using the .so link will break applications who rely on libraries that follow the normal versioning conventions. The .so link is for build tools that create programs that then rely on the proper .so.N, depending on the build environment installed on the system. Almost any application I look at with `ldd` behaves this way.
> 
> As I pointed out earlier, the .so links are normally part of a development package (i.e., `-devel` on Fedora or `-dev` on Ubuntu), and development packages should not be a runtime requirement in most cases.

But wait, we are talking about python bindings. Why do the python bindings care about the `SONAME`? It would check the actual unicorn version by `uc_version` API anyway.

For development, I believe the installed libraries (built by cmake) contain both `libunicorn.so` and `libunicorn.so.2`.

flyn-org:
> But wait, we are talking about python bindings. Why do the python bindings care about the `SONAME`? It would check the actual unicorn version by `uc_version` API anyway.
> 
> For development, I believe the installed libraries (built by cmake) contain both `libunicorn.so` and `libunicorn.so.2`.

We are indeed talking about Python, but I am arguing we should treat Python like C with respect to the library filename it loads. Doing otherwise arrives at the same problem of requiring development packages to run a program. I do need to look a little closer at comparable Python modules---but, it looks like numpy relies on libflexiblas.so.3, for example.

wtdcode:
> > But wait, we are talking about python bindings. Why do the python bindings care about the `SONAME`? It would check the actual unicorn version by `uc_version` API anyway.
> > For development, I believe the installed libraries (built by cmake) contain both `libunicorn.so` and `libunicorn.so.2`.
> 
> We are indeed talking about Python, but I am arguing we should treat Python like C with respect to the library filename it loads. Doing otherwise arrives at the same problem of requiring development packages to run a program. I do need to look a little closer at comparable Python modules---but, it looks like numpy relies on libflexiblas.so.3, for example.

Not necessarily the case. Take pytorch as an example:

https://github.com/pytorch/pytorch/blob/785b6905de121179dda5ae2f4356b6a705693c57/torch/__init__.py#L147

Basically, we don't expect `ctypes.CDLL` to do versioning work, which is done later by `uc_version` API and thus `libunicorn.so` works best to find existing libraries.

wtdcode:
closed due to inactive.

PR #1537, opened 2022-01-15
Clear Python pending hook exception before we enter the emulation.

gerph:
The pending exception hook is set when the hook raised an exception
and wants to report it outside the emulation loop. However, it is
never cleared back to None. This means that after an exception is
raised in a hook, all subsequent execution (even if successful) will
raise the exception.

This change clears the exception before we start another emulation,
which should ensure that if we have _hook_exception set, it really
is from hooks in this emulation run.


wtdcode:
Merged, thanks!

Issue #1508, opened 2021-12-04
static build for 2.0.0-rc5


xx24678:
How to do static build? I did the following:
```
UNICORN_VERSION=2.0.0-rc5
mkdir build && cd build
cmake \
    -DCMAKE_INSTALL_PREFIX="../../$UNICORN_VERSION" \
    -DUNICORN_BUILD_SHARED=OFF \
    -DCMAKE_BUILD_TYPE=Release \
    ..

make install -j8
```
This command ended up with only a 40KB libunicorn.a file in 2.0.0-rc5/lib directory, which only contains `cpu.c.o  uc.c.o  vl.c.o`. Am I doing something wrong here?

wtdcode:
That’s intended. You have to link against all static libraries including libsoftmmu-* and libunicorn-common.

xx24678:
any idea how to generate all those static libraries? My build with the above command only generates libunicorn.a.

wtdcode:
> any idea how to generate all those static libraries? My build with the above command only generates libunicorn.a.

Those static libraries are also generated in the same directory.

xx24678:
I see. They are generated in the build directory, but not copied to install direction.

wtdcode:
> I see. They are generated in the build directory, but not copied to install direction.

Sorry I didn't see you are doing an installation.

Yes, you are right. It's due to cmake limitation and we don't have a cross-platform solution to merge static libraries (at least I don't know).

wtdcode:
Fixed in 82559cb9a75c1bf84b04ba740cbf07641859aed4

Issue #1490, opened 2021-11-18
unicorn 2.0.0 rc4: fails to install shared libraries 


blshkv:
Hi, I'm trying to compile unicorn for 1 target only ```x86```, but it fails similarly with all targets enabled by default:
```
   /usr/lib64/libunicorn.so.2
 * QA Notice: Unresolved soname dependencies:
 * 
 *      /usr/lib64/libunicorn.so.2: libmips-softmmu.so libmips64-softmmu.so libmips64el-softmmu.so libmipsel-softmmu.so libunicorn-common.so libx86_64-softmmu.so

```

here is output of configuration stage:
```
 * Source directory (CMAKE_USE_DIR): "/var/tmp/portage/dev-util/unicorn-2.0.0_rc4/work/unicorn-2.0.0-rc4"
 * Build directory  (BUILD_DIR):     "/var/tmp/portage/dev-util/unicorn-2.0.0_rc4/work/unicorn-2.0.0_rc4_build"
>>> Source prepared.
>>> Configuring source in /var/tmp/portage/dev-util/unicorn-2.0.0_rc4/work/unicorn-2.0.0-rc4 ...
 * Source directory (CMAKE_USE_DIR): "/var/tmp/portage/dev-util/unicorn-2.0.0_rc4/work/unicorn-2.0.0-rc4"
 * Build directory  (BUILD_DIR):     "/var/tmp/portage/dev-util/unicorn-2.0.0_rc4/work/unicorn-2.0.0_rc4_build"
cmake -C /var/tmp/portage/dev-util/unicorn-2.0.0_rc4/work/unicorn-2.0.0_rc4_build/gentoo_common_config.cmake -G Ninja -DCMAKE_INSTALL_PREFIX=/usr -DUNICORN_BUILD_SHARED=ON -DUNICORN_ARCH=x86  -DCMAKE_BUILD_TYPE=Gentoo -DCMAKE_TOOLCHAIN_FILE=/var/tmp/portage/dev-util/unicorn-2.0.0_rc4/work/unicorn-2.0.0_rc4_build/gentoo_toolchain.cmake  /var/tmp/portage/dev-util/unicorn-2.0.0_rc4/work/unicorn-2.0.0-rc4
loading initial cache file /var/tmp/portage/dev-util/unicorn-2.0.0_rc4/work/unicorn-2.0.0_rc4_build/gentoo_common_config.cmake
-- The C compiler identification is GNU 11.2.0
[...]

PR #1544, opened 2022-01-18
Refactor CMake

mrexodia:
Everything from #1373 seems to have been undone, so I did it again


wtdcode:
Strange, I remember I cherry-picked it.

.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "docs/Unicorn_Engine_Documentation"]
 	path = docs/Unicorn_Engine_Documentation
-	url = git@github.com:kabeor/Unicorn-Engine-Documentation
+	url = ../kabeor/Unicorn-Engine-Documentation
```
wtdcode, line 3: There one seems irrelevant and I will fix it.

mrexodia: It's relevant because when cloning the project as https it errors

mrexodia: See https://github.com/mrexodia/unicorn_template/runs/4858696420?check_suite_focus=true#step:3:30

mrexodia: I removed the change.

mrexodia:
I think I also fixed the rust build.

CMakeLists.txt
```diff
@@ -4,30 +4,36 @@
...
 set(UNICORN_VERSION_PATCH 0)
 
-option(UNICORN_BUILD_SHARED "Build shared instead of static library" ON)
-
-if (NOT UNICORN_ARCH)
-    # build all architectures
-    set(UNICORN_ARCH "x86 arm aarch64 riscv mips sparc m68k ppc")
-endif()
-
+option(BUILD_SHARED_LIBS "Build shared instead of static library" ${PROJECT_IS_TOP_LEVEL})
```
wtdcode, line 26: I'm afraid `BUILD_SHARED_LIBS` will break something. See #1490

CMakeLists.txt
```diff
@@ -4,30 +4,36 @@
 cmake_minimum_required(VERSION 3.1)
 
 # Workaround to fix wrong compiler on macos.
-if ((APPLE) AND (NOT CMAKE_C_COMPILER))
+if(APPLE AND NOT CMAKE_C_COMPILER)
     set(CMAKE_C_COMPILER "/usr/bin/cc")
 endif()
+
+# Detect if unicorn is compiled as the top-level project
+set(PROJECT_IS_TOP_LEVEL OFF)
```
wtdcode, line 12: Why does this matter?

wtdcode: Does `STATIC` below override this?

CMakeLists.txt
wtdcode: Does STATIC below override this?

CMakeLists.txt
```diff
@@ -375,8 +387,8 @@ set(UNICORN_ARCH_COMMON
     qemu/accel/tcg/translator.c
 )
 
-if (UNICORN_HAS_X86)
-add_library(x86_64-softmmu
+if(UNICORN_HAS_X86)
+add_library(x86_64-softmmu STATIC
```
wtdcode, line 391: Does `STATIC` here override `BUILD_SHARED_LIBS`?

CMakeLists.txt
```diff
@@ -1097,91 +1109,87 @@ else()
...
-if (UNICORN_HAS_MIPS)
+if(UNICORN_HAS_MIPS)
     set(UNICORN_COMPILE_OPTIONS ${UNICORN_COMPILE_OPTIONS} -DUNICORN_HAS_MIPS -DUNICORN_HAS_MIPSEL -DUNICORN_HAS_MIPS64 -DUNICORN_HAS_MIPS64EL)
     set(UNICORN_LINK_LIBRARIES ${UNICORN_LINK_LIBRARIES} mips-softmmu mipsel-softmmu mips64-softmmu mips64el-softmmu)
     set(UNICORN_SAMPLE_FILE ${UNICORN_SAMPLE_FILE} sample_mips)
-    target_link_libraries(mips-softmmu unicorn-common)
-    target_link_libraries(mipsel-softmmu unicorn-common)
-    target_link_libraries(mips64-softmmu unicorn-common)
-    target_link_libraries(mips64el-softmmu unicorn-common)
+    target_link_libraries(mips-softmmu PUBLIC unicorn-common)
```
wtdcode, line 1165: Why not `PRIVATE`?

mrexodia: Yeah, I changed it.

CMakeLists.txt
mrexodia: Maybe an additional thought: the way that CMake (seems to be) designed is that your `CMakeLists.txt` is "supposed" to describe how to build your code and nothing more.

This is why for example I don't enable `UNICORN_STATIC_MSVCRT` when unicorn is not compiled as the top-level project. When you are part of someone else's project all you want to do is describe the required files, flags and includes for your library and let the top-level project decide the CMAKE_C_FLAGS and whether they want to install your library (probably not).

CMakeLists.txt
mrexodia: Yeah it does, but I don't see why you would event want to dynamically link the `*-softmmu` libraries. They are part of unicorn's private interface and all users can do something with is `unicorn.so`

In the previous CMake these were modeled by `OBJECT` libraries (likely an even better idea, since that means unicorn.lib will actually contain all the generated code).

CMakeLists.txt
wtdcode: > Maybe an additional thought: the way that CMake (seems to be) designed is that your `CMakeLists.txt` is "supposed" to describe how to build your code and nothing more.
> 
> This is why for example I don't enable `UNICORN_STATIC_MSVCRT` when unicorn is not compiled as the top-level project. When you are part of someone else's project all you want to do is describe the required files, flags and includes for your library and let the top-level project decide the CMAKE_C_FLAGS and whether they want to install your library (probably not).



> If you use unicorn as a submodule you do not want to enable a lot of things that you might enable when you are compiling unicorn.
> 
> Eg:
> 
> ```cmake
> # This will cause PROJECT_IS_TOP_LEVEL to be `OFF`, probably the user doesn't want unicorn's tests, install, fuzzing etc.
> add_subdirectory(unicorn)
> 
> # All you want as a (CMake) user unicorn
> add_executable(foo main.cpp)
> target_link_libraries(foo PRIVATE unicorn)
> ```
> 
> My example template: https://github.com/mrexodia/unicorn_template looks like this on Windows
> 
> image

Nice project! I just gave a star ;p

> Maybe an additional thought: the way that CMake (seems to be) designed is that your `CMakeLists.txt` is "supposed" to describe how to build your code and nothing more.
> 
> This is why for example I don't enable `UNICORN_STATIC_MSVCRT` when unicorn is not compiled as the top-level project. When you are part of someone else's project all you want to do is describe the required files, flags and includes for your library and let the top-level project decide the CMAKE_C_FLAGS and whether they want to install your library (probably not).

Yes exactly and make sense. In this case, I think some macro like `UNICORN_HAS_{ARCH}` should also be exposed?

CMakeLists.txt
mrexodia: If you go to CMake 3.12 (released ~4 years ago) you can change `STATIC` to `OBJECT` and `target_link_libraries` will work seamlessly (in 3.1 you need to use `$<TARGET_OBJECTS:foo>` which is kind of nasty).

https://cmake.org/cmake/help/latest/command/add_library.html#object-libraries

Some people like to complain that their ancient distro doesn't have a recent CMake version, here is a decent post about it: https://alexreinking.com/blog/how-to-use-cmake-without-the-agonizing-pain-part-1.html

CMakeLists.txt
mrexodia: > In this case, I think some macro like UNICORN_HAS_{ARCH} should also be exposed?

What do you mean with this? If you mean `target_compile_definitions(unicorn PUBLIC UNICORN_HAS_{ARCH})`, probably it's unnecessary as there won't be any compile errors if the user uses the x86 features of unicorn, it will just error in `uc_open` (I guess?). But it doesn't hurt to add it of course.

CMakeLists.txt
wtdcode: > Yeah it does, but I don't see why you would ever want to dynamically link the `*-softmmu` libraries. They are part of unicorn's private interface and all users can do something with is `unicorn.so`

I indeed would like to keep `*-softmmu` as static libraries no matter what options users give. Sorry for the confusion.

> 
> In the previous CMake these were modeled by `OBJECT` libraries (likely an even better idea, since that means unicorn.lib will actually contain all the generated code).

This may solve https://github.com/unicorn-engine/unicorn/issues/1508#issuecomment-986094042 if I understand it correctly.

CMakeLists.txt
wtdcode: > > In this case, I think some macro like UNICORN_HAS_{ARCH} should also be exposed?
> 
> What do you mean with this? If you mean `target_compile_definitions(unicorn PUBLIC UNICORN_HAS_{ARCH})`, probably it's unnecessary as there won't be any compile errors if the user uses the x86 features of unicorn, it will just error in `uc_open` (I guess?). But it doesn't hurt to add it of course.

Ah yes, those macros are not accessed in `unicorn.h`. Fine.

CMakeLists.txt
mrexodia: Yes, switching to OBJECT libraries would solve this issue (since it's a cross platform way to merge everything into unicorn.lib)

mrexodia: I think I'll leave this for now, getting linker errors about clashing symbols when attempting.

wtdcode:
Merged, thanks!

PR #1545, opened 2022-01-22
Support to pass unicorn handle to rust through FFI

bet4it:
After this PR, we can do something likes this:
C code:
```c
#include <unicorn/unicorn.h>
#include "rust_ffi.h"

int main()
{
	uc_engine *uc;
	int val = 0;
	uc_open(UC_ARCH_ARM, UC_MODE_ARM, &uc);
	rust_ffi(uc);
	uc_reg_read(uc, UC_ARM_REG_R0, &val);
	printf("R0: %d\n", val);
	uc_close(uc);
	return 0;
}
```
Rust code:
```rust
#![allow(non_camel_case_types)]

use std::ffi::c_void;
use std::convert::TryFrom;
pub type uc_handle = *mut c_void;
use unicorn_engine::Unicorn;
use unicorn_engine::RegisterARM;

#[no_mangle]
pub extern "C" fn rust_ffi(handle: uc_handle) {
    if let Ok(mut unicorn) = Unicorn::try_from(handle) {
        assert_eq!(unicorn.reg_write(RegisterARM::R0, 1234), Ok(()));
    } else {
        panic!("Failed convert handle to Unicorn")
    }
}
```
I think this PR is only meaningful after #1511 is merged, otherwise the memory layout of `uc_engine` may be different between Rust and C.

### DIFF
--- a/bindings/rust/src/ffi.rs
+++ b/bindings/rust/src/ffi.rs
@@ -106,7 +106,7 @@ where
     F: FnMut(&mut crate::Unicorn<D>, u64, usize) -> u64,
 {
     let user_data = unsafe { &mut *user_data };
-    debug_assert_eq!(uc, user_data.uc.inner().uc);
+    debug_assert_eq!(uc, user_data.uc.get_handle());
     (user_data.callback)(&mut user_data.uc, offset, size)
 }
 
@@ -120,7 +120,7 @@ pub extern "C" fn mmio_write_callback_proxy<D, F>(
     F: FnMut(&mut crate::Unicorn<D>, u64, usize, u64),
 {
     let user_data = unsafe { &mut *user_data };
-    debug_assert_eq!(uc, user_data.uc.inner().uc);
+    debug_assert_eq!(uc, user_data.uc.get_handle());
     (user_data.callback)(&mut user_data.uc, offset, size, value);
 }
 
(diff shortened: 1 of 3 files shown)


bet4it:
@domenukk What do you think of it?

domenukk:
Makes sense. Although I'd call the fn `from_uc_ptr` or `from_handle` and it needs to be declared `unsafe` (as you can shoot yourself in the foot).

You can also consider exposing the `handle` so you can do the inverse, calling a c fn with uc handle from rust.

Plus maybe make the fn take a void ptr? Just a thought

bet4it:
> Makes sense. Although I'd call the fn `from_uc_ptr` or `from_handle` and it needs to be declared `unsafe` (as you can shoot yourself in the foot).

I rewrite it with the `TryFrom` trait now. But I don't know how to declare `unsafe` on `try_from`😅

> You can also consider exposing the `handle` so you can do the inverse, calling a c fn with uc handle from rust.

Done

> Plus maybe make the fn take a void ptr? Just a thought

`uc_handle` and void ptr is totally the same thing:
https://github.com/unicorn-engine/unicorn/blob/5f89981a3c8aa2efe3568a79a3e0488ed5628d11/bindings/rust/src/ffi.rs#L10

PR #1547, opened 2022-01-27
s390x TCG shift instruction fixes

iii-i:
Cherry-pick the following upstream commits:

521130f267 target/s390x: Fix SLDA sign bit index
57556b28af target/s390x: Fix SRDA CC calculation
df103c09bc target/s390x: Fix cc_calc_sla_64() missing overflows
6da170beda target/s390x: Fix shifting 32-bit values for more than 31 bits


wtdcode:
Merged, thanks!

PR #1559, opened 2022-02-20
Problem: Could not allocate dynamic translator buffer

### LINKED ISSUE #1556: Could not allocate dynamic translator buffer after 2.0.0-rc5.post1 -> 2.0.0-rc6 upgrade
I have a working code (using Rust binding) that has pretty simple use of Unicorn (map some memory, set up registers, run code), and it worked fine with 2.0.0-rc5.post1. 

After upgrading to 2.0.0-rc6, I get this error:

```
Could not allocate dynamic translator buffer
```

I am going through the diff between these two versions, but maybe somebody can advise on how to narrow down my investigation?

Thank you!

yrashk:
`Could not allocate dynamic translator buffer` error while
trying to perform any memory-related operation.

This happens on aarch64 (MacBook Pro M1)

Solution: ensure aarch64 architecture is properly checked against

Fixes #1556

### DIFF
--- a/qemu/include/qemu/osdep.h
+++ b/qemu/include/qemu/osdep.h
@@ -124,7 +124,7 @@ struct uc_struct;
  * 
  * Issue: https://github.com/desktop/desktop/issues/12978
  */
-#if defined(__APPLE__) && defined(HAVE_PTHREAD_JIT_PROTECT) && defined(__arm__)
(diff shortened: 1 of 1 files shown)


wtdcode:
This is weird. I can't understand why this define is needed.

yrashk:
cc @zhouxy158

wtdcode:
Anyway, it looks harmless so let it go.

yrashk:
@wtdcode what's the `rc` release policy, any ETA for the next rc release? Thanks!

PR #1563, opened 2022-03-02
vb6 bindings: clarify Apache license as v2.0

dzzie:
hopefully this is correct or I will just attach the files


aquynh:
merged, thanks!

Issue #1440, opened 2021-09-01
Some of the x86 registers Unicorn enumerates don't really exist


elicn:
Available control registers should be `CR0`-`CR4`, `CR8`.
The following registers do not exist:
- `UC_X86_REG_CR5`
- `UC_X86_REG_CR6`
- `UC_X86_REG_CR7`
- `UC_X86_REG_CR9`
- `UC_X86_REG_CR10`
- `UC_X86_REG_CR11`
- `UC_X86_REG_CR12`
- `UC_X86_REG_CR13`
- `UC_X86_REG_CR14`
- `UC_X86_REG_CR15`

See:
- Intel SDM vol 1, ch. 3.7.2 "Register Operands"
- Intel SDM vol 3, ch. 2.5 "Control Registers"

Available debug registers should be `DR0`-`DR7` (though `DR4` and `DR5` are not available, they are defined by the arch)
The following registers do not exist:
- `UC_X86_REG_DR8`
- `UC_X86_REG_DR9`
- `UC_X86_REG_DR10`
- `UC_X86_REG_DR11`
- `UC_X86_REG_DR12`
- `UC_X86_REG_DR13`
- `UC_X86_REG_DR14`
- `UC_X86_REG_DR15`

See:
- Intel SDM vol 3, ch. 17.2 "Debug Registers"

The following registers do not seem to exist
- `UC_X86_REG_EIZ`
- `UC_X86_REG_RIZ`
- `UC_X86_REG_FP0`
- `UC_X86_REG_FP1`
- `UC_X86_REG_FP2`
- `UC_X86_REG_FP3`
- `UC_X86_REG_FP4`
- `UC_X86_REG_FP5`
- `UC_X86_REG_FP6`
- `UC_X86_REG_FP7`

wtdcode:
Link to #1449

wtdcode:
I removed some useless registers as you stated, but for FP[x] they are x87 registers.

wtdcode:
> x87 doesn't have registers named FP. It does have `ST0`..`ST7` and `R0`..`R7`, perhaps you are referring to the latter..? If so, I believe that naming those registers `FP0`..`FP7` would be wierd as there is no visible correlation; perhaps `FP_R0`..`FP_R7` would be clearer..
> 
> Consider using a `UC_X87_REG` prefix for x87 registers, where these ones may be named `UC_X87_REG_R0`..`UC_X87_REG_R7`.
> 
> References:
> 
[...]

PR #1572, opened 2022-03-21
Add "holes" to where the removed x86 registers used to be

### LINKED ISSUE #1492: Unicorn 2.0.0rc4 x86 registers definitions are misaligned with Capstone 4.0.2
It looks like the registers reduction mentioned in #1440 made it to Unicorn 2.0.0rc4 (except the FP registers).
Unfortunately that resulted in a misalignment between Capstone and Unicorn x86 registers enumeration, that were aligned to that point. Now Capstone and Unicorn definitions for x86 registers cannot be used interchangebly, which makes them difficult to use together.

Note: probably the best way to re-align the deinfitions would be having Capstone reduct the same non-existing registers, but I don't know if Capstone is still actively maintained.

iii-i:
A number of x86 registers were removed for #1440, causing a change in
numbering for many other registers. This is causing inconveniences at
the moment, e.g. it's not possible to use the Unicorn2 shared library
as a drop-in replacement for the Unicorn1 one.

Restore the old numbering.

Fixes #1492.


wtdcode:
Thanks and sorry for the late reply.

PR #1575, opened 2022-03-26
Fix rust binding for insn_in hook callback

n1tram1:
The x86 IN callback [uc_cb_insn_in_t](https://github.com/n1tram1/unicorn/blob/c10639fd4658a852049546162d116b123e2b1ec2/include/unicorn/unicorn.h#L231) is supposed to return a u32 (the value from the i/o port).

This fixes that by making the callback return a u32.


wtdcode:
Merged, thanks!

Issue #1579, opened 2022-04-02
Uc.emu_stop() doesn't work in a hook if PC is updated


drraid:
This is the same issue that was originally filed as #1133 , I am just filing to re-open (I don't think I can reopen that existing ticket?) because I encountered the same problem. Here is Python to repro the bug:

```
#!/usr/bin/python3

from unicorn import Uc, x86_const, UC_ARCH_X86, UC_MODE_64, UC_HOOK_INTR
from unicorn import UC_PROT_EXEC, UC_PROT_READ
import os
import sys

class Repro(object):
    def __init__(self):
        address  = 0x10000000
        #            nop int3 nop nop nop
        bytecode = b'\x90\xCC\x90\x90\x90'
        emu = Uc(UC_ARCH_X86, UC_MODE_64)
        emu.mem_map(address, 0x1000, UC_PROT_EXEC|UC_PROT_READ)
        emu.mem_write(address, bytecode)
        emu.hook_add(UC_HOOK_INTR, self.int3_callback)
        self.emu = emu
        self.address = address

    def int3_callback(self, emu, interrupt, size):
        if interrupt == 3:
            #print("writing nop back to %x" % (self.address+1))
            emu.mem_write(self.address+1, b'\x90')
            #print("press enter to set PC")
            #sys.stdin.readline()
            emu.reg_write(x86_const.UC_X86_REG_RIP, self.address+1)
            #print("press enter to emu_stop()")
            #sys.stdin.readline()
            emu.emu_stop()

    def run(self):
        emu = self.emu
        #print("press enter to emu_start()")
        #sys.stdin.readline()
        emu.emu_start(self.address, self.address+4)
        pc = emu.reg_read(x86_const.UC_X86_REG_RIP)
        expected = self.address + 1 
        print("expect: %x received: %x" % (expected, pc))

def main():
    #print(str(os.getpid()))
    repro = Repro()
    repro.run()

if __name__ == "__main__":
    main()
```

wtdcode:
Sorry I forget this issue. Will have a look ASAP.

drraid:
I originally opened a PR with this project but I screwed it up because I was making changes against a v2 RC instead of v1. Anyway I have since forked unicorn and applied my own patch to v1, if you want to use that: 

https://github.com/drraid/unicorn/commit/b237335e6f2a321b7e2f4acf4ce7fdfe91fcd61f

I can open a PR for these same changes against the real Unicorn repo v1 branch if you would like?

[...]

Issue #1356, opened 2020-11-27
Writing register IP causes previous emu_stop useless


wtdcode:
Code snippet.

```python
from unicorn import *
from unicorn.arm64_const import UC_ARM64_REG_W0, UC_ARM64_REG_X0
from unicorn.x86_const import *
import time

def test():
    count = 0
    def cb(a, b, c, d):
        nonlocal count
        count += 1
        print(f"Count {count}")
        if count <= 10:
            # This one never works.
            mu.emu_stop()
            # This line cause previous emu_stop useless.
            mu.reg_write(UC_X86_REG_IP, 0x1000)
        else:
            mu.emu_stop()



    try:
        mu = Uc(UC_ARCH_X86, UC_MODE_16)
        address_to_load = 0x1000
        mu.mem_map(0x1000, 4 * 1024, UC_PROT_ALL)
        cmd =  b'\xb8\x00\x00\xeb\xfb' #  ks.asm("lb:\nmov ax, 0h\njmp lb\n", 0x1000)
        mu.reg_write(UC_X86_REG_IP, 0x1000)
        mu.reg_write(UC_X86_REG_CS, 0)
        mu.mem_write(address_to_load, cmd)
        mu.hook_add(UC_HOOK_CODE, cb)
        mu.emu_start(address_to_load, address_to_load + len(cmd))
    except UcError as e:
        print("ERROR: %s" % e)

if __name__ == '__main__':
    test()
```

Output on my machine:

```
Count 1
Count 2
Count 3
Count 4
Count 5
Count 6
Count 7
Count 8
Count 9
Count 10
Count 11
```

Some investigation shows that:

```c
                    case UC_X86_REG_IP:
                        X86_CPU(uc, mycpu)->env.eip = *(uint16_t *)value;
                        // force to quit execution and flush TB
                        uc->quit_request = true;
                        uc_emu_stop(uc);
                        break;
```

`uc->quit_request` is set to true when writing to register ip.

And that resets `uc->stop_request` and thus makes previous `emu_stop` useless.

```c
static bool tcg_exec_all(struct uc_struct* uc)
{
    int r;
    bool finish = false;
    while (!uc->exit_request) {
        CPUState *cpu = uc->cpu;
        CPUArchState *env = cpu->env_ptr;

        //qemu_clock_enable(QEMU_CLOCK_VIRTUAL,
        //                  (cpu->singlestep_enabled & SSTEP_NOTIMER) == 0);
[...]

PR #1580, opened 2022-04-02
Proposed fix for #1579: adjusting stop_request and quit_request

drraid:
Slightly changed how stop_request and quit_request are set and 
checked. First, this adds uc_emu_soft_stop(), which now has all
the same logic that was previously in uc_emu_stop(), except it
does not set uc->stop_request.  The uc_emu_stop() function now
calls uc_emu_soft_stop() and then sets uc->stop_request. The
point of making uc_emu_soft_stop its own function was to split
out the logic of using stop_request vs quit_request: places
that just "quit" (as in flush but then continue emulating) still
set the uc->quit_request flag, but now call uc_emu_soft_stop()
instead of uc_emu_stop(). The logic in softmmu/cpus.c now only
checks if uc->stop_request is set to determine if it should
break out of the loop.

The target architectures have been updated to call the new
uc_emu_soft_stop instead of uc_emu_stop for their "quits".

TL;DR stop_request and quit_request are now one-or-the-other
instead of one-and-maybe-the-other.

This is to address #1579 

NOTE: This is my first time looking at the inner workings of
Unicorn, and a grown-up should review this carefully because
I probably did this wrong ;). I wanted to attempt a fix instead 
of simply filing a bug with a repro. I will not be offended if my
fix is partially or completely rejected!

### DIFF
--- a/include/unicorn/unicorn.h
+++ b/include/unicorn/unicorn.h
@@ -828,6 +828,20 @@ UNICORN_EXPORT
 uc_err uc_emu_start(uc_engine *uc, uint64_t begin, uint64_t until,
                     uint64_t timeout, size_t count);
 
+
+/*
+ "Soft" stop emulation: perform all the logic of stopping emulation
+ without setting the uc.stop_request. Most callers will use this when
+ setting quit_request to flush TB and continue execution. The function
+ uc_emu_stop() calls this function and then sets uc.stop_request.
+
+ @uc: handle returned by uc_open()
+
+ @return UC_ERR_OK on success, or other value on failure (refer to uc_err enum
+   for detailed error).
+*/
+uc_err uc_emu_soft_stop(uc_engine *uc);
+
 /*
  Stop emulation (which was started by uc_emu_start() API.
  This is typically called from callback functions registered via tracing APIs.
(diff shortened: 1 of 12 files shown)


wtdcode:
Both #1579 and #1173 have been known bugs for Unicorn for quite a long time (since I get familiar with this project #1356).

However, I do think it's not a critical bug or design flaw since users still have easy workaround to achieve what they would like.

Could you come up with a scenario where this API is really essential? Note we are extremely cautious about adding/removing an existing API.

drraid:
I'm going to close this pull request because I'm pretty sure I broke other stuff, and I will look at it more to propose a better fix that doesn't break stuff. 

I hadn't seen #1356 - my apologies for yet another duplicate.

I don't want to add or remove an API, but I think it would be good if the API behaved as expected. In my scenario I have created a library to wrap Unicorn and provide functionality like breakpoints, which is how I ran into this issue. When implementing a breakpoint I want to notify the user that the breakpoint has been reached without executing the instruction, and when notifying the user of the breakpoint I also want to have the PC point correctly to the instruction where the breakpoint is. I have implemented the breakpoint the same way I have in debuggers: catch the INT3, write back the original byte, set the PC back to that byte, then notify the user. If a user wants to stop execution (or emulation) at this point, they should have the choice to do so. 

I could hack around it by fully stopping the emulator when I receive the INT3, and then fully starting again if the user indicates that the breakpoint should proceed, but then that only solves the expected behavior for the breakpoint case, and doesn't change the expectation of uc_emu_stop(). 

I totally understand the desire to not change APIs, and especially understand your caution in accepting pull requests as I'm pretty sure my PR is broken :), but a correct fix feels like it would be aligned with user expectations of the API.

PR #1573, opened 2022-03-25
Fix clang-cl compile with CONFIG_INT128 option in QEMU

rose4096:
This fixes the linker error:
`lld-link: error: undefined symbol: __udivti3`
when compiling with clang-cl on windows.

PR #1581, opened 2022-04-03
Fix clang-cl compile with CONFIG_INT128 in qemu

rose4096:
[#1573](https://github.com/unicorn-engine/unicorn/pull/1573)


wtdcode:
Merged, thanks.

PR #1584, opened 2022-04-08
fix rust bindings build on windows

shuffle2:
Some things I noticed while trying to get unicorn-engine to work on windows/rust:

First I just tried 
```
unicorn-engine = "2.0.0-rc6"
```
and I got
```
--- stderr
  thread 'main' panicked at 'Fail to create build directory on Windows.: Error { kind: NotFound, message: "program not found" }'
```
This (IMO) is kind of nonsense error. because the problem was really that it could not find cmake.

OK, so I run `"c:\Program Files\Microsoft Visual Studio\2022\Community\vc\Auxiliary\Build\vcvarsall.bat" x64` and try again:
```
  = note: LINK : fatal error LNK1181: cannot open input file 'unicorn.lib'
```
ok...so I head to this github and see that there are related changes in `dev` branch, so I do this:
```
unicorn-engine = { git = "https://github.com/unicorn-engine/unicorn.git", branch = "dev" }
```
but still get
```
  = note: LINK : fatal error LNK1181: cannot open input file 'unicorn-static.lib'
```
and, interestingly, checking in `target\release\build\unicorn-engine-2c01a69f559d00b5\output`, I see:
```
cargo:warning=Unicorn not found. Downloading...
cargo:rustc-link-search=<mypath>\target\release\build\unicorn-engine-2c01a69f559d00b5\out\unicorn-engine-unicorn-c10639f\build_rust\Release
```
This means that build.rs of unicorn has downloaded tarball of **master** branch even though it's executing from dev branch checkout (note `c10639f` in the path). fwiw this is problematic even when checking out dev branch of unicorn manually and using `unicorn-engine = { path = "unicorn/bindings/rust" }`

So...:
1. I'm not that familiar with crates that need to invoke cmake from build.rs, but surely there is some better way than requiring the user to start cargo from a msvc build environment? Or would this not be a problem if the project had a real, working crate published? idk
2. IMO the cmake build generator should be Ninja, not VS. this can be relied upon to exist in VS env, but idk if you want to rely on it for other platforms
3. the download step of build.rs should be aware of cargo settings such as the wanted branch (I have no idea how to do that, or if it's the proper thing, just pointing it out :D) I eventually had to set `UNICORN_LOCAL`, clean build tree, and retry to get it working, which IMO is kinda gross.

This PR at least makes it build, but doesn't really fix the above problems.
What it does:
1. use ninja cmake generator on all platforms
2. use `cmake --build . --parallel` to have cmake automagically do the build, which should work everywhere. (this is handled by `cmake` crate)
3. uses build dir for `rustc-link-search`
4. removes the "download unicorn" logic, which was broken and not useful.


wtdcode:
Hello, the Windows rust build is indeed broken these days and I'm trying to fix it.

> IMO the cmake build generator should be Ninja, not VS. this can be relied upon to exist in VS env, but idk if you want to rely on it for other platforms

Yes, ninja is better in this case but still I don't think you can really build out of VS prompt.

> the download step of build.rs should be aware of cargo settings such as the wanted branch (I have no idea how to do that, or if it's the proper thing, just pointing it out :D) I eventually had to set UNICORN_LOCAL, clean build tree, and retry to get it working, which IMO is kinda gross.

Exactly. It's controlled by an environment variable for now and not ideal I admit.

bindings/rust/build.rs
```diff
@@ -1,170 +1,58 @@
...
-        }
+use std::path::Path;
+
+#[cfg(all(feature = "build_unicorn_cmake", target_os = "windows"))]
+fn setup_env_windows() {
+    let devenv_path = cc::windows_registry::find_tool(
+        env::var("TARGET").unwrap().as_str(), "devenv").unwrap();
+    let devenv_dir = devenv_path.path().parent().unwrap();
+
+    let cmake_path = devenv_dir.join(r"CommonExtensions\Microsoft\CMake\CMake\bin\cmake.exe");
```
wtdcode, line 14: This seems not good.

I feel it better to call `cmake` directly, which also works for msys2 if it (really) works.

bindings/rust/build.rs
```diff
@@ -1,170 +1,58 @@
...
+    let src_dir = manifest_path.parent().unwrap().parent().unwrap();
+
+    #[cfg(target_os = "windows")]
+    setup_env_windows();
+
+    // need to clear build target and append "build" to the path because
+    // unicorn's CMakeLists.txt doesn't properly support 'install', so we use
+    // the build artifacts from the build directory, which cmake crate sets
+    // to "<out_dir>/build/"
+    let dst = cmake::Config::new(src_dir)
```
wtdcode, line 46: You seem to forget Debug/Release switch.

shuffle2: it's handled by cmake crate

shuffle2:
> Another consideration of not using ninja is that ninja is not shipped with MSVC while msbuild absolutely will.

ninja will be present if cmake is, so it's not a problem for msvc/VS. To be clear, with my changes i can build my project (which depends on unicorn-engine) cleanly, and not have to start VS environment manually.

bindings/rust/build.rs
shuffle2: This is the best way for msvc. I'm not sure about msys2 (i assume you mean mingw).

shuffle2:
It looks like current version of this PR will work fine if I just make it fallback to cmake's default generator if ninja isn't found. That would solve all the problems i've listed, while still keeping the matrix of CI builders happy. IDK about things that are not in CI (like mingw on msys2).

bindings/rust/build.rs
```diff
@@ -1,170 +1,58 @@
...
-        .unwrap()
-        .bytes()
-        .unwrap();
-    let tar = GzDecoder::new(resp.reader());
 
-    let mut archive = Archive::new(tar);
-    archive.unpack(&out_dir).unwrap();
-
-    find_unicorn(&out_dir).unwrap()
+    if let Some(path) = env::var_os("PATH") {
```
wtdcode, line 23: I just saw you are patching `PATH` env and to be honest I don't think it's a clean and good solution...

bindings/rust/build.rs
wtdcode: https://docs.microsoft.com/en-us/cpp/build/building-on-the-command-line?view=msvc-170#use-the-developer-tools-in-an-existing-command-window

shuffle2: this crate is used by other rust projects; i.e. i just want to type `cargo run` and it works. similarly, other tools (like rust language servers) will break if the build.rs doesn't "just work".

wtdcode: > this crate is used by other rust projects; i.e. i just want to type `cargo run` and it works. similarly, other tools (like rust language servers) will break if the build.rs doesn't "just work".

I mean, if we really would like to help users build it without the VS prompt, we shouldn't do such hack which might break in a further MSVC release.

shuffle2: Anything can break on future msvc/VS release. do you see the big red warning directly above the section you linked? :)

bindings/rust/build.rs
shuffle2: we have 2 choices when modifying `PATH`: append or prepend. I've chosen append here. lmk if it should be prepend.

bindings/rust/build.rs
wtdcode: But my concern is that: Could `CommonExtensions\Microsoft\CMake\CMake\bin\cmake.exe` change?

shuffle2:
sorry, i had a goof, it should actually work now..i think

bindings/rust/build.rs
shuffle2: It's unlikely, but it could. It's equally likely the location of vsvars script itself changes. The actual solution is pushing the problem to be handled by the `cc` (or maybe `cmake`) crate, but for now they don't have the functionality. In real life, everyone has to update their tools with each new VS, anyway (it's just that it should be `cc` crate's problem to deal with that, not unicorn's).

bindings/rust/build.rs
```diff
@@ -1,174 +1,72 @@
...
-}
-
-#[cfg(feature = "build_unicorn_cmake")]
-fn out_dir() -> PathBuf {
-    let out_dir = env::var("OUT_DIR").unwrap();
-    Path::new(&out_dir).to_path_buf()
-}
-
-#[cfg(feature = "build_unicorn_cmake")]
-fn download_unicorn() -> PathBuf {
```
wtdcode, line 37: If we don't support downloading unicron tarball, users could only clone the full repo to use. I think it makes sense to allow users to use unicorn just by specifying the dependency in cargo.toml.

bindings/rust/build.rs
wtdcode: I suggest restoring the `PATH` variable after everything is done.

shuffle2: bit confused what you're trying to prevent? cargo compiles and runs build.rs. i.e. unicorn's build.rs will run as it's own process - any env vars it modifies will not be persisted to the parent process or external system. see https://doc.rust-lang.org/std/env/fn.set_var.html . persisting env vars outside your process is OS/shell specific.

btw, there's also this in rust docs (https://doc.rust-lang.org/cargo/reference/build-scripts.html):
> The build script will be rebuilt if any of its source files or dependencies change.
> By default, Cargo will re-run the build script if any of the files in the package changes. Typically it is best to use the rerun-if commands, described in the [change detection](https://doc.rust-lang.org/cargo/reference/build-scripts.html#change-detection) section below, to narrow the focus of what triggers a build script to run again.
see also https://github.com/rust-lang/cargo/issues/2261#issuecomment-170461557

So the `cargo:rerun-if-changed` lines which unicorn's build.rs script has are not required.

bindings/rust/build.rs
shuffle2: Do you mean you want to avoid people having to download the git history? I don't think there's a good way around that (might be worth looking into), and suspect that's what everyone else does...

The current way is probably just as bad because of the amount of dependencies it brings in :) (try `cargo tree` in the unicorn rust dir!)

wtdcode: > Do you mean you want to avoid people having to download the git history? I don't think there's a good way around that (might be worth looking into), and suspect that's what everyone else does...

No, users should have a way to use it as crate without cloning the repo. For example, just `cargo new hello` and add  unicorn to their cargo.toml (and get the bindings from crates.io) should work. Since at this moment rust still doesn't have a stable way to distribute binray so this is the only way I think.

wtdcode: Sorry I'm really not a rust expert and my solution indeed seems dumb.

It looks like the correct way to handle this and you may work on it. Thanks for your patience to revise the rust bindings :).

shuffle2:
still need to fix some things, dont bother running CI yet

bindings/rust/build.rs
```diff
@@ -69,6 +69,9 @@ fn build_with_cmake() {
     let compiler = cc::Build::new().get_compiler();
     let has_ninja = if compiler.is_like_msvc() {
         setup_env_msvc(&compiler);
+        // this is a BIG HACK that should be fixed in unicorn's cmake!!
```
shuffle2, line 72: I needed this for `cargo test` to succeed on my machine...the linking stuff in the CMakeLists.txt is pretty messy, but someone else can fix it..

shuffle2:
it _should_ be ready to go now

bindings/rust/build.rs
wtdcode: Weird, what symbol collisions do you get?

shuffle2: here is how `target\debug\build\unicorn-engine-<hash>\output` looks after e.g. `cargo clean && cargo build`:
https://gist.github.com/shuffle2/f9623c334f1ea82b3d995c5ef6155d0a

when you run `cargo test`, it links an executable instead of a library, so these `LNK4006` warnings turn into errors.

shuffle2: ah yea, looks like ubuntu CI builder had same issue

CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 # CMake setup for Unicorn 2.
 # By Huitao Chen & Nguyen Anh Quynh, 2019-2020
 
-cmake_minimum_required(VERSION 3.1)
+cmake_minimum_required(VERSION 3.15)
```
wtdcode, line 4: This breaks some slightly older Ubuntu.

shuffle2: https://stackoverflow.com/a/49921543 looks like "slightly" is pretty old...

CMakeLists.txt
```diff
@@ -32,6 +35,12 @@ option(UNICORN_INSTALL "Enable unicorn installation" ${PROJECT_IS_TOP_LEVEL})
 set(UNICORN_ARCH "x86;arm;aarch64;riscv;mips;sparc;m68k;ppc;s390x" CACHE STRING "Enabled unicorn architectures")
 option(UNICORN_TRACER "Trace unicorn execution" OFF)
 
+if(DEFINED ENV{CARGO})
```
wtdcode, line 38: Is there any specific reason not to put this to build.rs?

CMakeLists.txt
shuffle2: fwiw i raised it because it allows using `CMAKE_MSVC_RUNTIME_LIBRARY`, and changes behavior around how cmake handles the msvcrt settings. On older cmake, they supply a default value which is hard/impossible to override, and it is something that should be done for rust.

This also relates to adding the CARGO variable, which is mostly used for configuring msvcrt setting. I can see if I can move it somewhat back into build.rs, but it will still require more recent cmake version either way.

CMakeLists.txt
shuffle2: i've pushed another commit which just uses CARGO flag (that is set automatically by cargo) to control the msvcrt linking.

bindings/rust/build.rs
shuffle2: for now i've added the equivalent flag to non-msvc build

CMakeLists.txt
wtdcode: > https://stackoverflow.com/a/49921543 looks like "slightly" is pretty old...

Ubuntu 18 seems not that old, at least it's not EOL.

wtdcode: > fwiw i raised it because it allows using `CMAKE_MSVC_RUNTIME_LIBRARY`, and changes behavior around how cmake handles the msvcrt settings. On older cmake, they supply a default value which is hard/impossible to override, and it is something that should be done for rust.
> 
> This also relates to adding the CARGO variable, which is mostly used for configuring msvcrt setting. I can see if I can move it somewhat back into build.rs, but it will still require more recent cmake version either way.

Is there any reference for this? Let me also check how could we work out this.

bindings/rust/build.rs
shuffle2: ugh, let me fix + test locally on linux first

CMakeLists.txt
shuffle2: i've updated it to not need CARGO indicator at all now

bindings/rust/build.rs
shuffle2: should be fixed now, will have to see if it also works for clang/macos

CMakeLists.txt
shuffle2: > > https://stackoverflow.com/a/49921543 looks like "slightly" is pretty old...
> 
> Ubuntu 18 seems not that old, at least it's not EOL.

i've changed it back to min-version = 3.1, but if compiling for msvc it will either enable the new behavior or error out if < 3.15

Cargo.toml
```diff
@@ -0,0 +1,44 @@
+[package]
```
wtdcode, line 1: Any reason to put cargo.toml at root directory?

shuffle2: yea, it's required to include the C sources into the package (the Cargo.toml in bindings/rust cannot really reference things above it)

bindings/rust/build.rs
shuffle2: sigh, so lld on macos doesn't support the flag...

shuffle2: hm, did it not cause an error there to begin with? 🤔 https://github.com/unicorn-engine/unicorn/runs/6035089850?check_suite_focus=true

wtdcode: Seems so. I would also have a try locally tonight.

shuffle2: do you know what `cc` (in PATH) and `CC` (env var) are set to on mac os?

wtdcode: > do you know what `cc` (in PATH) and `CC` (env var) are set to on mac os? It's obviously clang, but it seems to screw up the detection the rust cc crate is using to identify the compiler
> 
> probably something like this: [alexcrichton/cc-rs#408](https://github.com/alexcrichton/cc-rs/issues/408)

cc is AppleClang on recent macOS but cmake might wrongly detect it

shuffle2: yea it seems cc is an actual binary (not a symlink). in any case i've worked around it

wtdcode:
Looks good to me! I really appreciate your work and it could be merged once I test it locally.

Thanks!

shuffle2:
Btw, I notice that the base windows environments provided for GitHub actions already include all the tools you use (msbuild,cmake,ninja,etc) : [windows server 2022](https://github.com/actions/virtual-environments/blob/main/images/win/Windows2022-Readme.md). Additionally it includes older versions of the tool chains (you can target vs2019, vs2022, etc ) from same image. So, your ci flow can be sped up a little by skipping some steps which are redundant (not for this pr, tho)

wtdcode:
> Btw, I notice that the base windows environments provided for GitHub actions already include all the tools you use (msbuild,cmake,ninja,etc) : [windows server 2022](https://github.com/actions/virtual-environments/blob/main/images/win/Windows2022-Readme.md). Additionally it includes older versions of the tool chains (you can target vs2019, vs2022, etc ) from same image. So, your ci flow can be sped up a little by skipping some steps which are redundant (not for this pr, tho)

Yes, the CI setup is quite old and can be optimized in many ways but that's not prioritized too much for now.

Issue #1586, opened 2022-04-11
Memory leaks caused by unexpected architecture in unicorn dev branch


liyansong2018:
Hi :)

When we used unexpected architecture such as `0`, there was a memory leak in unicorn2.

PoC
```c
int main(int argc, char **argv) {
    uc_engine *uc;
    uc_err err;

    /* Initialize emulator in X86-64bit mode */
    err = uc_open(0, UC_MODE_64, &uc);
    
    if (err == UC_ERR_OK) {
        printf("Failed on uc_open() with error returned: %u\n", err);
        printf("%s\n", uc_strerror(err));
        return -1;
    }

    if (err == UC_ERR_ARCH) {
        printf("Failed on uc_open() with error returned: %u\n", err);
        printf("%s\n", uc_strerror(err));
    } 

    uc_close(uc);
    return 0;
}

```

If we don't use `uc_close(uc)`, there will be a memory leak.
```shell
$ ./poc_test
Failed on uc_open() with error returned: 2
Invalid/unsupported architecture (UC_ERR_ARCH)

=================================================================
==22499==ERROR: LeakSanitizer: detected memory leaks
                                                                                                                           
Direct leak of 14720 byte(s) in 1 object(s) allocated from:
    #0 0x7f2082fde037 in __interceptor_calloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:154                  
    #1 0x7f2081bddfce in uc_open /home/lys/Documents/my/unicorn/uc.c:236
    #2 0x563051f0b278 in main /home/lys/Documents/unitest/poc_test.c:17
    #3 0x7f20816a47ec in __libc_start_main ../csu/libc-start.c:332

SUMMARY: AddressSanitizer: 14720 byte(s) leaked in 1 allocation(s).

```

If we use `uc_close(uc)`, there will be a segmentation fault!
```shell
$ ./poc_test
Failed on uc_open() with error returned: 2
Invalid/unsupported architecture (UC_ERR_ARCH)
AddressSanitizer:DEADLYSIGNAL
=================================================================
==22535==ERROR: AddressSanitizer: SEGV on unknown address 0x00000000075d (pc 0x7f5eb03ee46b bp 0x7ffe556843e0 sp 0x7ffe556843c0 T0)                                                                                                                   
==22535==The signal is caused by a READ memory access.                                                                     
==22535==Hint: address points to the zero page.
[...]

PR #1587, opened 2022-04-11
Fix https://github.com/unicorn-engine/unicorn/issues/1586

liyansong2018:
If `uc->init_arch == NULL`, uc need to be retruned because it already points to heap space. Of course, we can also directly free the corresponding memory space.

```c
if (uc->init_arch == NULL) {
    free(uc);
    return UC_ERR_ARCH;
}
```


wtdcode:
Merged, thanks.

Issue #1590, opened 2022-04-13
Integer overflow in memory check


liyansong2018:
When Unicorn applies for memory for a virtual machine, it will make a preliminary judgment on the starting address and size of the memory.

e.g. memory aligned, memory efficient.

![bug](https://user-images.githubusercontent.com/25031216/163095581-f97c2522-ddae-45dc-87c9-8f72dc1b8be0.png)

Integer overflows are prone to occur when checking memory boundaries.
```c
// address cannot wrapp around
if (address + size - 1 < address) {
    return UC_ERR_ARG;
}
```

For example, we apply `0xfffffffffffff000 ~ 0xfffffffffffff000 + 0x1000`
```c
#define ADDRESS 0xfffffffffffff000
#define SIZE 0x1000

int main(int argc, char **argv) {
    uc_engine *uc;
    uc_err err;

    err = uc_open(UC_ARCH_X86, UC_MODE_64, &uc);
    if (err != UC_ERR_OK) {
        printf("Failed on uc_open() with error returned: %u %s\n", err, uc_strerror(err));
        return -1;
    }

    err = uc_mem_map(uc, ADDRESS, SIZE, UC_PROT_ALL);
    if (err == UC_ERR_MAP | err == UC_ERR_ARG) {
        printf("Failed on uc_mem_map() with error returned: %u %s\n", err, uc_strerror(err));
    }

    uc_close(uc);
    return 0;
}
```

Memory check passed. Howerver, GVA Memory Region end address equals 0 !

![bug1](https://user-images.githubusercontent.com/25031216/163096606-d8f4b48d-8a0a-471d-bd43-c034825692e4.png)

wtdcode:
Closed due to PR merged.

wtdcode:
This is not bug and we have handled this case correctly.

liyansong2018:
I'm not sure if `virtual end address=0` is a reasonable value.  (0xfffffffffffff000 - 0).

For example,  linux kernel virtual addresses start at 0xC0000000 and go to 0xFFFFFFFF  from [WikiPedia](https://en.wikipedia.org/wiki/High_memory).
> The lower 896 MB, from 0xC0000000 to 0xF7FFFFFF, is directly mapped to the kernel physical address space, and the remaining 128 MB, from 0xF8000000 to 0xFFFFFFFF, is used on demand by the kernel to be mapped to high memory. 


[...]

PR #1592, opened 2022-04-13
Fix https://github.com/unicorn-engine/unicorn/issues/1590

liyansong2018:
We only need to add simple judgment to prevent integer overflow.

If we apply `0xfffffffffffff000 ~ 0xfffffffffffff000 + 0x1000` like #1590, we will not be able to bypass the memory check!
```shell
Failed on uc_mem_map() with error returned: 15 Invalid argument (UC_ERR_ARG)
```

### DIFF
--- a/uc.c
+++ b/uc.c
@@ -975,6 +975,10 @@ static uc_err mem_map_check(uc_engine *uc, uint64_t address, size_t size,
     }
 
     // address cannot wrapp around
+    if (address + size < address || address + size < size) {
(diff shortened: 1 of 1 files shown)


wtdcode:
Merged, thanks!

wtdcode:
Unfortunately, this fix is wrong and will be reverted.

`0xfffffffffffff000 ~ 0xfffffffffffff000 + 0x1000` is a valid region and thus it should be mapped here.

PR #1594, opened 2022-04-13
Fix static build install directory

bet4it:


wtdcode:
Fixed in the latest dev.

PR #1608, opened 2022-05-04
Issues and tests for jump hook address

QDucasse:
Hello, in x86 hooks for unmapped memory get an address with a 1-bit offset. 
The two tests added show that:
- The `movabs` instruction gets executed correctly as the value is set in `rax`
- The `jmp` instruction gets executed correctly as the value is set in `rip`
- The address obtained in the hook is incorrect (uncommenting the `printf` shows the address with a 1-bit offset)
- The same behavior is correct in ARM64 and RISC-V

Edit from merge with @PalumboN:

Fix patch https://github.com/unicorn-engine/unicorn/issues/1615 and https://github.com/unicorn-engine/unicorn/pull/1608
- Adding tb_exec_locks needed for Mac M1
- Ignore second hooks calls after memory mapping error (bug)
- Comment page align macro because is not working properly

Maybe we need to iterate this solutions, but now I have all tests green and it's working with PharoVM

### DIFF
--- a/qemu/accel/tcg/cputlb.c
+++ b/qemu/accel/tcg/cputlb.c
@@ -1419,6 +1419,9 @@ load_helper(CPUArchState *env, target_ulong addr, TCGMemOpIdx oi,
             uintptr_t retaddr, MemOp op, bool code_read,
             FullLoadHelper *full_load)
 {
+    // Hack for double load on error
+    if (env->uc->cpu->exit_request) return 0;
+
     uintptr_t mmu_idx = get_mmuidx(oi);
     uintptr_t index = tlb_index(env, mmu_idx, addr);
     CPUTLBEntry *entry = tlb_entry(env, mmu_idx, addr);
@@ -1449,7 +1452,10 @@ load_helper(CPUArchState *env, target_ulong addr, TCGMemOpIdx oi,
                     continue;
                 if (!HOOK_BOUND_CHECK(hook, addr))
                     continue;
-                if ((handled = ((uc_cb_eventmem_t)hook->callback)(uc, UC_MEM_FETCH_UNMAPPED, addr, size - uc->size_recur_mem, 0, hook->user_data)))
+                tb_exec_lock(uc->tcg_ctx);
+                handled = ((uc_cb_eventmem_t)hook->callback)(uc, UC_MEM_FETCH_UNMAPPED, addr, size - uc->size_recur_mem, 0, hook->user_data);
+                tb_exec_unlock(uc->tcg_ctx);
+                if ((handled))
                     break;
 
                 // the last callback may already asked to stop emulation
--- a/qemu/exec.c
+++ b/qemu/exec.c
@@ -1087,8 +1087,8 @@ RAMBlock *qemu_ram_alloc_from_ptr(struct uc_struct *uc, ram_addr_t size, void *h
     RAMBlock *new_block;
     ram_addr_t max_size = size;
 
-    size = HOST_PAGE_ALIGN(uc, size);
-    max_size = HOST_PAGE_ALIGN(uc, max_size);
+    // size = HOST_PAGE_ALIGN(uc, size);
+    // max_size = HOST_PAGE_ALIGN(uc, max_size);
     new_block = g_malloc0(sizeof(*new_block));
     if (new_block == NULL)
         return NULL;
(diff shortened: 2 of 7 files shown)


wtdcode:
Hello, note the target branch is wrong.

qemu/exec.c
```diff
@@ -1087,8 +1087,8 @@ RAMBlock *qemu_ram_alloc_from_ptr(struct uc_struct *uc, ram_addr_t size, void *h
     RAMBlock *new_block;
     ram_addr_t max_size = size;
 
-    size = HOST_PAGE_ALIGN(uc, size);
```
wtdcode, line 1090: The reason for this fix?

tests/unit/test_x86.c
```diff
@@ -981,6 +981,110 @@ static void test_x86_nested_uc_emu_start_exits(void)
...
+
+
+    uc_assert_err(
+        UC_ERR_FETCH_UNMAPPED,
+        uc_emu_start(uc, code_start, code_start + sizeof(code) - 1, 0, 0));
+
+    OK(uc_reg_read(uc, UC_X86_REG_RAX, &r_rax));
+    OK(uc_reg_read(uc, UC_X86_REG_RIP, &r_rip));
+    TEST_CHECK(r_rax == 0x7F00);
+    TEST_CHECK(r_rip == 0x7F00);
```
wtdcode, line 1022: I can't understand why `r_ip==0x7F00`. This jump won't be taken due to the error and the rip should be left unchanged.

qemu/accel/tcg/cputlb.c
```diff
@@ -1419,6 +1419,9 @@ load_helper(CPUArchState *env, target_ulong addr, TCGMemOpIdx oi,
             uintptr_t retaddr, MemOp op, bool code_read,
             FullLoadHelper *full_load)
 {
+    // Hack for double load on error
+    if (env->uc->cpu->exit_request) return 0;
```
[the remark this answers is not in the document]
PalumboN: Sorry, it's not LOAD. Should say: `Hack for double hooks call on memory map error`. It's the error reported in this PR.

I don't know exactly why it is happening, my debugging time ends and we need to have this working for the PharoVM team.

The problem is on `UC_MEM_FETCH_UNMAPPED` error. The simulator performs another cycle (increment IP -> try to execute -> finish in another `UC_MEM_FETCH_UNMAPPED`) before ends.

We need to debug the `translate.c` execution flow to understand better where is the source of the problem.

qemu/exec.c
PalumboN: In Mac M1 this is returning bad addresses. I'm not sure why and I do not understand this code here, because `size` and `max_size` has exactly the same value at this point.

qemu/accel/tcg/cputlb.c
```diff
@@ -1419,6 +1419,9 @@ load_helper(CPUArchState *env, target_ulong addr, TCGMemOpIdx oi,
             uintptr_t retaddr, MemOp op, bool code_read,
             FullLoadHelper *full_load)
 {
+    // Hack for double load on error
```
PalumboN, line 1422:
```suggestion
    // Hack for double hooks call on memory map error
```

qemu/exec.c
PalumboN: Maybe you can find a test to need this code, if not should we remove it?

wtdcode: This coepath I believe is very common and I can't come up with why it doesn't work on M1 since I don't have test environment.

qemu/accel/tcg/cputlb.c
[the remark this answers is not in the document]
wtdcode: If I understand it correctly, the tests added in this PR should reflect your issues, right?

BTW, how is unicorn related to PharoVM...?

tests/unit/test_x86.c
PalumboN: I think it's ok to do this assert (maybe could be in other test, but is good to be tested).

The point is: what's the expected behaviour? I understood that Unicorn worked as this test describe (IP = Jump's target address).

wtdcode:
This should be fixed in 5d37e21db501bcebaac9346b82c2d969b0c56c02.

For M1 related issue, please open another PR.

PR #1612, opened 2022-05-08
check pkg-config for unicorn libraries

lowlyw:
allow for non-standard install locations of unicorn engine

this is required for when i am trying to package unicorn for other package managers.


wtdcode:
Hello, thanks for your PR. But it should go to dev branch.

btw, may I know what package manager you are packaging unicorn for?

lowlyw:
i'm packaging for macports, i'll send a PR to the other branch now.

PR #1613, opened 2022-05-08
check pkg-config for unicorn libraries

lowlyw:
allow for non-standard install locations of unicorn engine


lowlyw:
I'm packaging this for macports, this is required here.

wtdcode:
Thanks for your PR ;)

PR #1617, opened 2022-05-16
 Make the behavior of `uc_ctl` consistent with `uc_query`

bet4it:
https://github.com/unicorn-engine/unicorn/blob/94cac9dfb494d7eb6cf563ae4880068da6a5eb42/uc.c#L1807-L1812


wtdcode:
This inconsistency is left by design because `uc_ctl_get_mode` is designed to get the **mode specified in `uc_open`** while`uc_query` should have been the same behavior but is left unchanged due to compatibility.

To query the **current** mode (only ARM may have this flexibility AFAIK), `uc_reg_read` is a better option.

bet4it:
Sounds reasonable. We may need to document it somewhere.

PR #1621, opened 2022-05-19
Support reads and writes over all Arm SIMD registers

ondryaso:
This PR adds support for using all types of Arm SIMD registers (Q0–Q15, D0–D31 and S0–S15) in `uc_reg_read` and `uc_reg_write`.


wtdcode:
Looks good and thanks!

PR #1629, opened 2022-06-08
New and improved Python bindings

elicn:
Re-implemented Unicorn Python bindings while remaining compatible with the original bindings API.

Highlights:
- OOP design for improved efficiency and extensibility
- Pythonic code featuring type annotations to facilitate its usage
- Cleaner code and documented methods
- Lower memory footprint
- Improved Uc resources deallocation

Changes:
- `UcError` now provides additional information, when applicable, to provide the users with better error context (e.g. the target address and access size for invalid memory accesses)
- Intel arch now supports reading and writing ZMM registers
- ARM arch now supports reading and writing Q registers


bindings/python/unicorn/unicorn.py
```diff
@@ -1,492 +1,303 @@
...
+            'libwinpthread-1.dll',
+            'libgcc_s_seh-1.dll',
+            'libgcc_s_dw2-1.dll'
+        )
+
+        for dllname in all_dlls:
+            if dllname not in loaded_dlls:
+                lib_file = path / dllname
+
+                if str(path.parent) == '.' or lib_file.exists():
```
wtdcode, line 61: `str(path.parent) == '.'` seems not equavelent to original logic? I suppose it should be `str(path)`.

bindings/python/unicorn/unicorn.py
```diff
@@ -502,454 +313,819 @@ def _finalizer(self, ref):
...
+
+        uc_maj, uc_min, _ = uc_version()
+        bnd_maj, bnd_min, _ = version_bind()
+
+        return (uc_maj, uc_min) == (bnd_maj, bnd_min)
+
+
+    def __new__(cls, arch: int, mode: int):
+        # prevent direct instantiation of unicorn subclasses
+        assert cls is Uc, f'{cls.__name__} is not meant to be instantiated directly'
```
wtdcode, line 443: This kills the possibility of inheriting `Uc` class, which I did in implementing multitask. A quick PoC:

```python
from unicorn import *

class A(Uc):
    
    def __init__(self, arch: int, mode: int):
        super().__init__(arch, mode)
        

a = A(UC_ARCH_X86, UC_MODE_32)
```

I understand you are trying to return a subclass to handle arch-specific things, however I doubt why not choose `pimpl` design? e.g. `Uc` holds a reference to an arch-specific object instead of being a subclass. IMHO, hacking with `__new__` is a bit dangerous, especially we care about the compatibility.

bindings/python/unicorn/unicorn.py
```diff
@@ -958,95 +1134,63 @@ def context(self):
...
-            all_archs += "-%s" % k
-
-    major, minor, _combined = uc_version()
-
-    return "python-%s-c%u.%u-b%u.%u" % (
-        all_archs, major, minor, uc.UC_API_MAJOR, uc.UC_API_MINOR
-    )
+
+
+__all__ = ['Uc', 'UcContext']
```
wtdcode, line 1196: This might be a bit aggressive as stated above.

bindings/python/unicorn/unicorn.py
```diff
@@ -1,492 +1,303 @@
...
-class uc_x86_mmr(ctypes.Structure):
-    """Memory-Management Register for instructions IDTR, GDTR, LDTR, TR."""
-    _fields_ = [
-        ("selector", ctypes.c_uint16),  # not used by GDTR and IDTR
-        ("base", ctypes.c_uint64),      # handle 32 or 64 bit CPUs
-        ("limit", ctypes.c_uint32),
-        ("flags", ctypes.c_uint32),     # not used by GDTR and IDTR
-    ]
-
-class uc_x86_msr(ctypes.Structure):
```
wtdcode, line 437: I love your new design of the registers read/write per arch, but it is a bit aggressive to remove these classes since they might be really used somewhere, though unexpected.

bindings/python/unicorn/arch/arm.py
```diff
@@ -0,0 +1,59 @@
...
+        return self.val
+
+    @classmethod
+    def from_param(cls, param: ARMCPRegValue):
+        assert type(param) is tuple and len(param) == len(cls._fields_)
+
+        return cls(*param)
+
+
+class UcAArch32(Uc):
```
wtdcode, line 42: As stated further in the review below, my thought is that, such the arch-specific class might be used as a member of the `Uc` object. This in fact is `pimpl` design instead of your `factory` design if I understand it correctly.

bindings/python/unicorn/unicorn.py
```diff
@@ -1,492 +1,303 @@
...
-
-# setup all the function prototype
-def _setup_prototype(lib, fname, restype, *argtypes):
-    try:
-        getattr(lib, fname).restype = restype
-        getattr(lib, fname).argtypes = argtypes
-    except AttributeError:
-        raise ImportError("ERROR: Fail to setup some function prototypes. Make sure you have cleaned your unicorn1 installation.")
-
-ucerr = ctypes.c_int
```
wtdcode, line 126: Though these alias doesn't make too much sense, I think it doesn't matter too much to just leave them there ;).

wtdcode (commented):
First of all, thanks for your brilliant work!

The overall design fits my taste but there are a few problems to be fixed. See my comments above.

bindings/python/unicorn/unicorn.py
elicn: As far as I understood it, the original logic was trying to figure out whether this is a relative library path (i.e. either a filename without a parent directory, or just not an absolute path). Testing for parent == "." is the way to determine this. Did you encounter any problem loading the library this way? (I admit I haven't tested it thoroughtly)

I am not fully comfortable with how the library lookup code ended up, but considering the previous implementation I think it looks very nice after all..

bindings/python/unicorn/unicorn.py
elicn: I could not think of any reasonable usecase for extending the `Uc` class, but if you think it is necessary I can patch it to allow user-defined subclasses. Regarding the multitask implementation, I don't know how it looks like - but perhaps we could integrate this into this model, still "hiding" the subclasses from the user..?

About using `__new__` to create the required subclass: that is not a hack - that is a common way to use this special method. The idea here was to organize everything in OOP (with all the benefits coming from it) and at the same time don't let the user be bothered about which class to instantiate - they just init `Uc`, which picks the right subclass for them. The user should not be aware (or care) of the specific subclass.

The incapsulation in the PIMPL design makes things more complicated and packs the code with trivial wrapper calls (i.e. forwarding the the exact same parameters to a method with exact same name). This design pattern is more reasonable for inflexible languages like C++, rather than Python. However, I don't see how it helps with the multithreading implementation because we'll still have a predefined list of subclasses to incapsulate.

bindings/python/unicorn/unicorn.py
elicn: Importing `*` from a library is considered a bad practice and should be avoided.
As far as I can tell the Unicorn bindings library is organized in way that `__init__.py` already imports all the needed modules and "flatten" the import path (i.e. you can import everything directly from `unicorn`; if you can't import - it means you shouldn't).

Keep in mind that the previous implementation kept everything in one big file, so there was no consideration of a clean namespace. Having said that, I think that moving to a new major version (uc2) is a good oportunity to make such changes that could break some scripts - but at the same time will be trivial to fix by the users by adding a few more imports.

bindings/python/unicorn/unicorn.py
elicn: These classes represent registers in the way Uc sees them, so I don't think they were meant to be used externally.
As stated above, this is another step towards a cleaner namespace.

bindings/python/unicorn/unicorn.py
elicn: They were relocated to somewhere else to help with code readability.

bindings/python/unicorn/unicorn.py
wtdcode: This multitask code is [here](https://github.com/qilingframework/qiling/pull/1105/files#diff-dca875e9c03a42d626868ef51f8f2c41ce860b96094bd53534116643e07af179R138).

bindings/python/unicorn/arch/intel.py
```diff
@@ -0,0 +1,207 @@
...
+    REG_RANGE_ZMM = range(const.UC_X86_REG_ZMM0, const.UC_X86_REG_ZMM31 + 1)
+
+    def hook_add(self, htype: int, callback: Callable, user_data: Any = None, begin: int = 1, end: int = 0, aux1: int = 0, aux2: int = 0) -> int:
+        if htype != UC_HOOK_INSN:
+            return super().hook_add(htype, callback, user_data, begin, end, aux1, aux2)
+
+        insn = ctypes.c_int(aux1)
+
+        def __hook_insn_in():
+            @_catch_hook_exception
```
wtdcode, line 104: This is incorrect. `_catch_hook_exception` will change the function signature and thus `__hook_insn_in_cb`'s first argument `uc` will be incorrectly set to the raw handle.

wtdcode:
See my PoC in #1633

bindings/python/unicorn/unicorn.py
elicn: I played around with this code quite a bit in order to find a way to keep the clean OOP approach and still allow users to extend the `Uc` class. As far as I can tell we have two options for that:
1. Require external `Uc` subclasses to be implemented as "pimpl", assuming this is probably a rare use case anyway. Looking at the multitask code, it seems that such a change would not affect it too much.
2. Require external `Uc` subclasses to be decorated with something like `@ucsubclass`. That way is a bit hacky and involves some meta-programming black magic, but it looks clean from user perspective. Taking your example above:
```python
@ucsubclass
class A(Uc):
    def __init__(self, arch: int, mode: int):
        super().__init__(arch, mode)
```
I lean towards the 2<sup>nd</sup> option; what do you think?

elicn:
@wtdcode, any thoughts or comments, considering [this comment](https://github.com/unicorn-engine/unicorn/pull/1629#discussion_r968874319) and [this commit](https://github.com/unicorn-engine/unicorn/pull/1629/commits/765ec5ffe474c7d1dbfb4a55525091bc02a1eeed)?

aquynh:
we have been thinking about merging this after 2.0.1 release.

can you add some sample code under bindings/python/ to demonstrate the new API?

elicn:
> can you add some sample code under bindings/python/ to demonstrate the new API?

The external facing API was not modified, it is only the internal implementation that was changed. Except minor additions (e.g. some archs now support more regs) nothing really changed in terms of functionality. `UcError` now provides more info, so that could be one possible feature to show. Do you have anything specific on your mind..?

aquynh:
oh i thought you added some more Pythonic API

wtdcode:
merged and thanks for your patient.

I'm doing a careful comparison to ensure nothing breaks.

elicn:
It passes all local tests, if that counts.

wtdcode:
Our tests for Python bindings are not complete and thus I'm looking into the history to manually port missing features.

PR #1632, opened 2022-06-23
Remove UnicornInner in Rust bindings

### LINKED ISSUE #1619: Rust bindings leak emulators
Steps to reproduce the bug:

1. Create an emulator
2. Add a hook to it
3. Drop the emulator (just let it go out of scope)

Expected behaviour: all resources related to the emulator are freed.

Actual behaviour: the emulator is not freed.

The following code showcases the problem:

```rust
use std::thread;
use std::time::Duration;

use unicorn_engine::unicorn_const::{Arch, Mode, Permission};
use unicorn_engine::Unicorn;

pub fn code_hook(_: &mut Unicorn<()>, _: u64, _: u32) {}

fn main() {
    const MEM_SIZE: usize = 0x10000000; // 256 MB
    const PAGE_SIZE: usize = 0x1000;
    const START_ADDRESS: u64 = 0x10000;

    for i in 0..0x100 {
        println!("{i}");

        let mut emu = unicorn_engine::Unicorn::new(Arch::X86, Mode::MODE_64)
            .expect("failed to initialize Unicorn instance");
        emu.mem_map(START_ADDRESS, MEM_SIZE, Permission::ALL)
            .expect("failed to map code page");

        // Touch all pages to actually map the memory
        for address in (START_ADDRESS..START_ADDRESS + MEM_SIZE as u64).step_by(PAGE_SIZE) {
            emu.mem_write(address, &[0xcc]).unwrap();
        }

        emu.add_code_hook(0, u64::MAX, code_hook).unwrap();

        thread::sleep(Duration::from_secs(1));
        // emu is dropped here
    }
}
```

At each point in time, only one emulator should exist, so the memory usage should stay constant, around `MEM_SIZE`. Instead you will see that after each second, the memory usage increases by `MEM_SIZE`, indicating that the emulator is not freed.

The root cause of the problem is that the `UnicornInner` does not get dropped when `Unicorn` is dropped. This happens because `inner` is an `Rc`, which only gets dropped if its reference count is exactly one.

https://github.com/unicorn-engine/unicorn/blob/ca81d46ad53e2f324780bbf4e60f49483cb98fef/bindings/rust/src/lib.rs#L131-L155

However, the `add_code_hook` function clones `inner`, increasing the reference count. This introduces a circluar reference, because the `inner` holds a reference to the hooks, but the hooks also hold a reference to `inner`; hence `inner` will never get dropped.

https://github.com/unicorn-engine/unicorn/blob/ca81d46ad53e2f324780bbf4e60f49483cb98fef/bindings/rust/src/lib.rs#L626

The same issue holds for all `add_*_hook` functions, as well as `mmio_map`.

bet4it:
Fix #1619

### DIFF
--- a/bindings/rust/build.rs
+++ b/bindings/rust/build.rs
@@ -1,4 +1,3 @@
-use pkg_config;
 use std::env;
 use std::path::PathBuf;
 use std::process::Command;
--- a/bindings/rust/src/ffi.rs
+++ b/bindings/rust/src/ffi.rs
@@ -88,13 +88,20 @@ extern "C" {
 }
 
 pub struct UcHook<'a, D: 'a, F: 'a> {
+    pub unicorn: *mut Unicorn<'a, D>,
     pub callback: F,
-    pub uc: Unicorn<'a, D>,
 }
 
 pub trait IsUcHook<'a> {}
 
 impl<'a, D, F> IsUcHook<'a> for UcHook<'a, D, F> {}
+
+macro_rules! destructure_hook {
+    ($hook:ident) => {{
+        let UcHook { unicorn, callback } = unsafe { &mut *$hook };
+        (unsafe { &mut **unicorn }, callback)
+    }};
+}
 
 pub extern "C" fn mmio_read_callback_proxy<D, F>(
     uc: uc_handle,
@@ -103,11 +110,11 @@ pub extern "C" fn mmio_read_callback_proxy<D, F>(
     user_data: *mut UcHook<D, F>,
 ) -> u64
 where
-    F: FnMut(&mut crate::Unicorn<D>, u64, usize) -> u64,
+    F: FnMut(&mut Unicorn<D>, u64, usize) -> u64,
 {
-    let user_data = unsafe { &mut *user_data };
-    debug_assert_eq!(uc, user_data.uc.get_handle());
-    (user_data.callback)(&mut user_data.uc, offset, size)
+    let (unicorn, callback) = destructure_hook!(user_data);
+    debug_assert_eq!(uc, unicorn.get_handle());
+    callback(unicorn, offset, size)
 }
 
 pub extern "C" fn mmio_write_callback_proxy<D, F>(
@@ -117,11 +124,11 @@ pub extern "C" fn mmio_write_callback_proxy<D, F>(
     value: u64,
     user_data: *mut UcHook<D, F>,
 ) where
-    F: FnMut(&mut crate::Unicorn<D>, u64, usize, u64),
+    F: FnMut(&mut Unicorn<D>, u64, usize, u64),
 {
-    let user_data = unsafe { &mut *user_data };
-    debug_assert_eq!(uc, user_data.uc.get_handle());
-    (user_data.callback)(&mut user_data.uc, offset, size, value);
+    let (unicorn, callback) = destructure_hook!(user_data);
+    debug_assert_eq!(uc, unicorn.get_handle());
+    callback(unicorn, offset, size, value);
 }
 
 pub extern "C" fn code_hook_proxy<D, F>(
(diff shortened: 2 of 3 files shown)


afgTheCat:
Hey I started using the new binds, and I have problems mapping memory within a hook. Example:
```rust
use unicorn_engine::unicorn_const::{Arch, HookType, MemType, Mode, Permission};
use unicorn_engine::Unicorn;

pub fn page_fault_hook<D>(
    emu: &mut Unicorn<D>,
    _mem_type: MemType,
    address: u64,
    _mem_size: usize,
    _value: i64,
) -> bool {
    let page_start = address & 0xfffffffffffff000;
    emu.mem_map(page_start, 0x1000, Permission::READ | Permission::WRITE)
        .unwrap();
    true
}

fn main() {
    let mut unicorn = Unicorn::new(Arch::X86, Mode::MODE_64).unwrap();
    unicorn
        .add_mem_hook(HookType::MEM_INVALID, 0, u64::MAX, page_fault_hook)
        .expect("Could not add mem hook");
}
```
The compiler will tell me that the `D` may not live long enough. I could not change the hook's signature so that I escape all the lifetime errors, however changing the `Unicorn` struct the following way seems to solve my issues:
```rust
pub struct Unicorn<'a, D: 'a> {
    handle: uc_handle,
    ffi: bool,
    arch: Arch,
    /// to keep ownership over the hook for this uc instance's lifetime
    hooks: Vec<(ffi::uc_hook, Box<dyn ffi::IsUcHook<'a> + 'a>)>,
    /// To keep ownership over the mmio callbacks for this uc instance's lifetime
    mmio_callbacks: Vec<MmioCallbackScope<'a>>,
    data: D,
}
```
Is there a reason why Unicorn's data can't be restricted to `'a`?

bet4it:
I think you are right! Fixed.

wtdcode:
@bet4it Could you check domenukk's comments?

bet4it:
> Hmm, aliasing, and later accessing, a `&mut Unicorn` borrow (via a raw ptr) is undefined behavior in Rust, AFAIK. I don't think the PR is sound as is(?)

I'm a newbie in Rust, and I really can't understand what @domenukk said is about.

PhilippTakacs:
> > Hmm, aliasing, and later accessing, a `&mut Unicorn` borrow (via a raw ptr) is undefined behavior in Rust, AFAIK. I don't think the PR is sound as is(?)
> 
> I'm a newbie in Rust, and I really can't understand what @domenukk said is about.

I haven't found the code he is referencing, but what he means is following:

Somewhere in the code you create a raw pointer from a `&mut Unicorn` and access this raw pointer after the reference is out of scope. This is undefined in Rust because rust is allow to move or destroy the underling object. So you need to ensure that the object is not moved or destroyed.

wtdcode:
Thanks for your help @PhilippTakacs !

A small tip on ub: https://github.com/rust-lang/miri . Using miri may be super helpful.

bet4it:
> OK, I have understand this now, for the callbacks you create a raw pointer of self and add this to the user data of the callback. later in the callback-wrapper you access the pointer and cast it to a Unicorn object. This is undefined in rust. The old code used `InnerUnicorn` and `UnsafeCell` to fix this. But this code has no fix for this problem.

Do you mean that we cast the pointer to an Unicorn object, but if the pointer doesn't point to an Unicorn object, the behavior is undefined? But the callback is only called by the Unicorn object, so we can assure the pointer is an Unicorn object in the callback. That's what we have done in C, right?

bet4it:
So you think `InnerUnicorn` is necessary? Then this PR is meaningless and we can close it😅

PhilippTakacs:
Yes

bet4it:
In favor of #1764

PR #1648, opened 2022-07-10
Bug fix for java binding of Apple silicon.

zhkl0228:
Because Java also has JIT code, Java binding in Apple M1 does not execute "tb_exec_lock" at the end of execution, which will cause "SIGBUS" exception in the execution of JIT code in Java.


wtdcode:
I don't think it's the correct way to fix though it might solve your issue for now. Why is java special in this case?

zhkl0228:
The java code will also be converted into local code by jit. When the unicorn execution is completed, if pthread_jit_write_protect is not called, the jit at the java layer will fail to execute.

AeonLucid:
This PR also fixed unicorn crashing in .NET 6 on a M1 Macbook Air. Would be great to get fixed.

wtdcode:
It's quite unclear to me how these fixes work without an M1 test environment. I can review this probably after I get one for debugging.

AeonLucid:
Not stale.

wtdcode:
Applied the correct fix here: 6e97e59f5416e2cd3ad7d476061d9f7dc38eb66c

Issue #1627, opened 2022-05-25
Repeated memory write callbacks


archercreat:
Hello. While tracking memory writes in heavily packed executable to some memory region I've found that in some cases the callback is triggered once for 8 byte write and at the same time 8 times for 1 byte write.
Unfortunately I cannot reproduce this behavior on simple example, so maybe someone could point on what is the possible error might be.
Example trace:
```
write size, value, disassembly
size: 8 value: 0xfffffffffffff7ff 0x1400cd26d: mov      qword ptr [r10], r11
size: 1 value: 0x00000000000000ff 0x1400cd26d: mov      qword ptr [r10], r11
size: 1 value: 0x00000000000000f7 0x1400cd26d: mov      qword ptr [r10], r11
size: 1 value: 0x00000000000000ff 0x1400cd26d: mov      qword ptr [r10], r11
size: 1 value: 0x00000000000000ff 0x1400cd26d: mov      qword ptr [r10], r11
size: 1 value: 0x00000000000000ff 0x1400cd26d: mov      qword ptr [r10], r11
size: 1 value: 0x00000000000000ff 0x1400cd26d: mov      qword ptr [r10], r11
size: 1 value: 0x00000000000000ff 0x1400cd26d: mov      qword ptr [r10], r11
size: 1 value: 0x00000000000000ff 0x1400cd26d: mov      qword ptr [r10], r11
size: 8 value: 0xffffffff000165a1 0x140007fad: push     rdi
size: 4 value: 0x00000000fffcb99d 0x140007fae: xor      dword ptr [rsp], eax
size: 8 value: 0x00000001400aae62 0x14009194e: push     rbx
```

You can see that only `size: 8 value: 0xfffffffffffff7ff 0x1400cd26d: mov      qword ptr [r10], r11` should be present but for some reason it triggers callback 8  more times.

wtdcode:
If the address is badly aligned, then a read/write is probably split into multiple aligned read/write. In worst cases on some arch, this means byte by byte read/write.
________________________________
From: Pavel ***@***.***>
Sent: Wednesday, May 25, 2022 4:17:03 PM
To: unicorn-engine/unicorn ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [unicorn-engine/unicorn] Repeated memory write callbacks (Issue #1627)


[...]

PR #1651, opened 2022-07-13
Handle pathological cases more consistently

rhelmot:
These are two changes I've needed to make to make angr work with unicorn 2.

First commit: presently, if we try to execute a non-executable region, the lifting code never checks to see if reads succeed, it just blindly trusts the result. As a consequence, we may execute some non-executable guest code or hooks before checking cpu->exit_request and breaking out of the loop. This fixes that.

Second commit: If you respond to an unmapped memory fault during instruction fetch by mapping the related memory in a callback, qemu will only attempt to lift a single instruction for the current block. This is incredibly annoying (and breaks angr's tests, which are sensitive to block counts and boundaries), so I fixed it.

### DIFF
--- a/qemu/accel/tcg/translate-all.c
+++ b/qemu/accel/tcg/translate-all.c
@@ -1584,9 +1584,8 @@ TranslationBlock *tb_gen_code(CPUState *cpu,
     phys_pc = get_page_addr_code(env, pc);
 
     if (phys_pc == -1) {
-        /* Generate a temporary TB with 1 insn in it */
-        cflags &= ~CF_COUNT_MASK;
-        cflags |= CF_NOCACHE | 1;
+        /* Generate a temporary TB; do not cache */
+        cflags |= CF_NOCACHE;
     }
 
     cflags &= ~CF_CLUSTER_MASK;
--- a/qemu/accel/tcg/cpu-exec.c
+++ b/qemu/accel/tcg/cpu-exec.c
@@ -595,6 +595,9 @@ int cpu_exec(struct uc_struct *uc, CPUState *cpu)
             }
 
             tb = tb_find(cpu, last_tb, tb_exit, cflags);
+            if (unlikely(cpu->exit_request)) {
+                continue;
+            }
             cpu_loop_exec_tb(cpu, tb, &last_tb, &tb_exit);
             /* Try to align the host and virtual clocks
                if the guest is in advance */
--- a/qemu/accel/tcg/cputlb.c
+++ b/qemu/accel/tcg/cputlb.c
@@ -1451,7 +1451,7 @@ load_helper(CPUArchState *env, target_ulong addr, TCGMemOpIdx oi,
                         continue;
                     if (!HOOK_BOUND_CHECK(hook, addr))
                         continue;
-                    if ((handled = ((uc_cb_eventmem_t)hook->callback)(uc, UC_MEM_FETCH_UNMAPPED, addr, size - uc->size_recur_mem, 0, hook->user_data)))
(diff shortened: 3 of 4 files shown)


rhelmot:
I have added a third commit which should address my concerns on #1627. This uses the uc->size_recur_mem field, which from what I can tell was completely unused prior to this, to detect whether we are in an unaligned write (which is implementing by a recursive call into the write helper).

wtdcode:
> I have added a third commit which should address my concerns on #1627. This uses the uc->size_recur_mem field, which from what I can tell was completely unused prior to this, to detect whether we are in an unaligned write (which is implementing by a recursive call into the write helper).

Regarding #1627, could you add a unit test? I assume it should be easy to reproduce?

rhelmot:
Pushed!

rhelmot:
Idk what's up with the failing test - it passed on my previous commit and I only changed an unrelated testcase. Is it flaky?

wtdcode:
> Idk what's up with the failing test - it passed on my previous commit and I only changed an unrelated testcase. Is it flaky?

mingw32 runner has some bugs for a while… I really should disable it. Anyway let me rerun and it will be fine.

qemu/accel/tcg/translate-all.c
```diff
@@ -1584,9 +1584,8 @@ TranslationBlock *tb_gen_code(CPUState *cpu,
     phys_pc = get_page_addr_code(env, pc);
 
     if (phys_pc == -1) {
-        /* Generate a temporary TB with 1 insn in it */
-        cflags &= ~CF_COUNT_MASK;
-        cflags |= CF_NOCACHE | 1;
+        /* Generate a temporary TB; do not cache */
+        cflags |= CF_NOCACHE;
```
wtdcode, line 1588: I tried your patch but I failed to understand the change here. Any reproduction code?

firodj:
thanks @wtdcode for directing to this. I try this patch for issue #1041, this patch fixes the write hook, but not for the read hook.

qemu/accel/tcg/translate-all.c
rhelmot: Are you asking for a testcase?

The meaning of the original code is to set the count bitfield of the cflags to 1 (only lift 1 instruction) and set the CF_NOCACHE bit (do not cache the tb). The new code performs only the latter change.

wtdcode: > Are you asking for a testcase?
> 
> The meaning of the original code is to set the count bitfield of the cflags to 1 (only lift 1 instruction) and set the CF_NOCACHE bit (do not cache the tb). The new code performs only the latter change.

It's better if a testcase could be provided so that I can understand it better. My question is that, why setting the count to 1 solves the problem?

rhelmot: You're misunderstanding the order of the diff - the old code set the count to 1, and this was undesired behavior because then it would only lift one instruction. Presumably the old code optimizes guests using exceptional control flow, but it produces inaccurate metrics (block counts).

rhelmot:
@firodj please try the commit that I just pushed!

@wtdcode I believe this is ready to merge.

rhelmot:
It seems that CI is failing on aarch64 and ppc64le platforms. I don't have either of those machines to debug this. How can I proceed?

wtdcode:
> It seems that CI is failing on aarch64 and ppc64le platforms. I don't have either of those machines to debug this. How can I proceed?

Weird, I can't come up with why your code will fail on these platforms. Maybe any undefined behaviors? Let me re-run it firstly.

wtdcode:
Merged and fixed in `dev` branch. Thanks!

wtdcode:
When performing memory read operation on aarch64 and ppc64, no `load_helper` is generated...

PR #1658, opened 2022-07-20
Fail when VEX.L is set in SSE instructions (AVX is not supported)

### LINKED ISSUE #1656: Ymm registers are 128 bit, but they should be 256 bit
Hello!

Next sample is copying only 16 bytes, but it should copy 32 bytes:

```python
from unicorn import *
import binascii

CODE = (b"\xc5\xfe\x6f\x09"      # vmovdqu ymm1,ymmword ptr [rcx]
        b"\xc5\xfd\x7f\x0a")     # vmovdqa ymmword ptr [rdx],ymm1

MEM_IN = b"\x11" * 32

ADDRESS_CODE = 0x1000000
ADDRESS_IN   = 0x2000000
ADDRESS_OUT  = 0x3000000

try:

    mu.mem_map(ADDRESS_CODE, 0x1000)
    mu.mem_map(ADDRESS_IN, 0x1000)
    mu.mem_map(ADDRESS_OUT, 0x1000)

    mu.mem_write(ADDRESS_CODE, CODE)
    mu.mem_write(ADDRESS_IN, MEM_IN)

    mu.reg_write(UC_X86_REG_RCX, ADDRESS_IN)
    mu.reg_write(UC_X86_REG_RDX, ADDRESS_OUT)

    mu.emu_start(ADDRESS_CODE, ADDRESS_CODE + len(CODE))

    mem_out = mu.mem_read(ADDRESS_OUT, 32)
    print(f"mem @rdx = {binascii.hexlify(mem_out)}")

except UcError as e:
    print("ERROR: %s" % e)

```

The output:

```
mem @rdx = b'1111111111111111111111111111111100000000000000000000000000000000'
```

Any ideas?

Thanks

mrexodia:
Closes #1656

According to https://gitlab.com/qemu-project/qemu/-/issues/132 YMM isn't supported in qemu upstream at all. I still have to test on a real CPU what exception is thrown, but other checks on `vex_l` seem to indicate it is `illegal_op`. An alternative would be to add the check to `vex_l` to every opcode, but since YMM isn't supported at all I think this is reasonable.

### DIFF
--- a/tests/regress/x86_vex.c
+++ b/tests/regress/x86_vex.c
@@ -47,12 +47,37 @@ static void test_vmovdqu(void)
     OK(uc_close(uc));
 }
 
+/* https://github.com/unicorn-engine/unicorn/issues/1656 */
+static void test_vex_l(void)
+{
+    uc_engine *uc;
+    uc_err err;
+
+    /* vmovdqu ymm1, [rcx] */
+    char code[] = { '\xC5', '\xFE', '\x6F', '\x09' };
+
+    /* initialize memory and run emulation  */
+    OK(uc_open(UC_ARCH_X86, UC_MODE_64, &uc));
+    OK(uc_mem_map(uc, 0, 2 * 1024 * 1024, UC_PROT_ALL));
+
+    OK(uc_mem_write(uc, 0, code, sizeof(code) / sizeof(code[0])));
+
+    err = uc_emu_start(uc, 0, sizeof(code) / sizeof(code[0]), 0, 0);
+    if(err != UC_ERR_INSN_INVALID) {
+        fprintf(stderr, "%s", uc_strerror(err));
+        assert(false);
+    }
+
+    OK(uc_close(uc));
+}
+
 
 /* TODO: Add more vex prefixed instructions
          Suggestions: vxorpd, vxorps, vandpd, ... */
 int main(int argc, char **argv, char **envp)
 {
         test_vmovdqu();
+        test_vex_l();
         return 0;
 }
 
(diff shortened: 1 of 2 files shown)


mrexodia:
Follow up: I do not have a real CPU that doesn't support AVX (I only have the first i3 model from 2011 that introduced AVX) so I wasn't able to figure out the exception thrown. Regardless I think it's better to fail loudly than to execute instructions incorrectly.

tests/regress/x86_vex.c
```diff
@@ -47,12 +47,37 @@ static void test_vmovdqu(void)
     OK(uc_close(uc));
 }
 
+/* https://github.com/unicorn-engine/unicorn/issues/1656 */
+static void test_vex_l(void)
```
wtdcode, line 51: Thanks for your contribution. The test case under `tests/regress` is not built & run with CI and should have been moved to `tests/unit`. Could you move this test to `tests/unit/test_x86.c`?

wtdcode:
> Follow up: I do not have a real CPU that doesn't support AVX (I only have the first i3 model from 2011 that introduced AVX) so I wasn't able to figure out the exception thrown. Regardless I think it's better to fail loudly than to execute instructions incorrectly.

Agree-ed. I think `illegal_op` is pretty fine as it will be populated to users correctly with `UC_ERR_INSN_INVALID`

tests/regress/x86_vex.c
mrexodia: I enabled “Allow edits and access to secrets by maintainers“, feel free to edit as you see fit!

wtdcode:
Merged & fixed.

mrexodia:
Awesome, thanks!

PR #1659, opened 2022-07-20
Fixed endianness when writing PPC32 CR register.

zachsez:
When writing a value to the CR (conditional register) the order of the nibbles was reversed. I believe the nibble ordering in the `reg_read` is correct, so I correct `reg_write` to maintain the same endianness. Also included a unit test to make sure that the value we write is the value we read out.


wtdcode:
Looks pretty good to me, thanks & welcome!

wtdcode:
You are sending to the wrong branch. It should go to `dev`. Anyway I fixed it already.

zachsez:
Sorry about that, if I PR in the future, I'll use `dev`.

PR #1665, opened 2022-07-23
Fixes #1661 - Hook block infinite loop when changing PC

### LINKED ISSUE #1661: Existence of HOOK_CODE changes behavior of HOOK_BLOCK
I have a unicorn instance hooking functions via the `HOOK_BLOCK` callback. In that callback I alter the stack and the IP.
This method worked fine for me in versions up until version 2.0.rc7 (rc6 and 1.0.3 worked fine).
When I changed the callback from `HOOK_BLOCK` to `HOOK_CODE` (in the newer version) the code worked fine.
The weird thing is that if I do something like this
```
temp_hook = uc.hook_add(UC_HOOK_CODE, temp_func)
uc.hook_del(temp_hook)
block_hook = uc.hook_add(UC_HOOK_BLOCK, block_callback)
```
then the code works as it was before.
Even before doing `emu_start` it seems to matter somehow.

AeonLucid:
Issue is described in #1661.

> I have the exact same issue. My arm64 code is jumping to a block that is hooked with HOOK_BLOCK, I then handle it in my own code and set the X0 and PC so the arm64 code resumes. This used to work fine with Unicorn 1. Now, when I run the same code it fails to work properly and keeps calling the block, creating an infinite loop.
> 
> Then, I added a HOOK_CODE block in an attempt to debug it. To my confusion it started working after I added the debug hook.

Reverting commit https://github.com/unicorn-engine/unicorn/commit/b7bc13650c56ebaad47264c7c7cf5a5a72e25fd4 fixes the issue. 

I can't tell you why because I don't understand the internals well, it is the first commit I tried when @paulkermann mentioned it was a regression in RC7.

If there is a better fix please feel free to commit to the PR.

### DIFF
--- a/qemu/accel/tcg/cpu-exec.c
+++ b/qemu/accel/tcg/cpu-exec.c
@@ -72,7 +72,7 @@ static inline tcg_target_ulong cpu_tb_exec(CPUState *cpu, TranslationBlock *itb)
          * of the start of the TB.
          */
         CPUClass *cc = CPU_GET_CLASS(cpu);
-        if (!HOOK_EXISTS(env->uc, UC_HOOK_CODE)) {
+        if (!HOOK_EXISTS(env->uc, UC_HOOK_CODE) && !env->uc->timeout) {
             // We should sync pc for R/W error.
             switch (env->uc->invalid_error) {
                 case UC_ERR_WRITE_PROT:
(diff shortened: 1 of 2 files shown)


wtdcode:
This revert will cause a few other regressions when no `UC_HOOK_CODE` is installed. To fix it correctly, could provide a reproduction script?

AeonLucid:
> This revert will cause a few other regressions when no `UC_HOOK_CODE` is installed. To fix it correctly, could provide a reproduction script?

I provided one, in both the issue and this PR.

wtdcode:
> > This revert will cause a few other regressions when no `UC_HOOK_CODE` is installed. To fix it correctly, could provide a reproduction script?
> 
> I provided one, in both the issue and this PR.

Good, I will look into it.

AeonLucid:
@wtdcode Great, thanks! If you have time, can you also check https://github.com/unicorn-engine/unicorn/pull/1648? I have added a repro for that one as well.

wtdcode:
Fixed in c4a0813

Issue #1683, opened 2022-08-15
Incorrect argument types for calls to uc_ctl_request_cache/uc_ctl_remove_cache in test_uc_ctl_tb_cache


relapids:
In test_uc_ctl_tb_cache the calls to uc_ctl_request_cache/uc_ctl_remove_cache expect uint64_t for the address arguments, but the test invokes them with int. For some configurations (e.g. I was using a Clang targeting x86 on Linux) this will result in the arguments being passed incorrectly (since they're passed as 32-bit).

This segfaults for me on a Release build of sample_ctl since the TB pointer argument to uc_ctl_request_cache is read incorrectly.

Note this is specific to sample_ctl since test_ctl uses the 'code_start' variable in the calculation, which is a uint64_t.

Easy fix is to cast them to 64-bit explicitly, or to change sample_ctl to use an explicit 64-bit type for ADDRESS similar to how test_ctl uses code_start.

Relevant code (unmodified) below:
```
    // Now we request cache for all TBs.
    for (int i = 0; i < TB_COUNT; i++) {
        err = uc_ctl_request_cache(uc, ADDRESS + i * TCG_MAX_INSNS, &tb);
        printf(">>> TB is cached at 0x%" PRIx64 " which has %" PRIu16
               " instructions with %" PRIu16 " bytes.\n",
               tb.pc, tb.icount, tb.size);
        if (err) {
            printf("Failed on uc_ctl() with error returned: %u\n", err);
            return;
        }
    }
```
```
    case UC_CTL_TB_REQUEST_CACHE: {

        UC_INIT(uc);

        if (rw == UC_CTL_IO_READ_WRITE) {
            uint64_t addr = va_arg(args, uint64_t);
            uc_tb *tb = va_arg(args, uc_tb *);
            err = uc->uc_gen_tb(uc, addr, tb);
        } else {
            err = UC_ERR_ARG;
        }
        break;
    }
```
```
    // Now we clear cache for all TBs.
    for (int i = 0; i < TB_COUNT; i++) {
        err = uc_ctl_remove_cache(uc, ADDRESS + i * TCG_MAX_INSNS,
                                  ADDRESS + i * TCG_MAX_INSNS + 1);
        if (err) {
            printf("Failed on uc_ctl() with error returned: %u\n", err);
            return;
        }
    }
```
```
    case UC_CTL_TB_REMOVE_CACHE: {

        UC_INIT(uc);

        if (rw == UC_CTL_IO_WRITE) {
[...]

PR #1685, opened 2022-08-15
Fix for test_uc_ctl_tb_cache. (#1683)

relapids:
This fixes incorrect argument types passed to uc_ctl_request_cache/uc_ctl_remove_cache in test_uc_ctl_tb_cache (specifically the one in sample_ctl - the one in test_ctl does not have the same issue). See #1683 for details.


wtdcode:
Thanks!

Issue #1680, opened 2022-08-15
Memory leak in PPC target


relapids:
Steps to reproduce: Build with LeakSanitizer (enabled by default with AddressSanitizer on Linux) and run the tests/samples.

Example build steps:
```
$ cmake -S ../repo/ -B . -G Ninja -D CMAKE_C_COMPILER=clang -D CMAKE_C_FLAGS=-fsanitize=address -D CMAKE_BUILD_TYPE=Debug
$ cmake --build .
```

Test output for test_ppc32_add:
```
Test test_ppc32_add...                          [ OK ]

=================================================================
==7079==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 3072 byte(s) in 1 object(s) allocated from:
    #0 0x557cedfbc3e1 in __interceptor_calloc (/home/builder/unicorn/build/test_ppc+0xdb3e1)
    #1 0x7f1c603e0ec4 in g_malloc0 /home/builder/unicorn/repo/glib_compat/gmem.c:139:15
    #2 0x7f1c603e0f74 in g_malloc0_n /home/builder/unicorn/repo/glib_compat/gmem.c:168:12
    #3 0x7f1c627465bc in init_ppc_proc /home/builder/unicorn/repo/qemu/target/ppc/translate_init.inc.c:9258:29
    #4 0x7f1c62720887 in ppc_cpu_realize /home/builder/unicorn/repo/qemu/target/ppc/translate_init.inc.c:9922:5
    #5 0x7f1c6271fc02 in cpu_ppc_init_ppc /home/builder/unicorn/repo/qemu/target/ppc/translate_init.inc.c:11191:5
    #6 0x7f1c62522278 in ppc_cpus_init /home/builder/unicorn/repo/qemu/target/ppc/unicorn.c:403:11
    #7 0x7f1c603d3b73 in machine_initialize /home/builder/unicorn/repo/qemu/softmmu/vl.c:56:12
    #8 0x7f1c603c48c0 in uc_init /home/builder/unicorn/repo/uc.c:228:9
    #9 0x7f1c603c8c4e in uc_mem_map /home/builder/unicorn/repo/uc.c:1042:5
    #10 0x557cee00750c in uc_common_setup /home/builder/unicorn/repo/tests/unit/test_ppc.c:10:5
    #11 0x557cee0049e7 in test_ppc32_add /home/builder/unicorn/repo/tests/unit/test_ppc.c:20:5
[...]

PR #1686, opened 2022-08-16
Fix memory leak in PPC target. (#1680)

relapids:
This frees the TLB buffers on uc_close. See #1680 for details. Test (test_ppc) passes cleanly with LeakSanitizer enabled after this change.


wtdcode:
Looks good to me, thanks!

Issue #1689, opened 2022-08-16
Test failures if certain architecture is not enabled


relapids:
Not sure if you consider this an issue or not, but I think some tests should perhaps be guarded by something like #ifdef UNICORN_HAS_ARM since test_ctl is included in the build by default (regardless of what you set for UNICORN_ARCH), but if for example you only have x86 enabled the tests won't work since they rely on having the arm architecture available.

Steps to reproduce:
```
PS E:\Dev\unicorn\build> cmake -S ..\repo\ -B . -G Ninja -D CMAKE_C_COMPILER=cl -D CMAKE_BUILD_TYPE=Debug -D UNICORN_ARCH="x86"
PS E:\Dev\unicorn\build> cmake --build .
PS E:\Dev\unicorn\build> cmake --build . --target test
```

Test output:
```
[0/1] Running tests...
Test project E:/Dev/unicorn/build
    Start 1: test_x86
1/3 Test #1: test_x86 .........................   Passed    1.65 sec
    Start 2: test_mem
2/3 Test #2: test_mem .........................   Passed    0.10 sec
    Start 3: test_ctl
3/3 Test #3: test_ctl .........................***Failed    0.06 sec

67% tests passed, 1 tests failed out of 3

Total Test time (real) =   1.81 sec

The following tests FAILED:
          3 - test_ctl (Failed)
```
```
PS E:\Dev\unicorn\build> .\test_ctl.exe
Test test_uc_ctl_mode...                        [ OK ]
Test test_uc_ctl_page_size...                   [ OK ]
Test test_uc_ctl_arch...                        [ OK ]
Test test_uc_ctl_time_out...                    [ OK ]
Test test_uc_ctl_exits...                       [ OK ]
Test test_uc_ctl_tb_cache...                    [ OK ]
Test test_uc_ctl_change_page_size...            [ FAILED ]
  test_ctl.c:183: Check __err == UC_ERR_OK... failed
    Invalid/unsupported architecture (UC_ERR_ARCH)
  test_ctl.c:184: Check __err == UC_ERR_OK... failed
    Invalid/unsupported architecture (UC_ERR_ARCH)
  Unhandled SEH exception... failed
    Exception code:    0xc0000005
    Exception address: 0x00007FFB19DC7036
  Access violation.
Test test_uc_ctl_arm_cpu...                     [ FAILED ]
  test_ctl.c:201: Check __err == UC_ERR_OK... failed
[...]

PR #1691, opened 2022-08-19
Disable ARM-specific tests when ARM is unavailable. (#1689)

relapids:
When ARM target is not enabled, disable the ARM-specific tests in the test suites which are included unconditionally. See #1689 for details.


wtdcode:
Thanks!

Issue #1692, opened 2022-08-19
MSVC runtime library handling in CMakeLists is broken


relapids:
The handling of the MSVC runtime library in the CMakeLists doesn't appear to work. The problem is that the logic to set the cmake_policy to >= 3.15 under MSVC to get the CMAKE_MSVC_RUNTIME_LIBRARY support is guarded behind an "if(MSVC)" check, but that variable won't be defined until after compiler detection occurs, which hasn't happened yet.

At first I was hoping I could simply call `enable_language(C)` to trigger the compiler identification immediately after the `cmake_minimum_required` call, but as it turns out (per the documentation for CMP0091), the policy we're trying to configure takes effect with either the first call to `project` or `enable_language`. So, it seems this approach is doomed.

Probably the best approach would be to do the version check and set the relevant policies if CMake >= 3.15, and then after the call to project, do the "if(MSVC)" call and check the version again, and error out if it's < 3.15. That way none of the existing (intended) behavior changes, and we don't need to raise the minimum required version globally or require people to change how they configure the build.

This seems to have regressed after #1584 was merged. @shuffle2 FYI in case you can think of a better fix.

PR #1693, opened 2022-08-19
Fix MSVC runtime library handling in CMakeLists. (#1692)

relapids:
The current runtime library handling never works. This PR fixes it to work as intended. See #1692 for details.

A cleaner approach would probably be to just raise the minimum required CMake version globally, but I'm not sure what the 'lowest common denominator' OS you support is, so it seemed safer to just do it this way.


wtdcode:
Looks good, sorry for the late reply.

PR #1710, opened 2022-09-24
Fix big endian issues

roehling:
This PR fixes a few test failures if the host architecture is Big Endian:
* On BE architectures, truncating an integer type will change its value, because the lower bits will be cut off, not the higher ones.  Most notably, the program counter in `uc_emu_start` will end up zeroed. Luckily, the existing `set_pc` function handles all corner cases correctly.
*  Reading guest memory leaks the endianness to the host, so if the host has a different endianness, this needs to be compensated for. I added a bunch of `cpu_to_le` and `le_to_cpu` functions (which are no-ops on LE hosts) for this.


roehling:
After applying this PR, [these two checks](https://github.com/unicorn-engine/unicorn/blob/6c1cbef6ac505d355033aef1176b684d02e1eb3a/tests/unit/test_x86.c#L175-L176) will continue to fail on BE hosts, and I am not entirely sure what the problem is. I disabled these two checks in Debian for now, but I hope someone can come up with an actual fix.

wtdcode:
I failed to understand the context here, anyway I can reproduce this?

roehling:
The test failures happen if you run UC on a host with big endian architecture, such as PowerPC, S390X, etc.
For example, see https://buildd.debian.org/status/fetch.php?pkg=unicorn-engine&arch=s390x&ver=2.0.0-1&stamp=1662876884&raw=0

tests/unit/unicorn_test.h
```diff
@@ -4,6 +4,7 @@
 #include <stdio.h>
 #include <stdint.h>
 #include <unicorn/unicorn.h>
+#include "qemu/bswap.h"
```
wtdcode, line 7: This is wrong since we don't import qemu things in test files as they are not linked against the test targets, though it may work...

qemu/target/ppc/translate/vsx-impl.inc.c
```diff
@@ -1603,7 +1603,7 @@ static void gen_xxspltw(DisasContext *ctx)
     tofs = vsr_full_offset(rt);
     bofs = vsr_full_offset(rb);
     bofs += uim << MO_32;
-#ifndef HOST_WORDS_BIG_ENDIAN
+#ifndef HOST_WORDS_BIGENDIAN
```
wtdcode, line 1606: This fix is good!

uc.c
```diff
@@ -753,82 +753,7 @@ uc_err uc_emu_start(uc_engine *uc, uint64_t begin, uint64_t until,
...
-        break;
-#endif
-#ifdef UNICORN_HAS_TRICORE
-    case UC_ARCH_TRICORE:
-        uc_reg_write(uc, UC_TRICORE_REG_PC, &begin);
-        break;
-#endif
-    }
-
+    uc->set_pc(uc, begin);
```
wtdcode, line 756: These two functions are not exactly the same unfortunately. You have to add:

```C
                // force to quit execution and flush TB
                uc->quit_request = true;
                uc_emu_stop(uc);
```

tests/unit/test_arm.c
```diff
@@ -164,9 +164,9 @@ static void test_arm_thumb_ite(void)
     OK(uc_reg_write(uc, UC_ARM_REG_R3, &r_r3));
 
     OK(uc_mem_map(uc, r_sp, 0x1000, UC_PROT_ALL));
-    r_r2 = 0x68;
+    r_r2 = cpu_to_le32(0x68);
```
wtdcode, line 167: A good way I think is to detect the host endianness directly and do this like:

```
#ifdef BIG_ENDIAN
r_r2 = bswap32(0x68)
#else
r_r2 = 0x68
#endif
```

tests/unit/test_x86.c
```diff
@@ -172,8 +172,11 @@ static void test_x86_inc_dec_pxor(void)
 
     TEST_CHECK(r_ecx == 0x1235);
     TEST_CHECK(r_edx == 0x788f);
+#ifndef HOST_WORDS_BIGENDIAN
```
wtdcode, line 175: Note again we don't use QEMU defines in the test files.

wtdcode:
Fixed as of a40bf26263f95e37a157d3db7bc035f2b6ad4fa8

Issue #1678, opened 2022-08-15
Test failures on M1 Mac


relapids:
Hi,

I am trying to use Unicorn on an M1 (arm64) Mac running macOS Monterey 12.5 and am running into some test failures. Unfortunately I'm not yet very familiar with the internals of the project so I can't speculate on why this might be. 

No special steps to reproduce afaik. Checked out and built master branch without any special flags other than setting Debug build type (but other build types fail in the same way).

I tried checking the Github Actions to see how this compared to the CI builds but it seems there are only x64 MacOS builds in the CI.

Steps I took and some basic debug info shown below. Let me know if there's any more information you'd like.

Thanks for providing this useful project.

Note: Build/configure output not shown for brevity, but I can provide it if it's relevant.
```
relapids@Relapids-Mini unicorn % git clone https://github.com/unicorn-engine/unicorn repo
relapids@Relapids-Mini unicorn % mkdir build && cd build
relapids@Relapids-Mini build % cmake -S ../repo -B . -G Ninja -D CMAKE_BUILD_TYPE=Debug && cmake --build .
```

```
relapids@Relapids-Mini build % cmake --build . --target test
[0/1] Running tests...
Test project /Users/relapids/dev/unicorn/build
      Start  1: test_x86
 1/12 Test  #1: test_x86 .........................***Failed    1.18 sec
      Start  2: test_arm
 2/12 Test  #2: test_arm .........................   Passed    0.29 sec
      Start  3: test_arm64
 3/12 Test  #3: test_arm64 .......................   Passed    0.14 sec
      Start  4: test_m68k
 4/12 Test  #4: test_m68k ........................   Passed    0.13 sec
      Start  5: test_mips
 5/12 Test  #5: test_mips ........................   Passed    0.13 sec
      Start  6: test_sparc
 6/12 Test  #6: test_sparc .......................   Passed    0.11 sec
      Start  7: test_ppc
 7/12 Test  #7: test_ppc .........................   Passed    0.12 sec
      Start  8: test_riscv
 8/12 Test  #8: test_riscv .......................   Passed    0.14 sec
      Start  9: test_s390x
 9/12 Test  #9: test_s390x .......................   Passed    0.14 sec
      Start 10: test_tricore
10/12 Test #10: test_tricore .....................   Passed    0.11 sec
      Start 11: test_mem
11/12 Test #11: test_mem .........................***Failed    0.14 sec
      Start 12: test_ctl
[...]

PR #1711, opened 2022-09-24
Adjust big memory test for host pagesize

roehling:
On machines with a page size larger than 4K, the requested memory size in `test_map_big_memory` gets rounded up and overflows to zero.

This PR adds some code to query the page size and adjust the requested memory size accordingly. See also the discussion in #1678 and thanks to @relapids for the initial debugging, which helped me resolve this issue for Debian on mipsel and s390x much more quickly.


wtdcode (approved):
Looks good to me, waiting for approval

PR #1715, opened 2022-09-25
Fix test_uc_hook_cached_uaf for MacOS M1 (aarch64).

relapids:
Apple Silicon does not allow RWX pages so the mmap call was returning -1. This changes the test to map as RW and then mprotect it back and forth between RX/RW as needed.


wtdcode:
LGTM, thanks!

Issue #1718, opened 2022-09-27
Check Unicorn memory regions after address translation


imre-kis-arm:
**Description**
While emulating a binary that configures and enables the MMU both physical and virtual address ranges has to be added to the Unicorn instance beforehand. Unicorn will never access the added virtual range but it will do the address translation and as expected it accesses the memory area at the physical address range. However it will crash with an unmapped memory error if the virtual address range is not added to the unicorn instance.
What I see as a problem is that [load_helper](https://github.com/unicorn-engine/unicorn/blob/6c1cbef6ac505d355033aef1176b684d02e1eb3a/qemu/accel/tcg/cputlb.c#L1418) and [store_helper](https://github.com/unicorn-engine/unicorn/blob/6c1cbef6ac505d355033aef1176b684d02e1eb3a/qemu/accel/tcg/cputlb.c#L1965) functions do the Unicorn memory region lookup first and then do the address translation after it. This works fine if the MMU is disabled as the address translator will simply return the original address. However with the MMU enabled Unicorn won't find the address range for the virtual address before translation but it would find it after translating it to a physical address.
I've observed the issue with ARM and ARM64 architectures but I can imagine it affects others too.

I think these is the same issue: #1602

Tested with Unicorn 2.0.0 and aarch64-linux-gnu-gcc 12.2.0

**Example**
I've prepared a simple ARM64 example for demonstrating the issue. I've also attached the complete test environment.
[unicorn_mmu_bug.zip](https://github.com/unicorn-engine/unicorn/files/9656283/unicorn_mmu_bug.zip)

**test_mmu.s**
- The code starts with the MMU being turned off
- We load the binary to the `0x0000_0000` address
- We read data from `0x4000_0000`, note that this is now a physical address
- Enable MMU with the mapping below
  - `0x0000_0000- 0x3fff_ffff` -> `0x0000_0000 - 0x3fff_ffff`
  - `0x4000_0000- 0x7fff_ffff` -> `0x4000_0000 - 0x7fff_ffff`
  - `0x8000_0000- 0xbfff_ffff` -> `0x4000_0000 - 0x7fff_ffff`
  - `0xc000_0000- 0xffff_ffff` -> `0x4000_0000 - 0x7fff_ffff`
- With this MMU setting we can read the `0x4000_0000` address by reading `0x4000_0000`, `0x8000_0000` or `0xc000_0000`.
[...]

PR #1720, opened 2022-10-05
make unicorn use the physical addresses

PhilippTakacs:
This allows to use unicorn for full system emulation. The emulated code can completely use the MMU and change the page tables.

For better use there is also an hack to emulate non full system programs on x86. This more or less map virtual addresses directly to the physical addresses when the MMU is disabled. I'm not sure if such hacks are also required for other platforms.


wtdcode:
1. The target branch is `dev` branch, please make the pr mode readable.
2. I can't understand the changes of 60ccb0d248e1b10202eb8e297a0087f45b20a0ac . Note you are changing some core features of Unicorn, please add some tests/samples to illustrate better.
3. The commit 24e9e3c5954a37ebca22c133df931124376e2d61 is too hacky. At least you should support all architectures unicorn currently offers. But still I don't think it's a __correct__ way to do so. I have a copy of integrating qemu usermode with Unicorn but need more time to publish and you might wait for it.

PhilippTakacs:
> I can't understand the changes of https://github.com/unicorn-engine/unicorn/commit/60ccb0d248e1b10202eb8e297a0087f45b20a0ac

The basic changes are add the physical address to the tlb entry and move the memory hooks after the tlb lookup.

I'll add some tests. I might copy the PoC in #1718.

> The commit https://github.com/unicorn-engine/unicorn/commit/24e9e3c5954a37ebca22c133df931124376e2d61 is too hacky.

This changes only under which condition the currently implemented hack is used. Before this change the TLB lookups are done but ignored so couldn't use the MMU.

> At least you should support all architectures unicorn currently offers.

I don't think such hacks are required for other architectures, because there MMUs are not so complex as the x86/amd64 MMU. I currently looking at the code and the definitions of the other architectures.

> But still I don't think it's a correct way to do so. I have a fork of integrating qemu usermode with Unicorn but need more time to polish and you might wait for it.

A simpler way would be to implement a virtual TLB. So you could choose between the TLB implementation of the architecture and this virtual TLB. The virtual TLB defaults to a 1:1 mapping but allow to change entries.

PhilippTakacs:
First test for arm64 is there more will come.

for the hacks: I haven't seen any additional points for hacks yet. But I'm not finished with the checks.

PhilippTakacs:
The amd64 test illustrate why I would like this change (and later a virtual TLB). To emulate multiple processes and shared memory it's necessary to have more access to the TLB.

I'm currently searching, if there are other places where register writes have side effects.

PhilippTakacs:
Hi

I have found that other also might need some changes that you can control the special registers over unicorn, but I think this is a different issue.

I believe my patch works on all architectures and current work on tests for the other architectures.

It would be nice to get some more feedback on this PR.

wtdcode:
> Hi
> 
> I have found that other also might need some changes that you can control the special registers over unicorn, but I think this is a different issue.
> 
> I believe my patch works on all architectures and current work on tests for the other architectures.
> 
> It would be nice to get some more feedback on this PR.

Thanks for your work and sorry I'm still busy these days. I think i could review the whole thing next week or so.

qemu/target/i386/excp_helper.c
```diff
@@ -635,19 +635,19 @@ static int handle_mmu_fault(CPUState *cs, vaddr addr, int size,
     /* align to page_size */
     pte &= PG_ADDRESS_MASK & ~(page_size - 1);
     page_offset = addr & (page_size - 1);
-    paddr = get_hphys(cs, pte + page_offset, is_write1, &prot);
+    /* HACK allow full 64 bit mapping in u64 without paging */
+    if (env->cr[0] & CR0_PG_MASK) {
+        paddr = get_hphys(cs, pte + page_offset, is_write1, &prot);
+    } else {
+        paddr = addr;
```
wtdcode, line 642: It looks like not congruent with the previous logic? At least we should have `paddr = vaddr = addr & TARGET_PAGE_MASK`.

qemu/accel/tcg/cputlb.c
```diff
@@ -1436,7 +1438,28 @@ load_helper(CPUArchState *env, target_ulong addr, TCGMemOpIdx oi,
     bool handled;
     HOOK_FOREACH_VAR_DECLARE;
     struct uc_struct *uc = env->uc;
-    MemoryRegion *mr = memory_mapping(uc, addr);
+
+    /* Handle CPU specific unaligned behaviour */
+    if (addr & ((1 << a_bits) - 1)) {
```
wtdcode, line 1443: Any reason to bring this forward?

wtdcode (commented):
I start to understand your motivation and I like the current work! See my comments above.

qemu/target/i386/excp_helper.c
PhilippTakacs: This is already done direct before the tlb_set_page_with_attrs() call (line 648). But if you like it better I can add this.

qemu/include/exec/cpu-defs.h
```diff
@@ -77,9 +77,9 @@ typedef uint64_t target_ulong;
 #define CPU_VTLB_SIZE 8
 
 #if HOST_LONG_BITS == 32 && TARGET_LONG_BITS == 32
-#define CPU_TLB_ENTRY_BITS 4
-#else
 #define CPU_TLB_ENTRY_BITS 5
+#else
+#define CPU_TLB_ENTRY_BITS 6
```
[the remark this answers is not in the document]
PhilippTakacs: I have add the `paddr` to the `struct CPUTLBEntry`, which has a `padding to get a power of two size`. To get this padding right this has to be increased.

tests/regress/arm64_mmu.c
```diff
@@ -0,0 +1,65 @@
+#include "unicorn/unicorn.h"
```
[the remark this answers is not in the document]
PhilippTakacs: I have moved the tests, a sample will come

qemu/accel/tcg/cputlb.c
PhilippTakacs: I have split up the definition and the initialization, if you meant something else please explain.

PhilippTakacs:
I have noticed, that `uc_ctl_flush_tlb` does only clear the tb mappings and not the tlb. Should I create an extra pr for this or add a fix to this pr?

tests/regress/arm64_mmu.c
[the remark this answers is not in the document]
PhilippTakacs: I have add some simple sample code. Do you want more tests or sample code?

PhilippTakacs:
Hi

Just wanted to ask if you still miss some changes/fixes or you haven't had time to look again at this pr?

wtdcode:
> Hi
> 
> Just wanted to ask if you still miss some changes/fixes or you haven't had time to look again at this pr?

Sorry for late. I think I could find some time next week.

PhilippTakacs:
Just wanted to ask when you have time to look again at this pr?

wtdcode:
> Just wanted to ask when you have time to look again at this pr?

Sorry I'm just too busy recently but I will try to get it done these days.

wtdcode:
@PhilippTakacs Any reason for the force push?

PhilippTakacs:
> @PhilippTakacs Any reason for the force push?

I have cleaned up some conflicts. These conflicts are not related to this PR. I believe it is better to rebase the PR to current dev. This keeps the changes clean from other changes and makes the git history better to understand.

PhilippTakacs:
close for #1746

PR #1722, opened 2022-10-14
aarch64: Add FPCR and FPSR registers

TSRBerry:
This small PR adds support for the FPCR and FPSR registers for aarch64.

I'm still new to unicorn and trying to learn how it works, so I hope I didn't miss anything. These changes are based on an old commit by merryhime from 2018 and I just adjusted it to work with the current code base.

Currently projects like [Ryujinx](https://github.com/Ryujinx/Ryujinx) need to patch unicorn to support them for their tests, but I think integrating them into unicorn makes a lot more sense, than maintaining a patch. Especially when it's this small and might also benefit some other projects.


include/unicorn/arm64.h
```diff
@@ -313,6 +313,10 @@ typedef enum uc_arm64_reg {
 
     UC_ARM64_REG_PSTATE,
 
+    //> floating point control and status registers
```
wtdcode, line 316: This will change the values of other enums which breaks compatibility. Please add the registers in the endings.

TSRBerry:
Should be better now

wtdcode:
Thanks and welcome!

PR #1728, opened 2022-10-31
Add github workflow to publish Nuget for .NET bindings

TSRBerry:
This PR makes a few changes to the dotnet bindings, so it can be packaged as a Nuget.
Additionally it adds a new workflow to automatically publish the .NET bindings as a Nuget package if they changed in `dev` or `master`.

Packages created from `dev` get a suffix, this way they are marked as prereleases and don't change the latest release.

At the moment this workflow only publishes to Github Packages, but this could be changed to also publish to Nuget.org if you want.
You can read more about the necessary steps [here](https://learn.microsoft.com/en-us/nuget/nuget-org/publish-a-package).

To actually make the workflow work a secret [API key](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-nuget-registry#authenticating-to-github-packages) (`GHPR_TOKEN`) needs to be added. This could be a PAT or the access token of a Github app.

Since this Nuget package will only include the bindings and not the actual binaries, I have a few questions I'd like to ask:

- Should the binaries be included here as well, or should a separate Nuget package be used to include them? A separate package could also declare a dependency on the first one.
- If you want to include the binaries somehow would you prefer one Nuget package for all architectures, or separate packages for each one? This might be a lot harder to do, but also sounds like the best way to reduce file/download size.

Nuget people seem to be happy as long as we only include [one assembly](https://learn.microsoft.com/en-us/nuget/create-packages/select-assemblies-referenced-by-projects#recommended-one-assembly-per-package) in each package.

### DIFF
--- a/.github/workflows/Nuget-publishing.yml
+++ b/.github/workflows/Nuget-publishing.yml
@@ -105,7 +105,7 @@ jobs:
               fs.writeFileSync(`/home/runner/${artifact.name}.zip`, Buffer.from(download.data));
             
               console.log(`Unzipping: /home/runner/${artifact.name}.zip`);            
-              await exec.exec("unzip", [`/home/runner/${artifact.name}.zip`]);
+              await exec.exec("unzip", [`/home/runner/${artifact.name}.zip`, "-d", "/home/runner/"]);
               console.log(`Extracting library from 7z file to: ${destDir}/${sourceFile}`);
               await exec.exec("7z", ["e", `-o${destDir}/`, `/home/runner/${artifact.name}`, `${sourceDir}${sourceFile}`], options);
               if (sourceFile != destFile) {
--- a/.github/workflows/Nuget-publishing.yml
+++ b/.github/workflows/Nuget-publishing.yml
@@ -45,6 +45,7 @@ jobs:
             let sourceFile = "";
             let destDir = "";
             let destFile = "";
+            var output = {};
             for (const artifact of allArtifacts.data.artifacts) {
               switch(artifact.name) {
                 case 'ubuntu-cmake-aarch64.7z':
@@ -109,7 +110,9 @@ jobs:
               console.log(`Extracting library from 7z file to: ${destDir}/${sourceFile}`);
               await exec.exec("7z", ["e", `-o${destDir}/`, `/home/runner/${artifact.name}`, `${sourceDir}${sourceFile}`], options);
               if (sourceFile != destFile) {
-                console.log(`Renaming library to: ${destFile}`);
+                output = await exec.getExecOutput("ls", [destDir], options);
+                sourceFile = output.stdout.trim();
+                console.log(`Renaming ${sourceFile} to ${destFile}`);
                 await exec.exec("mv", [`${destDir}/${sourceFile}`, `${destDir}/${destFile}`], options);
               }
             }
--- a/.github/workflows/Nuget-publishing.yml
+++ b/.github/workflows/Nuget-publishing.yml
@@ -102,12 +102,12 @@ jobs:
                artifact_id: artifact.id,
                archive_format: 'zip',
               });
-              fs.writeFileSync(`/tmp/${artifact.name}.zip`, Buffer.from(download.data));
+              fs.writeFileSync(`/home/runner/${artifact.name}.zip`, Buffer.from(download.data));
             
-              console.log(`Unzipping: /tmp/${artifact.name}.zip`);            
-              await exec.exec("unzip", [`/tmp/${artifact.name}.zip`]);
-              console.log(`Extracting library from 7z file to: ${destDir}${sourceFile}`);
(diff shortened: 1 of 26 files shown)


wtdcode:
Hey, sorry for the late reply. I'm too busy these days and have to review it a bit late.

The changes look good to me at first glance but I have to test the workflow before merging it.

For your questions:

> Should the binaries be included here as well, or should a separate Nuget package be used to include them? A separate package could also declare a dependency on the first one.

We would prefer distribution with binaries.

> If you want to include the binaries somehow would you prefer one Nuget package for all architectures, or separate packages for each one? This might be a lot harder to do, but also sounds like the best way to reduce file/download size.

One for all looks good to me. Is that also good for Nuget?

> At the moment this workflow only publishes to Github Packages, but this could be changed to also publish to Nuget.org if you want.

I prefer to publish to both but I'm not sure about the difference. I haven't written C# for at least 3 years ;)

TSRBerry:
> I prefer to publish to both but I'm not sure about the difference. 

Nuget.org requires a Microsoft account and another secret for the workflow. Aside from that Nuget.org is configured per default for .NET, so you don't need to configure anything to restore packages from there.

Github Packages will require people to configure a nuget source for github.
A recent example for this kind of configuration can be found here: [Farmtronics/nuget.config](https://github.com/JoeStrout/Farmtronics/blob/main/Farmtronics/nuget.config)
It uses the following Github package: [MiniScript](https://github.com/JoeStrout/miniscript/packages/1644230)

> One for all looks good to me. Is that also good for Nuget?

Sure! I'll get to work on that then, thanks!

TSRBerry:
If my workflow works correctly, all the libraries should be included in the Nuget package.

TSRBerry:
It finally works correctly! You can take a look here: https://github.com/TSRBerry/unicorn/actions/runs/3401102101/jobs/5655975469#step:7:7

Note: The workflow needs to be in master, since it will only run the version of the default branch when using the `workflow_run` event.

wtdcode:
Seems good! Will merge it after I manually test the CI. Please be a bit patient on it.

Thanks for your efforts!

.github/workflows/Nuget-publishing.yml
```diff
@@ -0,0 +1,141 @@
+name: Nuget 📦 Distribution
+
+on:
+  workflow_run:
+    workflows:
+      - "Build UC2"
```
wtdcode, line 6: I think we don't need to reply on this action because it's slow in most cases. You could just build binaries in this workflow though it might introduce some duplication.

TSRBerry: Wouldn't that take exactly as long as the workflow above?

wtdcode:
That worlflow is slow because of ppc runners, which you could skip in your workflow.

TSRBerry:
Now the workflow includes jobs to build all the native libraries before creating the Nuget package.

I'll be linking the result of the new workflow as soon as it finishes.

TSRBerry:
@wtdcode It's been some time now, could I ask for an update on this PR? Just want to make sure there are no other issues that need to be taken care of.

wtdcode:
> @wtdcode It's been some time now, could I ask for an update on this PR? Just want to make sure there are no other issues that need to be taken care of.

Not yet. Sorry too busy these days

wtdcode:
@TSRBerry The workfllow looks good to me. I'm checking with our team about the proper way to publish this.

TSRBerry:
Thank you very much!

wtdcode:
@TSRBerry 

The nuget package is at https://www.nuget.org/packages/UnicornEngine.Unicorn for a placeholder, could you:

1. Upload the nupkg files to artifacts for every run.
2. Write an extra step to publish to nuget, note we have `secrets.NUGET_KEY` for that.

TSRBerry:
Added everything!

On a side note, does this repo have Dependabot enabled? I noticed there might be a few Github Actions that could be outdated and Dependabot would make it a lot easier to keep these up to date.

wtdcode:
> Added everything!
> 
> On a side note, does this repo have Dependabot enabled? I noticed there might be a few Github Actions that could be outdated and Dependabot would make it a lot easier to keep these up to date.

Not yet, how to configure that and what's the benefits?

TSRBerry:
Benefits of staying up to date would be that actions don't stop working randomly, when github makes changes to actions again.

The usual benefits of staying up to date also apply of course, but for unicorn it's probably more important to ensure the workflows won't stop working one day.

wtdcode:
> Benefits of staying up to date would be that actions don't stop working randomly, when github makes changes to actions again.
> 
> The usual benefits of staying up to date also apply of course, but for unicorn it's probably more important to ensure the workflows won't stop working one day.

Seems not so urgent for now and I think we don't need it currently. I will try to make the CI work today.

wtdcode:
I test it again and it works well.

Merged, thanks :)

wtdcode:
It's up! Thanks for your work!

PR #1736, opened 2022-11-14
add riscv constants to python binding module

ks0777:
The riscv constants have not been added to the python binding module yet. They thus cannot be imported and used properly.


wtdcode:
Merged, thanks!

PR #1742, opened 2022-12-07
Remove usage of UC_INIT

ZehMatt:
In my opinion the code makes little sense the way it is, the state is only initialized once and the user is required to call uc_open in where we can do the initialization. There is currently nothing that would influence the initialization between uc_open and any subsequent API calls that had UC_INIT. This PR eliminates quite a few lines without actually changing anything API/ABI wise.

### DIFF
--- a/uc.c
+++ b/uc.c
@@ -189,14 +189,6 @@ bool uc_arch_supported(uc_arch arch)
     }
 }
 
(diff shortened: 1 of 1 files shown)


ZehMatt:
Looks like I jumped to a conclusion too early.. the tests fail because setting the mode after uc_open fails.

Issue #1704, opened 2022-09-11
Unicorn on Windows takes 1GB of RAM when just instantiating an Emulator and registering a hook


expend20:
Hello!
So, I just create an instance and register the hook without actually mapping any memory or executing single opcode, and it's already +1gb of ram on Windows. ~~With rougly the same code in C it's only 11mb, so I think the problem could be somewhere in Rust code.~~ UPD: it's actually not related to Rust bindings.

```rust
use unicorn_engine::unicorn_const as ucc;
use unicorn_engine::Unicorn;

fn eat_1gb() {
    let uni = Unicorn::new(ucc::Arch::X86, ucc::Mode::MODE_64);
    if uni.is_err() {
        println!("Unable to create unicorn instance");
        return;
    }
    let mut emu = uni.unwrap();
    let hook = emu.add_mem_hook(
        ucc::HookType::MEM_UNMAPPED,
        0,
        u64::MAX,
        |_uc, _access, _addr, _size, _value| {
            true
        },
    );
    std::thread::sleep(std::time::Duration::from_secs(1));
    println!("1GB allocated");

    emu.remove_hook(hook.unwrap()).unwrap();
}
fn main() {
    for i in 0..30 {
        eat_1gb();
        println!("Iteration {}, check ram usage...", i);
        // sleep for 1 second
        std::thread::sleep(std::time::Duration::from_secs(1));
        println!("1GB freed");
    }
}
```

Latest version is used:
```
[dependencies]
unicorn-engine = "2.0.0"
```
![vmconnect_0aEYJovtJI](https://user-images.githubusercontent.com/36543551/189547320-0294589e-e26c-4cd3-b801-ae2e5bb1c791.gif)

mrexodia:
Tested the C bindings and it reproduces with `2.0.0` (commit hash 6c1cbef6ac505d355033aef1176b684d02e1eb3a). It looks like there is a gigantic 1GB RWX page allocated.

expend20:
Oh, sorry for that, actually not a bindings issue. Let me rename the issue then.

PhilippTakacs:
This is the TCG buffer. Look at qemu/accel/tcg/translate-all.c

Not sure if this is a real issue, because the memory is only allocated and not used (not sure how windows behaves in this case).

wtdcode:
[...]

PR #1743, opened 2022-12-07
Allow user to set TCG buffer size

ZehMatt:
This partially solves the issue of #1704. The default of 1GiB makes way more sense when an actual VM is used but for emulating shellcode having 1GiB is overkill.

### DIFF
--- a/include/uc_priv.h
+++ b/include/uc_priv.h
@@ -315,6 +315,7 @@ struct uc_struct {
     /* qemu/exec-vary.c */
     TargetPageBits *init_target_page;
     int target_bits; // User defined page bits by uc_ctl
+    unsigned long tcg_buffer_size; // User defined tcg buffer size by uc_ctl
     int cpu_model;
     BounceBuffer bounce;                // qemu/cpu-exec.c
     volatile sig_atomic_t exit_request; // qemu/cpu-exec.c
--- a/include/unicorn/unicorn.h
+++ b/include/unicorn/unicorn.h
@@ -536,7 +536,11 @@ typedef enum uc_control_type {
     UC_CTL_TB_REMOVE_CACHE,
     // Invalidate all translation blocks.
     // No arguments.
-    UC_CTL_TB_FLUSH
+    UC_CTL_TB_FLUSH,
+    // Current buffer size for TCG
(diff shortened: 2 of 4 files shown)


wtdcode:
This patch has some slight flaw and I will work on it to fix.

wtdcode:
I will close this PR because I'm reluctant to add an API that only makes sense on a specific platform, which destroys the simplicity and brevity.

See [this](https://github.com/unicorn-engine/unicorn/issues/1704#issuecomment-1407488655) for the actual fix.

vrubleg:
It still could be useful for 32-bit platforms where address space is limited and it is not possible to reserve more than 2-4 GB of memory even when it is actually not used.

wtdcode:
> @wtdcode It still could be useful for 32-bit platforms where address space is limited and it is not possible to reserve more than 2-4 GB of memory even when it is actually not used.

Good point, I will rethink a bit.

wtdcode:
Also link to https://github.com/pwndbg/pwndbg/issues/1534

ZehMatt:
> This patch has some slight flaw and I will work on it to fix.

Could you tell me what the flaw is? I might be able to fix it myself

> I will close this PR because I'm reluctant to add an API that only makes sense on a specific platform, which destroys the simplicity and brevity.
> 
> See [this](https://github.com/unicorn-engine/unicorn/issues/1704#issuecomment-1407488655) for the actual fix.

This is quite an odd statement considering that when this specific config is not set absolutely nothing changes for the user, as for simplicity, you need to be more specific to how this adds complexity.

vrubleg:
Maybe it is better to disable this bot? Unmerged pull requests and non-closed issues don't hurt.

wtdcode:
> Maybe it is better to disable this bot? Unmerged pull requests and non-closed issues don't hurt.

Ok, will do it because my time is tight starting from this year so this bot is no more helpful ;(

wtdcode:
I hope this is finally solved by 3d5b2643f0af742d9b90b4511d0ee137775c8526, could you try this? @vrubleg

wtdcode:
I refine this and manually merge this to dev branch in 3145e3c4267ef287a780643d3ca0de70a173aa8c. Thanks!

PR #1746, opened 2022-12-14
Virtual tlb

PhilippTakacs:
depends on #1720

This allows to use the MMU from unicorn independent of the CPU architecture.

The API is currently not complete, because the glib version in unicorn miss some features.


wtdcode:
I have started to look into your two PRs, sorry for so late. Could you explain a bit about your use case of this PR?

PhilippTakacs:
We currently develop a fuzzer based on unicorn. To do this we emulate the OS on the host. Some problems we have is the memory management. We currently work with uc_mem_map_ptr() and make it mostly by hand. But the softmmu already has most of the features we need, but there is currently no interface to use it.

In #1720 I have add the possibility to use the mmu implementation of the emulated cpu. This allows full system emulation. This PR allows to manipulate the tlb out of unicorn, which we would like to use in our fuzzer.

wtdcode:
> We currently develop a fuzzer based on unicorn. To do this we emulate the OS on the host. Some problems we have is the memory management. We currently work with uc_mem_map_ptr() and make it mostly by hand. But the softmmu already has most of the features we need, but there is currently no interface to use it.
> 
> In #1720 I have add the possibility to use the mmu implementation of the emulated cpu. This allows full system emulation. This PR allows to manipulate the tlb out of unicorn, which we would like to use in our fuzzer.

We are working on providing two extra API: g2h and h2g which coverts guest memory address and vice versa. Is that what you need?

PhilippTakacs:
> We are working on providing two extra API: g2h and h2g which coverts guest memory address and vice versa. Is that what you need?

This is another feature we would like to have, but I don't think it's enough.

So a simple example with a linux binary: the emulated code uses mmap multible times to map some parts of a file to the memory. Later it uses munmap and mprotect on these mappings. This is based on what ld.so actually does. It is currently possible to do all this with unicorn, but we duplicate big parts of the memory management (i.e. split region). 

The softmmu is already implemented and called, but set up in a way you can't use it. The vaddr == paddr assumption is a sane default. But this limits a bit what you can do with unicorn and adds extra costs to some other use cases.

wtdcode:
> So a simple example with a linux binary: the emulated code uses mmap multible times to map some parts of a file to the memory. Later it uses munmap and mprotect on these mappings. This is based on what ld.so actually does. It is currently possible to do all this with unicorn, but we duplicate big parts of the memory management (i.e. split region).

So this could be solved by `uc_mem_map_ptr` which reuses the memory content I believe?

> 
> The softmmu is already implemented and called, but set up in a way you can't use it. The vaddr == paddr assumption is a sane default. But this limits a bit what you can do with unicorn and adds extra costs to some other use cases.

The problem is that, the API you added is making the API design complex, which we tried our best to avoid. But it might be make sense to expose the MMU to users and we are also designing about this.

PhilippTakacs:
> > So a simple example with a linux binary: the emulated code uses mmap multible times to map some parts of a file to the memory. Later it uses munmap and mprotect on these mappings. This is based on what ld.so actually does. It is currently possible to do all this with unicorn, but we duplicate big parts of the memory management (i.e. split region).
> 
> So this could be solved by `uc_mem_map_ptr` which reuses the memory content I believe?

Yes, but then you have to duplicate many parts of memory management of unicorn (and qemu). Which is the reason I started all this. Also I'm currently working on other improvements and features (i.e. split region with region manipulation instand of copy) where the ptr-mappings would benefit less from.

> > The softmmu is already implemented and called, but set up in a way you can't use it. The vaddr == paddr assumption is a sane default. But this limits a bit what you can do with unicorn and adds extra costs to some other use cases.
> 
> The problem is that, the API you added is making the API design complex, which we tried our best to avoid. But it might be make sense to expose the MMU to users and we are also designing about this.

I'm open to change the API. The simplest solution would be to add a callback. If you have other idea let me know.

PhilippTakacs:
I'm working on an alternative version based on a callback. This would still need the UC_CTL_TLB_TYPE, but instand of the uc_add_tlb_entry() api there would be only an extra callback type. Do you like this API better?

The code looks currently like this:
```
bool unicorn_fill_tlb(CPUState *cs, vaddr address, int size,
                      MMUAccessType rw, int mmu_idx,
                      bool probe, uintptr_t retaddr)
{
    bool handled = false;
    bool ret = false;
    struct uc_struct *uc = cs->uc;
    struct tlb_entry e;
    struct request req;

    HOOK_FOREACH(uc, hook, UC_HOOK_TLB_FILL) {
        if (hook->to_delete)
            continue;
        if (!HOOK_BOUND_CHECK(hook, vaddr))
            continue;
        if ((handled = ((uc_cb_tlbevent_t)hook->callback(uc, address, rw, &e, hook->user_data))))
            break;
    }      
           
    if (!handled) {
        e.paddr = address & TARGET_PAGE_MASK;
        e.size = TARGET_PAGE_SIZE;
        e.perms = UC_PROT_READ|UC_PROT_WRITE|UC_PROT_EXEC;
    }       
    MemoryRegion *mapping = find_memory_region(uc, e.paddr);
        
    if (!mapping) {
        goto tlb_miss;
    } 
        
    switch (rw) {
    case MMU_DATA_LOAD:
        ret = mapping->perms & e.perms & UC_PROT_READ;
        break;
    case MMU_DATA_STORE:
        ret = mapping->perms & e.perms & UC_PROT_WRITE;
        break;
    case MMU_INST_FETCH:
        ret = mapping->perms & e.perms & UC_PROT_EXEC;
        break;
    default: 
        //TODO UC_ERR_EXCEPTION
        ret = false;
        break;
    }

    if (ret) {
        tlb_set_page(cs, address & TARGET_PAGE_MASK, e.paddr & TARGET_PAGE_MASK, perms_to_prot(e.perms & mapping->perms), mmu_idx, TARGET_PAGE_SIZE);
        return true;
    }

tlb_miss:
    if (probe) {
        return false;
    }
    raise_mmu_exception(cs, address, rw, ret);
    return false;
}
```

wtdcode:
> > > So a simple example with a linux binary: the emulated code uses mmap multible times to map some parts of a file to the memory. Later it uses munmap and mprotect on these mappings. This is based on what ld.so actually does. It is currently possible to do all this with unicorn, but we duplicate big parts of the memory management (i.e. split region).
> > 
> > 
> > So this could be solved by `uc_mem_map_ptr` which reuses the memory content I believe?
> 
> Yes, but then you have to duplicate many parts of memory management of unicorn (and qemu). Which is the reason I started all this. Also I'm currently working on other improvements and features (i.e. split region with region manipulation instand of copy) where the ptr-mappings would benefit less from.

Wait, then why manipulating tlb would do the trick? I assume it also needs to manage memory by hand?

> I'm working on an alternative version based on a callback. This would still need the UC_CTL_TLB_TYPE, but instand of the uc_add_tlb_entry() api there would be only an extra callback type. Do you like this API better?

For this design, I assume it could be added to `uc_hook_add` seamlessly, note `uc_hook_add` supports variadic parameters.

PhilippTakacs:
The current version uses a callback.

Before you merge this, I would like to ask if you think is it reasonable to change the default to the virtual tlb and remove the hacks required to keep the vaddr = paddr assumption? The hacks I mean are the mem_redirect for mips and the hack in handle_mmu_fault for x86.

wtdcode:
> Before you merge this, I would like to ask if you think is it reasonable to change the default to the virtual tlb and remove the hacks required to keep the vaddr = paddr assumption? The hacks I mean are the mem_redirect for mips and the hack in handle_mmu_fault for x86.

I will prefer removing the hacks while enabling the flat memory model by default.

PhilippTakacs:
> > Before you merge this, I would like to ask if you think is it reasonable to change the default to the virtual tlb and remove the hacks required to keep the vaddr = paddr assumption? The hacks I mean are the mem_redirect for mips and the hack in handle_mmu_fault for x86.
> 
> I will prefer removing the hacks while enabling the flat memory model by default.

Ok, Im currenlty working on this. Also I make some cleanup, samples and tests for the vtlb functionality.

For the tests I have noticed again the `uc_ctl_flush_tlb` issue. The problem is that the `uc_ctr` define for this function is `UC_CTL_TB_FLUSH`. I don't know how to resolve this. easiest would be add a `tlb_flush(uc->cpu);` call to `uc_tb_flush`. Most correct would be to add an extra control `UC_CTL_TLB_FLUSH` and a define `uc_ctl_flush_tb`

PhilippTakacs:
Sorry for the force push, had to do some rework.

Now I think this can be merged.

wtdcode:
> Sorry for the force push, had to do some rework.
> 
> Now I think this can be merged.

ACK. I will review it this week.

qemu/accel/tcg/cputlb.c
```diff
@@ -1577,25 +1626,6 @@ load_helper(CPUArchState *env, target_ulong addr, TCGMemOpIdx oi,
         }
     }
 
-    /* Handle CPU specific unaligned behaviour */
-    if (addr & ((1 << a_bits) - 1)) {
```
wtdcode, line 1581: Any reason you have to move this part forward? Just for making the `address` parameter physical address?

qemu/accel/tcg/translate-all.c
```diff
@@ -1005,6 +1005,7 @@ static void tb_htable_init(struct uc_struct *uc)
 
 static void uc_tb_flush(struct uc_struct *uc) {
     tb_flush(uc->cpu);
+    tlb_flush(uc->cpu);
```
wtdcode, line 1008: Maybe another `UC_CTL_XXX`? The purpose of this `CTL` is slightly different I believe.

qemu/target/mips/unicorn.c
```diff
@@ -17,21 +17,6 @@ typedef uint32_t mipsreg_t;
 
 MIPSCPU *cpu_mips_init(struct uc_struct *uc);
 
-static uint64_t mips_mem_redirect(uint64_t address)
```
wtdcode, line 20: Loves this change ;)

qemu/softmmu/virtual_tlb.c
```diff
@@ -0,0 +1,106 @@
+#include <stdint.h>
```
wtdcode, line 1: Is this file from QEMU? If not, I tend to rename it to `unicorn.c`.

include/unicorn/unicorn.h
```diff
@@ -612,6 +640,7 @@ See sample_ctl.c for a detailed example.
 #define uc_ctl_request_cache(uc, address, tb)                                  \
     uc_ctl(uc, UC_CTL_READ_WRITE(UC_CTL_TB_REQUEST_CACHE, 2), (address), (tb))
 #define uc_ctl_flush_tlb(uc) uc_ctl(uc, UC_CTL_WRITE(UC_CTL_TB_FLUSH, 0))
+#define uc_ctl_tlb_mode(uc, mode) uc_ctl(uc, UC_CTL_WRITE(UC_CTL_TLB_TYPE, 1), (mode))
```
wtdcode, line 643: Could you illustrate more on this in this file?

qemu/softmmu/virtual_tlb.c
```diff
@@ -0,0 +1,106 @@
...
+    if (perms & UC_PROT_WRITE) {
+        ret |= PAGE_WRITE;
+    }
+    if (perms & UC_PROT_EXEC) {
+        ret |= PAGE_EXEC;
+    }
+    return ret;
+}
+
+bool unicorn_fill_tlb(CPUState *cs, vaddr address, int size,
```
wtdcode, line 46: If we go through this subroutine, we are bypassing all architectures' mmu implementation, correct?

wtdcode:
I think I need a bit more time to debug into your code before finishing the whole review, thanks for your nice and complete work!

qemu/accel/tcg/cputlb.c
PhilippTakacs: This is not so good shown in the diff, I moved the `find_memory_region` and the callback handling back after the mmu lookup. Yes this is to get the physical `address`.

The unaligned check and handling is moved forward because it is done before the mmu lookup.

qemu/accel/tcg/translate-all.c
PhilippTakacs: Haven't got feedback to my comment https://github.com/unicorn-engine/unicorn/pull/1746#issuecomment-1434660135 so I have implemented the lacy solution. I'll update this.

qemu/softmmu/virtual_tlb.c
PhilippTakacs: `unicorn.c` is a bit generic, how about `unicorn_vtlb.c`?

include/unicorn/unicorn.h
PhilippTakacs: Ok, in general I only add the necessary types and defines. To avoid to restructuring the complete file I moved some typedefs forword.

But step by step. First the two typedefs (uc_tlb_entry and uc_mem_type). They are necessary before the callback typedef, because the callback has these types as parameter. But I wanted to keep the structure of the file so I moved only the typedef and not the complete struct/enum definition.

Next the `uc_cb_tlbevent_t` typedef, a functionpointer type for the tlb callback. I hope the comment above is adequate.

Next remove the typedef around the `enum uc_mem_type`. The typedef is now a forword declaration in line 247.

Next add the callback enum for tlb lookups.

Next create a enum for the suported tlb modes. The mode can changed with `UC_CTL_TLB_TYPE`. `uc_ctl_tlb_mode` is a helper function for that.

qemu/softmmu/virtual_tlb.c
PhilippTakacs: Yes, or more correct we call this subroutine instand of the architectures' mmu implementation.

qemu/softmmu/virtual_tlb.c
wtdcode: yes, that's also fine

include/unicorn/unicorn.h
wtdcode: Sorry I didn't express myself clearly. I mean could you leave more inline comments here, mostly about usage and caveats?

PhilippTakacs:
I know you do this in your free time, but I wanted to ask when you have a bit time to finish the review?

qemu/softmmu/unicorn_vtlb.c
```diff
@@ -0,0 +1,106 @@
...
+        ret = e.perms & UC_PROT_READ;
+        break;
+    case MMU_DATA_STORE:
+        ret = e.perms & UC_PROT_WRITE;
+        break;
+    case MMU_INST_FETCH:
+        ret = e.perms & UC_PROT_EXEC;
+        break;
+    default:
+        //TODO UC_ERR_EXCEPTION
```
wtdcode, line 90: Setting `uc->invalid_error` and `uc->invalid_addr` and propagating is pretty enough.

qemu/unicorn_common.h
```diff
@@ -91,6 +94,19 @@ static inline void target_page_init(struct uc_struct* uc)
     uc->target_page_align = TARGET_PAGE_SIZE - 1;
 }
 
+static uc_err uc_set_tlb(struct uc_struct *uc, int mode) {
+    switch (mode) {
+        case UC_TLB_VIRTUAL:
```
wtdcode, line 99: I will prefer to setting `UC_TLB_CPU` by default since we want more compatibility for previous code, though they rely on the dirty or even wrong abstraction we provided before.

wtdcode:
Sorry for late, let's get it done this week.

btw, does this overlap with #1720?

qemu/softmmu/unicorn_vtlb.c
```diff
@@ -0,0 +1,106 @@
...
+        if (!HOOK_BOUND_CHECK(hook, address)) {
+            continue;
+        }
+        handled = true;
+        if ((ret = ((uc_cb_tlbevent_t)hook->callback)(uc, address & TARGET_PAGE_MASK, rw_to_mem_type(rw), &e, hook->user_data))) {
+            break;
+        }
+    }
+
+    if (handled && !ret) {
```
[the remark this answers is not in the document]
PhilippTakacs: To allow the user to raise a CPU exception and stop the emulation. When the user want to have default TLB entries (paddr = vaddr) he can remove (or not register) the callback for this page.

qemu/unicorn_common.h
PhilippTakacs: This function is the implementation for `UC_CTL_TLB_TYPE`.

`UC_TLB_CPU` is set as default in uc.c:235, There is a check to prevent overriding a previous setting of the tlb_type.

qemu/softmmu/unicorn_vtlb.c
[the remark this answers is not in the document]
wtdcode: What's the difference with stopping the emulation directly?

uc.c
```diff
@@ -230,6 +230,11 @@ static uc_err uc_init(uc_engine *uc)
         return UC_ERR_RESOURCE;
     }
 
+    // init tlb function
+    if (!uc->cpu->cc->tlb_fill) {
+        uc->set_tlb(uc, UC_TLB_VIRTUAL);
```
wtdcode, line 235: But you are setting `UC_TLB_VIRTUAL` as default, no?

wtdcode: Yes I understand your reasoning here. However, with MMU disabled, `vaddr=paddr` is satisfied in most cases, yes? If so, we can still preserve the `TLB_CPU` implementation by default.

qemu/softmmu/unicorn_vtlb.c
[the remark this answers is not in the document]
PhilippTakacs: This just goes to tlb_miss which calls (for non probe request) `raise_mmu_exception`. `raise_mmu_exception` stops the emulation. Not sure when `probe = true` is used, but I wanted to have this implemented in a way it's compatible with the qemu api.

uc.c
PhilippTakacs: > However, with MMU disabled, vaddr=paddr is satisfied in most cases, yes?

Most cases yes, but the biggest problem I see is x86_64. By default only a part off the address space is usable. But when you prefere TLB_CPU I can change it.

qemu/softmmu/unicorn_vtlb.c
PhilippTakacs: A thanks, will add.

PhilippTakacs:
> btw, does this overlap with #1720?

Yes, this is based on #1720.

wtdcode:
I tried both TLB implementations and they all failed the same Qiling tests. Hmmm, I will try to fix it today but not sure can get it done though.

wtdcode:
Solved and let me check this a bit more

wtdcode:
I'm done with this PR review, let's wait for CI!

PhilippTakacs:
Before you merge this I would like to clean up the history. Remove the fixup commits and update some commit messages.

wtdcode:
> Before you merge this I would like to clean up the history. Remove the fixup commits and update some commit messages.

I can squash to 1 single commit if you wish.

PhilippTakacs:
> I can squash to 1 single commit if you wish.

I would prefer to have multiple commits with meaningful and correct commit messages.

wtdcode:
> > I can squash to 1 single commit if you wish.
> 
> I would prefer to have multiple commits with meaningful and correct commit messages.

That's of course fine. Ping me once you are done. Besides, can we close #1720 if this gets merged?

PhilippTakacs:
> Besides, can we close https://github.com/unicorn-engine/unicorn/pull/1720 if this gets merged?

Yes, will close it.

PhilippTakacs:
> Ping me once you are done.

@wtdcode  I'm done

wtdcode:
Here you go. Really appreciate for your patience :).

mrexodia:
This change is breaking the following code:

```
self.regs.cs = windows_user_segment.cs
self.regs.ss = windows_user_segment.ss
self.regs.ds = windows_user_segment.ds
self.regs.es = windows_user_segment.es
self.regs.fs = windows_user_segment.fs
self.regs.gs = windows_user_segment.gs
self.regs.gs_base = self.teb
```

On `self.regs.fs = x` the whole process dies with exit code `-1`.

PR #1763, opened 2023-01-17
Fix Rust binding memory leak

LG3696:
Fix https://github.com/unicorn-engine/unicorn/issues/1619 by using Weak references.
This works without unsafe, so it should not have the soundness issue of https://github.com/unicorn-engine/unicorn/pull/1632


LG3696:
Sorry, I missed the info to pull-request on the dev branch. I created a new pull request against the dev branch.

PR #1764, opened 2023-01-18
Fix Rust binding memory leak

LG3696:
Fix https://github.com/unicorn-engine/unicorn/issues/1619 by using Weak references.
This works without unsafe, so it should not have the soundness issue of https://github.com/unicorn-engine/unicorn/pull/1632

### DIFF
--- a/bindings/rust/src/ffi.rs
+++ b/bindings/rust/src/ffi.rs
@@ -1,10 +1,11 @@
 #![allow(non_camel_case_types)]
 #![allow(dead_code)]
 
-use crate::Unicorn;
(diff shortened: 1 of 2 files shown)


wtdcode:
LGTM. Waiting for CI.

wtdcode:
The CI fails: https://github.com/unicorn-engine/unicorn/actions/runs/3939433382

LG3696:
It looks like insn_in_hook_proxy was missing the u32 return type and up until this change it happened to work by chance. I added the return type, so it should work robustly now.

wtdcode:
Merged and thanks!

PR #1765, opened 2023-01-18
Do not search for the RAMBlock in split_region

PhilippTakacs:
The MemoryRegion already contains a pointer to the RAMBlock. So it is not necesary to search for it.


wtdcode:
LGTM. Thanks!

PR #1785, opened 2023-02-22
Create c-cpp.yml

xets007:


wtdcode:
What's the meaning of this PR?

wtdcode:
Close due to not making any sense

PR #1790, opened 2023-02-26
[dotnet] Set @begin > @end address for hooks

tunz:
Begin address must be higher than End address to enable hooks for whole address space.


wtdcode:
LGTM.

wtdcode:
Thanks and welcome!

PR #1792, opened 2023-02-27
Use libunicorn.so.2 as fallback path

rhelmot:
This adds support for FreeBSD.


wtdcode:
Why not add a new key in `_lib`?

Ref: https://stackoverflow.com/questions/446209/possible-values-from-sys-platform

rhelmot:
That would work, but it remains that "libunicorn.so" is just straight up never correct in any situation, right?

wtdcode:
> That would work, but it remains that "libunicorn.so" is just straight up never correct in any situation, right?

Okay, make sense.

wtdcode:
But you should target dev branch ;)

PR #1793, opened 2023-03-01
[dotnet] Fix garbage collected hooks and wrong signature

tunz:
`*HookInternal` objects can be garbage collected, so their reference should be kept somewhere. With this PR, Unicorn object has `_hookInternals` member where all the hook internal objects are stored.
```
Process terminated. A callback was made on a garbage collected delegate of type 'UnicornEngine!UnicornEngine.MemReadHookInternal::Invoke'.
```

And, `MemReadHook` and `MemWriteHook` trampolines have a wrong callback signature. The second argument should be the event type.
https://github.com/unicorn-engine/unicorn/blob/7b8c63dfe650b5d4d2bf684526161971925e6350/qemu/accel/tcg/cputlb.c#L1469


wtdcode:
Thanks!

PR #1799, opened 2023-03-07
Remove the size limit for memory read and write

secretnonempty:
This change eliminate the maximum size restriction for uc_mem_read and uc_mem_write. Which is required to support applications, such as LLVM CFI, that map or unmap memory blocks with sizes equal to or greater than INT_MAX.

`uc_mem_unmap` and `uc_mem_protect` would need to split the memory if they only operate on a portion of a memory block. The splitting procedure can be broken into several steps as follows:
1. Reads the entire block of memory .
2. Unmap the whole memory block.
3. Remap the memory ranges that should not be unmapped.
4. Write the data back to memory.

If the memory block is equal or larger than INT_MAX, then the first step would fail, which would eventually make `uc_mem_unmap` or `uc_mem_protect` fail.

[linker_cfi.cpp](https://android.googlesource.com/platform/bionic/+/master/linker/linker_cfi.cpp#155) is a specific example that would use memory block with size equal to 2G(0x80000000 bytes) on a 64-bit platform. 

Removing the restriction would allow us to emulate these applications with less pain.


PhilippTakacs:
So looked at this a bit more and found #219 and #132 which explain the reasons. As far as I see this can changed now, but you missed some relevant points. For example `uc->read_mem`.

secretnonempty:
> Your change will make the situation worse. Your change will silently truncate the len to `MAX_RW_LENGTH`. So you still limited about the length but you are not informed about the fail.
>
Thank you for reviewing! I am confused with this, in my unit tests, this patch works as expected with memory blocks that are bigger than INT_MAX.

In the `uc_mem_read`
```c
    // memory area can overlap adjacent memory blocks                                                                                                                                                                
    while (count < size) {                                                                                                                                                                                           
        MemoryRegion *mr = memory_mapping(uc, address);                                                                                                                                                              
        if (mr) {                                                                                                                                                                                                    
            len = (size_t)MIN(size - count, mr->end - address);                                                                                                                                                      
            len = (size_t)MIN(len, MAX_RW_LENGTH);                                                                                                                                                                   
            if (uc->read_mem(&uc->address_space_memory, address, bytes, len) ==                                                                                                                                      
                false) {                                                                                                                                                                                             
                break;                                                                                                                                                                                               
            }                                                                                                                                                                                                        
            count += len;                                                                                                                                                                                            
            address += len;                                                                                                                                                                                          
            bytes += len;                                                                                                                                                                                            
        } else { // this address is not mapped in yet                                                                                                                                                                
            break;                                                                                                                                                                                                   
        }                                                                                                                                                                                                            
    }  
```
We can see if the memory block is bigger than INT_MAX, the loop will read multiple times until error occurs or count == size.

We can use the following code to test the patch:
```
#include <stdio.h>
#include <stdlib.h>
#include <assert.h>
#include <string.h>
#include <unicorn/unicorn.h>


int main(int argc, char *argv[]) {
  uc_engine * uc;
  uint64_t mem_addr = 0x1000000;
  size_t mem_size = 0x8f000000;
  size_t i;
  char buf[0x100];
  int len;
  uint8_t *pmem = NULL;

  uc_err err = uc_open(UC_ARCH_ARM64, UC_MODE_ARM, &uc);
  assert(err == UC_ERR_OK);

  err = uc_mem_map(uc, mem_addr, 0x8f000000, UC_PROT_ALL);
  assert(err == UC_ERR_OK);

  for (i = 0; i < mem_size; i += 0x1000) {
    len = snprintf(buf, sizeof(buf), "%#zx", i);
    assert(len > 0);
    err = uc_mem_write(uc, mem_addr + i, buf, len);
    assert(err == UC_ERR_OK);
  }

  pmem = calloc(1, mem_size);
  assert(pmem != NULL);

  err = uc_mem_read(uc, mem_addr, pmem, mem_size);
  assert(err == UC_ERR_OK);

  for (i = 0; i < mem_size; i += 0x1000) {
    len = snprintf(buf, sizeof(buf), "%#zx", i);
    if (memcmp(pmem + i, buf, len) != 0) {
      fprintf(stderr, "data does not match");
      abort();
    }
    printf("%s matches\n", buf);
  }

  free(pmem);
  uc_close(uc);

  puts("Okay, test passed!");
  return 0;
}
```
We can see the patch works as expected.

secretnonempty:
> So looked at this a bit more and found #219 and #132 which explain the reasons. As far as I see this can changed now, but you missed some relevant points. For example `uc->read_mem`.

I am not sure what I have missed, I am going to dive in the background. Thank you for the information!

secretnonempty:
I have updated the patch to utilize the update from Qemu, would very appreciate it if you can help review at your convenience. The latest patch works fine with the test program aforementioned.

uc.c
```diff
@@ -611,17 +607,13 @@ uc_err uc_mem_read(uc_engine *uc, uint64_t address, void *_bytes, size_t size)
 
 UNICORN_EXPORT
 uc_err uc_mem_write(uc_engine *uc, uint64_t address, const void *_bytes,
```
wtdcode, line 609: You forget to change the declaration in unicorn.h

wtdcode:
> I have updated the patch to utilize the update from Qemu, would very appreciate it if you can help review at your convenience. The latest patch works fine with the test program aforementioned.
> 
> Be aware that the original patch can work with Unicorn 1.x while the latest one can't.

Your changes are overall correct but this will involve more complex things as we are changing both API and ABI implicitly. Somethings you leave out:

1. Update declarations in the unicorn.h, I believe your compiler should have told you how to do it.
2. Update bindings. I could do this for you if you are not familiar with them but might be a bit slow.
3. Add a few tests about your changes in `test_mem`. Note you will have to also make sure no memory contents are truncated.

Also note that uc1 will only accept really severe security patches and these breaking changes won't go to uc1 so don't worry about uc1.

uc.c
secretnonempty: Thank you so much for pointing it out, because the `unicorn.h` is only used by program, and `size_t` is compatible with `uint64_t` on 64-bit platform, my compiler didn't complain about the inconsistency. My bad.

secretnonempty:
> > I have updated the patch to utilize the update from Qemu, would very appreciate it if you can help review at your convenience. The latest patch works fine with the test program aforementioned.
> > Be aware that the original patch can work with Unicorn 1.x while the latest one can't.
> 
> Your changes are overall correct but this will involve more complex things as we are changing both API and ABI implicitly. Somethings you leave out:
> 
> 1. Update declarations in the unicorn.h, I believe your compiler should have told you how to do it.
> 2. Update bindings. I could do this for you if you are not familiar with them but might be a bit slow.
> 3. Add a few tests about your changes in `test_mem`. Note you will have to also make sure no memory contents are truncated.
> 4. Targe dev branch.
> 
> Also note that uc1 will only accept really severe security patches and these breaking changes won't go to uc1 so don't worry about uc1.

Thank you, I think the commit now is ready for review. I have updated the language bindings. 

Most of the files under `bindings/dotnet/` directly have no substantial change except for  `bindings/dotnet/UnicornEngine/Binding/NativeBinding.fs`, please feel free to ignore them.

rliebig:
This MR has been stale since two years ago: is it psossible that it will be merged in the near future? 

If there is a chance that this MR will still be merged,  I would be available at least support it.

rliebig:
I was able to confirm that this indeed does fix larger writes: however a lot of the merge conflicts will be hard to fix. The changes can be ported to the current dev branch with ease, at least changes to the C files and python bindings. @wtcode, I would highly appreciate a short comment on the future support. IMHO it is a highly useful change and increases the amount of binaries supported massively.

wtdcode:
I have no idea why it was stale. Maybe my fault forgetting to review it.

But anyway, feel free to take it over and I will review it.

wtdcode:
I see, the reason is that it needs an minor bump for the API/ABI breakages while the latest such release is 2.1.x. At that point, I forgot this PR so we need 2.2.x for this to go. Fortunately, as #2102 suggests, the next release will be 2.2.0 and it’s time to revive this PR if you wish. @rliebig 

Link to #2102

rliebig:
I would very much like it to be in the next 2.2.0 release and will start work on it. If there is a specific timeline to be aware of, I would appreciate a heads up.

wtdcode:
> I would very much like it to be in the next 2.2.0 release and will start work on it. If there is a specific timeline to be aware of, I would appreciate a heads up.

I would expect 2.2.0 to out in June but no sure yet.

PR #1800, opened 2023-03-07
difference between stop_request and quit_request

PhilippTakacs:
quit_request is for internal use. This means the IP register was updated and qemu needs to rebuild the translation blocks.

stop_request is set by the user (uc_emu_stop) to indecate that unicorn sould stop emulating.

see #1579


wtdcode:
Merged, thanks!

PR #1802, opened 2023-03-11
[dotnet] Use native binding calls whenever possible

tunz:
- Use the native `uc_hook_del` call to delete hooks
- Use the native `uc_hook_add` call to add each hook. Now we can install hooks on different address ranges.
- Support hooks of specific address ranges for `AddEventMemHook`, `AddInHook`, `AddOutHook`, `AddSyscallHook`.

This PR does not change the existing interfaces, so users do not need to update their code.


wtdcode:
LGTM.

PR #1806, opened 2023-03-21
Cache index for find_memory_region

tunz:
`find_memory_region` uses a cache index to check whether the most recently used block is the target block, but the index is never set. It looks like it was accidentally removed in this PR: https://github.com/unicorn-engine/unicorn/pull/1414

The cache index is still effective even with the binary search.


wtdcode:
Nice findings, thanks!

PR #1809, opened 2023-03-23
Zig bindings - Generator

kassane:
All architectures were listed in Ziglang.

They are compatible with both zig stage 1 (0.10.x or lower) and stage2 (0.10.x or higher) backends.
**Note:** Zig version 0.9 and 0.10 have support for both backends, but 0.11 it will be stage2 only.

zig 0.10: stage2 (default), add build flag: `-fstage1` switch to stage1.
zig 0.9: stage1 (default), add build flag: `-fno-stage1` switch to stage2.


kassane:
> Amazing! So zig could interact with raw dynamic library api without further ffi wrapper?

Exactly, no need libC! cFFI in Zig is optional.

aquynh:
Please also update README with this new binding.

wtdcode:
Note you also need to target dev branch.

kassane:
Hi @aquynh 

> Interesting! What are you doing with this binding?

Plan to package unicorn for use with zig projects.
But, I realized that it will be complicated to port the scripts generated during the qemu build.

Draft and stdout:
https://gist.github.com/kassane/48c4fff6d95e5fa76d63b004f8f39b4c

> Please also update README with this new binding.

Ok

wtdcode:
Thanks and welcome!

PR #1812, opened 2023-03-30
add rust bindings for uc_ctl

ks0777:
This PR adds a wrapper for the uc_ctl function to unicorn's Rust bindings. Since Rust neither allows overloading of functions let alone variadic functions the implementation may seem a bit odd. I'm open to suggestions if you know of a better way to implement this. At the moment the uc_ctl function takes at most two optional arguments that are either a pointer or value that fits into 64-bit of memory. Consequently we need three functions in the Rust bindings to represent these cases. Since the uc_ctl functions does not define any type for the optional arguments it is hard to define a meaningful function signature with Rust being strongly typed. I defined the arguments as void pointers which means that arguments passed to this function first need to be unsafely transmuted. The other option would be to define a function for each argument count and type which seems even less appealing.

### DIFF
--- a/bindings/rust/src/lib.rs
+++ b/bindings/rust/src/lib.rs
@@ -885,6 +885,33 @@ impl<'a, D> Unicorn<'a, D> {
         }
     }
 
+    pub fn add_tlb_hook<F>(&mut self, begin: u64, end: u64, callback: F) -> Result<ffi::uc_hook, uc_error>
+    where
+        F: FnMut(&mut crate::Unicorn<D>, u64, MemType) -> Option<TlbEntry> + 'a,
+    {
+        let mut hook_ptr = core::ptr::null_mut();
+        let mut user_data = Box::new(ffi::UcHook {
+            callback,
+            uc: Rc::downgrade(&self.inner),
+        });
+        let err = unsafe {
+            ffi::uc_hook_add(self.get_handle(),
+                &mut hook_ptr,
+                HookType::TLB,
+                ffi::tlb_lookup_hook_proxy::<D, F> as _,
+                user_data.as_mut() as *mut _ as _,
+                begin,
+                end,
+            )
+        };
+        if err == uc_error::OK {
+            self.inner_mut().hooks.push((hook_ptr, user_data));
+            Ok(hook_ptr)
+        } else {
+            Err(err)
+        }
+    }
+
     /// Remove a hook.
     ///
     /// `hook` is the value returned by `add_*_hook` functions.
--- a/bindings/rust/src/lib.rs
+++ b/bindings/rust/src/lib.rs
@@ -1189,8 +1189,26 @@ impl<'a, D> Unicorn<'a, D> {
         }
     }
 
-    pub fn ctl_flush_tlb(&self) -> Result<(), uc_error> {
+    pub fn ctl_flush_tb(&self) -> Result<(), uc_error> {
         let err = unsafe { ffi::uc_ctl(self.get_handle(), UC_CTL_WRITE!(ControlType::UC_CTL_TB_FLUSH)) };
+        if err == uc_error::OK {
+            Ok(())
+        } else {
+            Err(err)
+        }
+    }
+
+    pub fn ctl_flush_tlb(&self) -> Result<(), uc_error> {
+        let err = unsafe { ffi::uc_ctl(self.get_handle(), UC_CTL_WRITE!(ControlType::UC_CTL_TLB_FLUSH)) };
+        if err == uc_error::OK {
+            Ok(())
+        } else {
+            Err(err)
+        }
+    }
+
+    pub fn ctl_tlb_type(&self, t: TlbType) -> Result<(), uc_error> {
+        let err = unsafe { ffi::uc_ctl(self.get_handle(), UC_CTL_WRITE!(ControlType::UC_CTL_TLB_TYPE), t as i32) };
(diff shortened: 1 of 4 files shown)


PhilippTakacs:
A better concept would be to add functions similar to the defines from `include/unicorn/unicorn.h` (line 623). This would be more work, but it provides better type safety and is easier to use.

ks0777:
I somehow managed to overlook these definitions, thanks for the suggestion! I implemented a function for each define from the header file.

bindings/rust/src/lib.rs
```diff
@@ -1064,4 +1065,154 @@ impl<'a, D> Unicorn<'a, D> {
...
+
+    pub fn ctl_get_exits(&self) -> Result<&[u64], uc_error> {
+        let exits_count: libc::size_t = self.ctl_get_exits_count()?;
+        let mut buffer = unsafe { libc::malloc(exits_count * 8) };
+        if buffer.is_null() {
+            return Err(uc_error::NOMEM);
+        }
+        let err = unsafe { ffi::uc_ctl(self.get_handle(), UC_CTL_UC_EXITS | UC_CTL_IO_READ, &mut buffer, exits_count) };
+        if err == uc_error::OK {
+            let slice: &[u64] = unsafe { std::slice::from_raw_parts(std::mem::transmute::<*mut c_void, *const u64>(buffer), exits_count) };
```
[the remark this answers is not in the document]
PhilippTakacs: Not tested, but I think working with `Vec<u64>` and `as_mut_ptr` would be a better approach. This would also avoid the `malloc`.

bindings/rust/src/lib.rs
```diff
@@ -1064,4 +1065,154 @@ impl<'a, D> Unicorn<'a, D> {
...
+            let slice_owned = slice.clone();
+            unsafe { libc::free(buffer) };
+            Ok(slice_owned)
+        } else {
+            unsafe { libc::free(buffer) };
+            Err(err)
+        }
+    }
+
+    pub fn ctl_set_exits(&self, exits: Vec<*mut c_void>) -> Result<(), uc_error> {
```
PhilippTakacs, line 1164: The type here should be `&[u64]` not `Vec<*mut c_void>`

bindings/rust/src/lib.rs
```diff
@@ -1064,4 +1065,154 @@ impl<'a, D> Unicorn<'a, D> {
...
+        let err = unsafe { ffi::uc_ctl(self.get_handle(), UC_CTL_UC_EXITS | UC_CTL_IO_WRITE, exits.as_ptr(), exits.len() as libc::size_t) };
+        if err == uc_error::OK {
+            Ok(())
+        } else {
+            Err(err)
+        }
+    }
+
+    pub fn ctl_get_cpu_model(&self) -> Result<i32, uc_error> {
```
PhilippTakacs, line 1173: Would be nice to had some explicit type for the CPU. I know this is not easy because the values of different architectures overlap.

bindings/rust/src/unicorn_const.rs
```diff
@@ -187,3 +187,25 @@ bitflags! {
...
+pub const UC_CTL_UC_TIMEOUT: u32 = 3;
+pub const UC_CTL_UC_USE_EXITS: u32 = 4;
+pub const UC_CTL_UC_EXITS_CNT: u32 = 5;
+pub const UC_CTL_UC_EXITS: u32 = 6;
+pub const UC_CTL_CPU_MODEL: u32 = 7;
+pub const UC_CTL_TB_REQUEST_CACHE: u32 = 8;
+pub const UC_CTL_TB_REMOVE_CACHE: u32 = 9;
+pub const UC_CTL_TB_FLUSH: u32 = 10;
+pub const UC_CTL_IO_READ: u32 = 1<<31;
+pub const UC_CTL_IO_WRITE: u32 = 1<<30;
```
PhilippTakacs, line 211: Would be better to have these constants in an enum.

PhilippTakacs (commented):
only the `ctl_get_exits` and `ctl_set_exits` realy need to changed. Other comments are only nice to have things. The rest looks good to me.

bindings/rust/src/lib.rs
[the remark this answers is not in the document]
ks0777: Yeah that code was a bit messy, I modfiied it to use a Vector instead which is a lot cleaner.

ks0777:
Thanks for your feedback. As suggested I improved the `ctl_get_exits`, and `ctl_set_exits` functions and cleaned up the constants a bit. As @PhilippTakacs noted the CPU an explicit type for the CPU model would be nice, but I would suggest to resolve this at a later point in time.

PhilippTakacs:
Oh, I just noticed you have missed the latest changes. The implementation of the uc_ctl_flush_tlb define has changed a bit. Also you have missed the uc_ctl_tlb_mode(). Can you add this?

PhilippTakacs:
I have a branch with the changes and also the new tlb hook: https://github.com/PhilippTakacs/unicorn/tree/rust_tlb_callback can you include this?

ks0777:
Sure, I included the changes from your branch in the PR. Do you want me to squash the commits or is this ready for a merge?

PhilippTakacs:
@wtdcode friendly reminder for this PR. I would also like to have this merged.

wtdcode:
Looks pretty good. Thanks!

PR #1820, opened 2023-04-05
add memory snapshots

PhilippTakacs:
Hi

To implement a fuzzer based on unicorn we would like to have the ability to make snapshots of the memory. This is the current version of my changes. There are some corner case I have to check and think about how best handle them.

The basic idea is to use the priority of `MemoryRegion` and overlap multiple `MemoryRegion` based on the current snapshot level. On a write (after a snapshot) a new `MemoryRegion` will be created and mapped with the priority of the current snapshot level. To restore a snapshot the `MemoryRegions` with smaller priority will be unmapped.

This also includes a updated version for `split_region` which works with aliases. This allows to unmap parts of a mapping or changing priority without copy.

The implementation only adds a few 0 checks when snapshots are not used.


aquynh:
In unicorn.h we dont use uc_struct, but uc_engine.

wtdcode:
Pretty nice work. For API design, I have two ideas:

1. mimic `uc_context` API

```
uc_err uc_alloc_snapshot(uc_engine* uc, uc_snapshot** snapshot);
uc_err uc_take_snapshot(uc_engine* uc, uc_snapshot* snapshot);
uc_err uc_restore_snapshot(uc_engine* uc, uc_snapshot* snapshot);
uc_err uc_free_snapshot(uc_engine* uc, uc_snapshot* snapshot);
```

2. merge into `uc_context` API.

If we don't change current signature of the `uc_context`, internally we could first add a new `uc_ctl` to indicate the contents a context will save, say `cpu` or `memory` or `cpu+memory`. That is to say, `uc_context_save` will save different contexts regarding this `uc_ctl`. Note we will record what we have saved (cpu, memory or both) in `uc_context` struct so that when we restore or free the context, we know what we should overwrite.

PhilippTakacs:
My approach was to use Copy on Write to implement snapshots. So making a snapshot is literally only increasing a counter. On write operations then is checked if the priority of the `MemoryRegion` is smaller then the current `snapshot_level`. When this is the case a new `MemoryRegion` is created and mapped at the address with a higher priority. To restore a snapshot the MemoryRegions are filtered by the priority and the `snapshot_level` is decreased. So the `snapshot_level` can only increased to take a snapshot and decreased to restore the last snapshot.

> For API design, I have two ideas:
> 
>     1. mimic `uc_context` API

This don't work with this implementation. The problem is making complete dumps of the complete Memory would be to big to be useful.

>     2. merge into `uc_context` API.

Some sort of combination with a context save is planed, but it's not that easy because the `uc_context` and snapshots work different. You can have multiple `uc_context` but snapshots can only increased and decreased.

wtdcode:
> My approach was to use Copy on Write to implement snapshots. So making a snapshot is literally only increasing a counter. On write operations then is checked if the priority of the `MemoryRegion` is smaller then the current `snapshot_level`. When this is the case a new `MemoryRegion` is created and mapped at the address with a higher priority. To restore a snapshot the MemoryRegions are filtered by the priority and the `snapshot_level` is decreased. So the `snapshot_level` can only increased to take a snapshot and decreased to restore the last snapshot.
> 
> > For API design, I have two ideas:
> > ```
> > 1. mimic `uc_context` API
> > ```
> 
> This don't work with this implementation. The problem is making dumps of the complete Memory would be to big to be useful.
> 
> > ```
> > 2. merge into `uc_context` API.
> > ```
> 
> Some sort of combination with a context save is planed, but it's not that easy because the `uc_context` and snapshots work different. You can have multiple `uc_context` but snapshots can only increased and decreased.

I got it. Then 

> My approach was to use Copy on Write to implement snapshots. So making a snapshot is literally only increasing a counter. On write operations then is checked if the priority of the `MemoryRegion` is smaller then the current `snapshot_level`. When this is the case a new `MemoryRegion` is created and mapped at the address with a higher priority. To restore a snapshot the MemoryRegions are filtered by the priority and the `snapshot_level` is decreased. So the `snapshot_level` can only increased to take a snapshot and decreased to restore the last snapshot.
> 
> > For API design, I have two ideas:
> > ```
> > 1. mimic `uc_context` API
> > ```
> 
> This don't work with this implementation. The problem is making dumps of the complete Memory would be to big to be useful.

I can't get it. Why not just track which memory region we should restore? I mean, say:

```
uc_snapshot_alloc(uc, &s1); // Also save this snapshot pointer in UC instance
uc_take_snapshot(uc, s1);  // Point s1 to current memory regions.

uc_mem_write(uc...); // CoW overlays a new memory region

uc_restore_snapshot(uc, s1); // Note s1 points to all previous memory regions so we can remove all regions overlayed before.
```

Note this infers that "snapshots" must be organized as a tree structure (which is also a common design in hypervisors like VMWare). That is to say, if we take the second snapshot s2 after s1, when we restore s1, s2 should be invalidated.

> 
> > ```
> > 2. merge into `uc_context` API.
> > ```
> 
> Some sort of combination with a context save is planed, but it's not that easy because the `uc_context` and snapshots work different. You can have multiple `uc_context` but snapshots can only increased and decreased.

PhilippTakacs:
To support this the internal data structure of the memory need to be also a something with supports CoW (i.e. a Merkle tree). What could be done is to also add some serialization and combine this with the snapshot_level to only copy changed memory regions. But this is out of scope for this PR.

The memory is internal stored in lists (QTAILQ), therefor I can't do CoW on the `AddressSpace`/`MemoryRegion`. But I can use the priority feature to have a `MemoryRegion` override another without changing the Data. But this way the "snapshot" is only stored internal and can only increased and decreased. What I can add is a way to increase decrease the level more then one step. This leads to some corner cases I don't know how to handle, therefor I haven't implemented this.

> Why not just track which memory region we should restore

In some way this is already done, but only internal. This stores the `snapshot_level` as `priority` of the `MemoryRegion`. So the rollback is only remove the regions with a higher or equal priority then the current level and then decrease the level. This is given more or less for free, because the priority is also used to decide which region is active.

What can be added is that `context_save` also makes a snapshot and on `context_restore` this snapshot is restored. But this only works with the same `uc_engine` object.

A bit more context what we want to do with this feature: We would like to emulate a program till it's reads an input then make a snapshot (including context save and save the internal state of the OS emulation). After that emulate till the next input. then check if this has lead to a new block, if not roll back and test with a different input.

wtdcode:
> To support this the internal data structure of the memory need to be also a something with supports CoW (i.e. a Merkle tree). What could be done is to also add some serialization and combine this with the snapshot_level to only copy changed memory regions. But this is out of scope for this PR.
> 
> The memory is internal stored in lists (QTAILQ), therefor I can't do CoW on the `AddressSpace`/`MemoryRegion`. But I can use the priority feature to have a `MemoryRegion` override another without changing the Data. But this way the "snapshot" is only stored internal and can only increased and decreased. What I can add is a way to increase decrease the level more then one step. This leads to some corner cases I don't know how to handle, therefor I haven't implemented this.

I got your point and I forget memory regions are organized in lists, which is not ideal for CoW indeed. 

> 
> > Why not just track which memory region we should restore
> 
> In some way this is already done, but only internal. This stores the `snapshot_level` as `priority` of the `MemoryRegion`. So the rollback is only remove the regions with a higher or equal priority then the current level and then decrease the level. This is given more or less for free, because the priority is also used to decide which region is active.
> 
> What can be added is that `context_save` also makes a snapshot and on `context_restore` this snapshot is restored. But this only works with the same `uc_engine` object.

You could assume a context is only restored with the same `uc_engine` object. Or we could firstly record all snapshots (contexts) we have saved and check against it when a context is going to be restored. Note this was once implemented as a workaround to support nested `uc_emu_start` but soon got removed.

> 
> A bit more context what we want to do with this feature: We would like to emulate a program till it's reads an input then make a snapshot (including context save and save the internal state of the OS emulation). After that emulate till the next input. then check if this has lead to a new block, if not roll back and test with a different input.

That sounds like what fuzzware once did and we would like to backport and I perfectly understand your motivation. Could you try to merge this to `uc_context_save` and provide a switch as `uc_ctl`?

I will start to review the implementation details probably next week, sorry for possible delay. XD

PhilippTakacs:
> Could you try to merge this to `uc_context_save` and provide a switch as `uc_ctl`?

I have add this. If enabled uc_context_save makes a snapshot. uc_context_restore will restore to the point directly after the snapshot is taken.

> That sounds like what fuzzware once did and we would like to backport

This sounds interesting. Is there some sort of overview about what and how they implement this?

wtdcode:
> > Could you try to merge this to `uc_context_save` and provide a switch as `uc_ctl`?
> 
> I have add this. If enabled uc_context_save makes a snapshot. uc_context_restore will restore to the point directly after the snapshot is taken.

Really nice, thanks!

> 
> > That sounds like what fuzzware once did and we would like to backport
> 
> This sounds interesting. Is there some sort of overview about what and how they implement this?

Check this: https://github.com/fuzzware-fuzzer/fuzzware-emulator/blob/20a3deb9227606f2efd88baa345bc3ad695d34a2/harness/fuzzware_harness/native/uc_snapshot.c

They implemented CoW by manipulating memory protections as explained in the file.

PhilippTakacs:
Found some issues with the alias handling, which needs to be resolved before merging this.

Problem is to find the region inside the address_space_memory. A simple solution would be to disallow `uc_mem_unmap`/`uc_mem_prot` in combination with snapshots.

PhilippTakacs:
Sorry for the force push, but I had to do some rewrite.

The snapshots still works the same way, but now `uc_mem_unmap` and `uc_mem_protect` can't be used in combination with snapshots. I don't think this is a big issue, because with the vtlb and the tlb_hook change protection and unmapping is still possible.

The snapshots should work as expected. I'll write some more tests and documentation.

PhilippTakacs:
> I will start to review the implementation details probably next week, sorry for possible delay. XD

Hi just wanted to know if you had time to look at the code or when you have time for it?

wtdcode:
> > I will start to review the implementation details probably next week, sorry for possible delay. XD
> 
> Hi just wanted to know if you had time to look at the code or when you have time for it?

I'm sick recently and spend most of my time in bed. Sorry I can't estimate when I could review this but I will soon have another docker appointment recently.

qemu/exec.c
```diff
@@ -955,6 +955,21 @@ void flatview_add_to_dispatch(struct uc_struct *uc, FlatView *fv, MemoryRegionSe
...
+{
+    RAMBlock *block;
+    ram_addr_t result = 0;
+
+    RAMBLOCK_FOREACH(block) {
+        result = MAX(block->offset + block->max_length, result);
+    }
+
+    if (result + size > RAM_ADDR_MAX) {
+        abort();
```
wtdcode, line 968: This is dangerous and should be avoided

qemu/exec.c
```diff
@@ -1145,6 +1164,7 @@ void qemu_ram_free(struct uc_struct *uc, RAMBlock *block)
 
     QLIST_REMOVE_RCU(block, next);
     uc->ram_list.mru_block = NULL;
+    uc->ram_list.freed = true;
```
wtdcode, line 1167: I didn't understand why a new field `freed` was added.

uc.c
```diff
@@ -2394,6 +2419,31 @@ gint cmp_vaddr(gconstpointer a, gconstpointer b, gpointer user_data)
     return va < vb ? -1 : 1;
 }
 
+UNICORN_EXPORT
+uc_err uc_snapshot(struct uc_struct *uc)
```
wtdcode, line 2423: If I understand correctly, the reason why you still need an extra API though `uc_context_*` already include the functionality is that you want to save memory but not CPU context, right? If so, I suppose we use CTL to specify what we need to save, e.g. a new CTL named like `UC_CTL_CONTEXT_CONTENT` and the possible values are like `UC_CTL_CONTEXT_CPU | UC_CTL_CONTEXT_MEMORY` etc. `snapshot` looks almost identical to `context` and it's not good to expose a separate API here AFAIK.

qemu/softmmu/memory.c
```diff
@@ -148,6 +193,21 @@ MemoryRegion *memory_map_io(struct uc_struct *uc, ram_addr_t begin, size_t size,
...
+void memory_region_filter_subregions(MemoryRegion *mr, int32_t level)
+{
+    MemoryRegion *subregion, *subregion_next;
+    memory_region_transaction_begin();
+    QTAILQ_FOREACH_SAFE(subregion, &mr->subregions, subregions_link, subregion_next) {
+        if (subregion->priority >= level) {
+            memory_region_del_subregion(mr, subregion);
+            subregion->destructor(subregion);
+            g_free(subregion);
+            mr->uc->memory_region_update_pending = true;
```
wtdcode, line 205: This seems not necessary? Though it is harmless.

qemu/softmmu/memory.c
```diff
@@ -79,6 +82,48 @@ MemoryRegion *memory_map_ptr(struct uc_struct *uc, hwaddr begin, size_t size, ui
     return ram;
 }
 
+static void make_contained(struct uc_struct *uc, MemoryRegion *current)
+{
+    hwaddr addr = current->addr;
+    MemoryRegion *container = g_new(MemoryRegion, 1);
+    memory_region_init(uc, container, int128_get64(current->size));
+    memory_region_del_subregion(uc->system_memory, current);
+    memory_region_add_subregion_overlap(container, 0, current, current->priority);
```
wtdcode, line 91: Just a reminder, this could slow down normal read/write operations though should be not significant in your case (fuzzing).

qemu/accel/tcg/cputlb.c
```diff
@@ -2147,6 +2147,21 @@ store_helper(CPUArchState *env, target_ulong addr, uint64_t val,
         }
     }
 
+    if (uc->snapshot_level && mr->ram && mr->priority < uc->snapshot_level) {
+        mr = memory_cow(uc, mr, addr & TARGET_PAGE_MASK, TARGET_PAGE_SIZE);
+        if (!mr) {
+            uc->invalid_error = UC_ERR_NOMEM;
+            cpu_exit(uc->cpu);
```
wtdcode, line 2154: Should be `cpu_loop_exit`?

qemu/accel/tcg/cputlb.c
```diff
@@ -2147,6 +2147,21 @@ store_helper(CPUArchState *env, target_ulong addr, uint64_t val,
         }
     }
 
+    if (uc->snapshot_level && mr->ram && mr->priority < uc->snapshot_level) {
+        mr = memory_cow(uc, mr, addr & TARGET_PAGE_MASK, TARGET_PAGE_SIZE);
+        if (!mr) {
+            uc->invalid_error = UC_ERR_NOMEM;
```
wtdcode, line 2153: You forget to set `uc->invalid_addr` though I can't remember exactly the location we use it.

wtdcode:
I carefully check your code and I love the idea to utilize existing qemu mechanism to support this. However, some concerns I have came up so far:

- compatibility with `uc_mem_map_ptr`, e.g. will we invalidate the pointer?
- API design, see commands above
- what if user unmaps the regions with `snapshot_level > 0`?

qemu/exec.c
PhilippTakacs: This is an optimization. With the benchmark we found that ram write time increase over the number of snapshots. So I have look at this with a profiler and found that most time is spend in `find_ram_offset()`. This function search for the best fit of a new `RAMBlock` in the `ram_addr_t` address space (special internal address space for RAMBlocks). Because of the data structure this takes O(n^2). 

When no `RAMBlock` was freed in this address space the new `RAMBlock` will be added at the end. `ram_list.freed` stores this information. Based on this boolean `find_ram_offset_last()` is called. This function skips the expensive search and just add the RAMBlock at the end of the address space.

qemu/exec.c
PhilippTakacs: This is because `find_ram_offset()` has no sane way to indicate an error. Therefore `find_ram_offset_last()` behaves like `find_ram_offset()`. I currently don't see a better way to handle this case.

qemu/softmmu/memory.c
PhilippTakacs: Not sure what you referencing but I would guess the `update_padding`:

This will force the `FlatView` to be rebuild. Without this later access to memory might go to removed regions. This could also done by rebuild the `FlatView` only after all `MemoryRegion` are filtered or by rebuild the `FlatView` forced after a `snapshot_restore`

qemu/accel/tcg/cputlb.c
PhilippTakacs: I'll add this. `uc->invalid_addr` is currently only set but not read.

uc.c
PhilippTakacs: Sounds good. I'll look at this a bit more and probably remove the `uc_snapshot` api.

PhilippTakacs:
>     * compatibility with `uc_mem_map_ptr`, e.g. will we invalidate the pointer?

The `uc_mem_map_ptr` will work as before. The change is on CoW a new RAMBlock is allocated and mapped overlapping. So without using snapshots everything behaves the same. Only after a snapshot was made writes through unicorn will not access the pointer and reads for outside might not correspond to the current state.

>     * what if user unmaps the regions with `snapshot_level > 0`?

I have look long time to find a solution for this problem. I haven't found a good one, so `uc_mem_unmap` will just return an error if `snapshot_level > 0`. To do unmapping there is still the tlb accessible. Therefore I would this would be nice to have but not important.

qemu/softmmu/memory.c
PhilippTakacs: I have not micro-benchmarked it but it's amazing fast. Also keep in mind this is done only on CoW and only one level deep. The `FlatView` does the rest.

PhilippTakacs:
Sorry for the force push wanted the CI to run and remove the fixup commits.

wtdcode:
> > ```
> > * compatibility with `uc_mem_map_ptr`, e.g. will we invalidate the pointer?
> > ```
> 
> 
> The `uc_mem_map_ptr` will work as before. The change is on CoW a new RAMBlock is allocated and mapped overlapping. So without using snapshots everything behaves the same. Only after a snapshot was made writes through unicorn will not access the pointer and reads for outside might not correspond to the current state.

That’s exactly my concern because the pointer will point to some outdated content. But I also don’t see any obvious solution so maybe we can only document this clearly.

> 
> 
> > ```
> > * what if user unmaps the regions with `snapshot_level > 0`?
> > ```
> 
> 
> I have look long time to find a solution for this problem. I haven't found a good one, so `uc_mem_unmap` (edit: same for `uc_mem_protect`) will just return an error if `snapshot_level > 0`. To do unmapping there is still the tlb accessible. Therefore I would this would be nice to have but not important.

What if we just drop all snapshots and unmap the region? Regarding uc_mem_protect, probably more cases have to be carefully handled.

PhilippTakacs:
> > > ```
> > > * compatibility with `uc_mem_map_ptr`, e.g. will we invalidate the pointer?
> > > ```
> > 
> > 
> > The `uc_mem_map_ptr` will work as before. The change is on CoW a new RAMBlock is allocated and mapped overlapping. So without using snapshots everything behaves the same. Only after a snapshot was made writes through unicorn will not access the pointer and reads for outside might not correspond to the current state.
> 
> That’s exactly my concern because the pointer will point to some outdated content. But I also don’t see any obvious solution so maybe we can only document this clearly.

This is kind of expected, because you can use the same pointer at two places with different writes (i.e. multiple mmap with MAP_PRIVAT). I'll write some examples and documentation for this.

> > > ```
> > > * what if user unmaps the regions with `snapshot_level > 0`?
> > > ```
> > 
> > 
> > I have look long time to find a solution for this problem. I haven't found a good one, so `uc_mem_unmap` (edit: same for `uc_mem_protect`) will just return an error if `snapshot_level > 0`. To do unmapping there is still the tlb accessible. Therefore I would this would be nice to have but not important.
> 
> What if we just drop all snapshots and unmap the region? Regarding uc_mem_protect, probably more cases have to be carefully handled.

For full unmap it would be possible to unmap the region and store it inside the uc_struct with the current level. On a restore just remap the region. The problem is partial unmap. With the current way CoW is implemented it would require to itterate over the FlatView and copy the old region in the worst case per page (expect unmaped parts). This could be done, but it's would be to expensive for us.

I have tried to implement this in a way without copy so that the unmap/protect without snapshots would be faster as well. But I have not found a sane way to do this.

wtdcode:
> > > > ```
> > > > * compatibility with `uc_mem_map_ptr`, e.g. will we invalidate the pointer?
> > > > ```
> > > 
> > > 
> > > The `uc_mem_map_ptr` will work as before. The change is on CoW a new RAMBlock is allocated and mapped overlapping. So without using snapshots everything behaves the same. Only after a snapshot was made writes through unicorn will not access the pointer and reads for outside might not correspond to the current state.
> > 
> > 
> > That’s exactly my concern because the pointer will point to some outdated content. But I also don’t see any obvious solution so maybe we can only document this clearly.
> 
> This is kind of expected, because you can use the same pointer at two places with different writes (i.e. multiple mmap with MAP_PRIVAT). I'll write some examples and documentation for this.
> 
> > > > ```
> > > > * what if user unmaps the regions with `snapshot_level > 0`?
> > > > ```
> > > 
> > > 
> > > I have look long time to find a solution for this problem. I haven't found a good one, so `uc_mem_unmap` (edit: same for `uc_mem_protect`) will just return an error if `snapshot_level > 0`. To do unmapping there is still the tlb accessible. Therefore I would this would be nice to have but not important.
> > 
> > 
> > What if we just drop all snapshots and unmap the region? Regarding uc_mem_protect, probably more cases have to be carefully handled.
> 
> For full unmap it would be possible to unmap the region and store it inside the uc_struct with the current level. On a restore just remap the region. The problem is partial unmap. With the current way CoW is implemented it would require to itterate over the FlatView and copy the old region in the worst case per page (expect unmaped parts). This could be done, but it's would be to expensive for us.

I see and make sense and my idea is:

- If it's a full unmap, we do everything nicely, unmap the region and map the region when restoring the snapshot later on.
- If it's a partial unmap, tell users an error as you previous proposed. This is fine if we document it.

PhilippTakacs:
So unmap is implemented. the hack for `uc->mapped_blocks` bugs me a bit, but this would require a bigger redesign to make this clean.

For the types (`mr->addr` and `mr->size`) I'm currently looking for a solution. Next I'll add some test, samples and some documentation.

PhilippTakacs:
So added some more tests and fixed some bugs now the unmap works also with snapshots. I'm not realy happy with it, because it needs some dirty hacks. The read of the first subregion I can't fix. The hack with the snapshot_level at unmap can be fixed by using a new struct. All in all it's many fiddling around with the memory regions. Also I don't expect that this is used very often.

To fix the merge conflicts I would like to rebase the branch (and forcepush). This way I can also remove the fixup commits and update some commit messages.

By the way: can you add me to the CREDITS.TXT?

PhilippTakacs:
Hi, just a friendly reminder for this pr. Or did I forgot some change requests?

uc.c
```diff
@@ -242,6 +244,10 @@ static uc_err uc_init(uc_engine *uc)
         uc->reg_reset(uc);
     }
 
+    uc->context_content = UC_CTL_CONTEXT_CPU;
+
+    uc->unmapped_regions = g_array_new(false, false, sizeof(MemoryRegion*));
```
wtdcode, line 249: I fail to find the location to free this memory. Do I miss something?

qemu/exec.c
wtdcode: I read `find_ram_offset` again and think a bit. Is it possible that we disable this optimization for fragmentation?

```C
        /* If it fits remember our place and remember the size
         * of gap, but keep going so that we might find a smaller
         * gap to fill so avoiding fragmentation.
         */
        if (next - candidate >= size && next - candidate < mingap) {
            offset = candidate;
            mingap = next - candidate;
        }
````

In other words:

```C
        /* If it fits remember our place and remember the size
         * of gap, but keep going so that we might find a smaller
         * gap to fill so avoiding fragmentation.
         */
        if (next - candidate >= size && next - candidate < mingap) {
            offset = candidate;
            mingap = next - candidate;
            break; // Stop searching.
        }
```

Could you benchmark if this solution is fast enough?

wtdcode:
The PR looks pretty good to me and is ready to go! Just a few minor changes left to be solved.

Thanks!

uc.c
PhilippTakacs: No, I missed the free.

qemu/exec.c
PhilippTakacs: As far as I understand is the ram_addr_t address space an optimization for the dirty memory tracking. The allocator avoids fragmentation to have less bookkeeping in the memory tracking. So changing this allocator to fragment more might influence the execution speed in general, independent from the use of snapshots.

> Could you benchmark if this solution is fast enough?

Will do, but my benchmark only benchmark the snapshots, not the overall execution of an complex binary with dynamic maps/unmaps. So I can't tell if this change influence the emulation speed.

PhilippTakacs:
> The PR looks pretty good to me and is ready to go! Just a few minor changes left to be solved.

Did I miss something or is now everything ready to go?

wtdcode:
> > The PR looks pretty good to me and is ready to go! Just a few minor changes left to be solved.
> 
> Did I miss something or is now everything ready to go?

Really sorry that I thought I should have replied to you. This PR is ready to go and just fixing the conflicting files should be fine enough.

qemu/exec.c
PhilippTakacs: I also guess having many ram blocks is quite uncommon. Also dynamic allocating many ram blocks is not a thing in qemu.

wtdcode:
Great thanks!

PhilippTakacs:
Thanks

PR #1826, opened 2023-04-17
use int128_get64 in memory_unmap

PhilippTakacs:
This fixes build errors introduced by bbbc7856accbae7697db84d10a28c5a6aec81856


wtdcode:
Thanks for pointing out this! Make this mistake again ;(

Issue #1539, opened 2022-01-16
Unable to read ARM64 v or q register using java binding


xmaxxaa:
Host : macOS 11.6.2
Jdk : tried openjdk 8 and openjdk 11
Unicorn  version : unicorn-2.0.0-rc5
Both c and python codes work fine:
```c
    // c
    uc_engine *uc;
    int64_t r_result;
    int64_t r = 0x123;
    uc_open(UC_ARCH_ARM64, UC_MODE_ARM, &uc);
    uc_reg_write(uc, UC_ARM64_REG_V0, &r);
    uc_reg_read(uc, UC_ARM64_REG_V0, &r_result);
    printf("r_result: %llx\n", r_result);
```
```python
    # python
    uc = unicorn.Uc(unicorn.UC_ARCH_ARM64, unicorn.UC_MODE_ARM)
    uc.reg_write(unicorn.arm64_const.UC_ARM64_REG_V0, 1)
    result = uc.reg_read(unicorn.arm64_const.UC_ARM64_REG_V0, 1)
    print(result)
```
And java code just crashed:
```java
    public static void main(String[] args) {
        Unicorn uc = new Unicorn(Unicorn.UC_ARCH_ARM64, Unicorn.UC_MODE_ARM);
        uc.reg_write(Unicorn.UC_ARM64_REG_X0, 0x1); // works fine
        uc.reg_write(Unicorn.UC_ARM64_REG_V0, 0x2); // works fine
        uc.reg_read(Unicorn.UC_ARM64_REG_X0); // works fine
        uc.reg_read(Unicorn.UC_ARM64_REG_V0); // exception
    }
```
crash log:
```
V  [libjvm.dylib+0x2f86e9]  alloc_object(_jclass*, Thread*)+0x25
V  [libjvm.dylib+0x2f900b]  jni_NewObject+0xed
C  [libunicorn_java.dylib+0x15bf]  Java_unicorn_Unicorn_reg_1read_1num+0xdf
j  unicorn.Unicorn.reg_read_num(I)Ljava/lang/Number;+0
j  unicorn.Unicorn.reg_read(I)Ljava/lang/Object;+30
j  com.xxx.Main.main([Ljava/lang/String;)V+25
v  ~StubRoutines::call_stub
V  [libjvm.dylib+0x2c3812]  JavaCalls::call_helper(JavaValue*, methodHandle*, JavaCallArguments*, Thread*)+0x688
[...]

Issue #1164, opened 2019-10-24
Java binding use CodeHook on Windows, will invoke callback before every instruction.


wizdzz:
Unicorn version 1.0.2-rc1
OS : Windows7 x64
JDK version: 1.8.0_121

I am trying use Unicorn on Windows Java, but found something that not turn out with my expect.
I compiled the needed dll(libunicorn_java.dll) for my Java code, and use the official test code of "Sample_arm.java";
and the "CodeHook" makes me confusing.

The code hooked ADDRESS(which is 0x10000) on line 68, and then call "emu_start" to start it;
so MyCodeHook.hook will be called **when the pc is 0x10000 only**,
but the fact is **it has been called before any instructions**:

![1](https://user-images.githubusercontent.com/38684262/67462910-0362f680-f673-11e9-881f-420453d95ab3.png)

so i checked "unicorn-master\bindings\java\unicorn_Unicorn.c";
 found the Jni function invoke "uc_hook_add" to add the hook handler:
```
err = uc_hook_add((uc_engine*)eng, &hh, (uc_hook_type)type, cb_hookcode, env, 1, 0, arg1, arg2);  // unicorn_Unicorn.c, line 587
```
noticed the last four params are:
"1, 0, arg1, arg2", on the caller, I really dont' know what the meaning of "1, 0",  but the last two params are begin address and end address;
However, according to the definition of uc_hook_add:
```
uc_err uc_hook_add(uc_engine *uc, uc_hook *hh, int type, void *callback,
        void *user_data, uint64_t begin, uint64_t end, ...);  // unicorn.h, line 561
```
"1, 0, arg1, arg2" should be explain as "begin is 1, end is 0, arg1 and arg2 will seem as variable arguments";
So, this is the problem, the params of "1, 0" hit the condition of **"For the special case in which begin > end, the callback will be invoked for ALL instructions.
"**, 
then i **simply modified the four params to "arg1, arg2"**, and compile it, with the new dll, i get the "right answer".

I found the last modification of unicorn_Unicorn.c was three years ago, but "uc.c" was two months ago, so maybe the "unicorn_Unicorn.c" need update ?

wtdcode:
Just quote from #1434.

> The issue you described seems the expected behavior, too.
> 
> BTW, thanks for your interest. Sorry that Unicorn lack of contributors currently and I'm not familiar with JNI unfortunately.

PR #1833, opened 2023-05-09
Rework the Java bindings

nneonneo:
This is a complete rewrite of the Java bindings for Unicorn, focused on speed, correctness and feature parity.

The existing bindings have some significant shortcomings. An incomplete list:

- Hooks don't actually work properly. All hooks are bound to all addresses, causing all hooks to fire at all addresses regardless of the registered bounds. Hook implementation is very slow due to lots of redundant work.
- `reg_read` and `reg_write` are very slow for the common use case (i.e. reading long-sized registers or smaller), as they have to box and unbox `Long`s.
- Bugs! `reg_read` returns stack garbage in the high bits because the read value isn't zeroed. The entire JVM crashes if you try and read a vector register.
- Context allocation is a raw port of the C API, demanding manual memory management and providing no access to context registers.
- It hasn't been updated since pre-2.0, and lacks many new APIs.

This PR contributes a ground-up rewrite of the bindings. It brings the Java API up to par with Python feature-wise and substantially simplifies the hook implementation, enabling proper bounds-checked hooks.

The new implementation is much faster than the old one. As a point of comparison, the following code hook, executed 10,000,000 times on an x86 `loop` instruction, takes 9.9s with the old bindings, but just 1.6s in the new bindings - a 6x performance improvement. A similar hook takes 38s to run with the Python bindings.

```java
    public static class MyCodeHook implements CodeHook {
        public long accum;
        public void hook(Unicorn uc, long address, int size, Object user_data) {
            // need to cast to int here b/c old bindings don't properly clear upper bits of value
            accum += (int)(long)uc.reg_read(Unicorn.UC_X86_REG_ECX);
        }
    }
```

The rewrite strives for compatibility with the previous API, but there are some breaking changes. It is possible to push closer to full backwards compatibility if required, at the cost of reintroducing some of the suboptimal designs. Here are the main points of breakage:

- ReadHook and WriteHook are gone, replaced simply by MemHook. Hooking valid memory accesses now requires a type parameter. This enables fetch and read-after hooks with a unified API and a single callback object.
- mem_read now takes an int, not a long. We are unable to allocate more than 2GB in a single request anyway (Java limitation).
- Instruction hooks now require specifying the instruction explicitly, instead of guessing based on the hook type. This is necessary to distinguish sysenter/syscall and ARM64 mrs/msr/sys/sysl, without excessively bloating the library with redundant hook types. Bounds must also be specified, to support bounds-checked instruction hooks.
- Reading object-type registers (any register larger than 64 bits, or registers with special formats) requires a second argument to reg_read. This allows us to provide a fast reg_read that returns a long for the common cases, while still supporting a more general reg_read for other registers.
- mem_map_ptr is rewritten to take a *direct* java.nio.Buffer, which enables many more use cases than a simple byte array, and improves performance (a byte array cannot really be used as a mapped buffer without GC-pinning it, which hurts the GC performance).
- Context handling API is redesigned to be safer and more object-oriented.

A lot of bugs are fixed with this implementation:
- Unicorn instances can be properly garbage-collected, instead of hanging around forever in the Unicorn.unicorns table.
- Hooks no longer fire outside of their bounds (#1164), and in fact, hook bounds are properly respected (previously, all hooks were just registered globally to all addresses).
- Hooks are substantially faster, as they are now dispatched directly via a single method call rather than being indirected through invokeCallbacks.
- Loading vector registers works now, rather than crashing the VM (#1539).

Several features are now enabled in the Java implementation:

- All of the current ctl_* calls are implemented.
- mmio_map is implemented.
- New virtual TLB mode is implemented.
- reading/writing Context registers is implemented.
- New hook types are added: TcgOpcodeHook, EdgeGeneratedHook, InvalidInstructionHook, TlbFillHook, and the instruction hooks Arm64SysHook, CpuidHook.
- All known special registers are supported.

### Detailed list of backwards incompatible changes

- `EventMemHook`: added `type` parameter
- Eliminated `ReadHook` and `WriteHook` and replaced both with `MemHook`. `type` added to both hooks, `value` added to read hook.
- `Unicorn`:
    - Removed `public long eng`
    - `query(int)` returns `long` now, instead of `int`
    - `reg_read(int)` returns `long` now, instead of `Object` (`Object` return replaced by `reg_read(int, Object)`)
    - `hook_add(ReadHook, long, long, Object)` removed
    - `hook_add(WriteHook, long, long, Object)` removed
    - `hook_add(MemHook, long, long, Object)` takes a new `int type` param
    - All `hook_add` methods now return a `long` instead of `void`
    - `mem_map_ptr` changed to take `long, Buffer, int` instead of `long, long, int, byte[]`


nneonneo:
By the way, if backwards-compatibility will be an issue, it should be possible to restore most of the old APIs. However, the resulting design may be less clean.

nneonneo:
OK, at this point I'm fairly happy with the code. It's decently well-tested, the samples match the C code (and the output matches too), and the performance is good. It's ready to be reviewed.

wtdcode:
> OK, at this point I'm fairly happy with the code. It's decently well-tested, the samples match the C code (and the output matches too), and the performance is good. It's ready to be reviewed.

Thanks for your work and I will review it these days as finally I feel a bit better! By the way, could you illustrate your use case?

nneonneo:
I am planning to integrate Unicorn into some Ghidra-based analysis tools. As Ghidra is written entirely in Java, it is much more efficient to use Java bindings than trying to go through Jython-ctypes.

wtdcode:
> I am planning to integrate Unicorn into some Ghidra-based analysis tools. As Ghidra is written entirely in Java, it is much more efficient to use Java bindings than trying to go through Jython-ctypes.

That's cool! At this moment I don't have too much to say about this PR because I haven't written java for a few months. I need to give it a try before giving more review but thanks again for your brilliant work!

nneonneo:
Any news on this?

wtdcode:
> Any news on this?

Rushing on a conference deadline ;(

Will have a look once getting it done.

bindings/java/Makefile
```diff
@@ -1,29 +1,89 @@
...
+jar: jarfiles
+	jar cf $(JARFILE) unicorn/*.class
+
+test: lib samples tests
+	java -Xcheck:jni -cp .:testdep/hamcrest-2.2.jar:testdep/junit-4.13.2.jar org.junit.runner.JUnitCore $(subst /,.,$(TESTS:.java=))
 
 install: lib jar
-	$(MAKE) -f Makefile.build install
+	cp libunicorn_java$(LIB_EXT) /usr/lib
+	cp $(JARFILE) /usr/share/java
```
wtdcode, line 73: Is this the canonical location? Maybe `/usr/local` is more appropriate?

bindings/java/Makefile
```diff
@@ -1,29 +1,89 @@
...
 
 uninstall:
-	$(MAKE) -f Makefile.build uninstall
+	rm -f /usr/lib/libunicorn_java$(LIB_EXT)
+	rm -f /usr/share/java/$(JARFILE)
 
 gen_const:
-	cd .. && python3 const_generator.py java
+	cd .. && python const_generator.py java
```
wtdcode, line 80: Why remove `python3` :)?

bindings/java/testdep/hamcrest-2.2.jar
wtdcode, line 1: I haven't written java for quite a long time but I still remember there exists some decent package managers (like `maven`?) to do this. I'm reluctant to add binaries files to our repo.

bindings/java/unicorn_Unicorn.c
```diff
@@ -28,768 +28,1605 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
...
 
-   jclass clz = (*env)->FindClass(env, "unicorn/X86_MMR");
-   if ((*env)->ExceptionCheck(env)) {
-      return NULL;
-   }
+static uc_err generic_reg_read(jlong ptr, jint isContext, jint regid,
+                               void *result)
+{
+    if (isContext) {
+        return uc_context_reg_read((uc_context *)ptr, regid, result);
```
wtdcode, line 476: Maybe it's time to use your new API now ;)

bindings/java/unicorn_Unicorn.c
```diff
@@ -28,768 +28,1605 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
...
+    return (jlong)hh;
+}
+
+/*
+ * Class:     unicorn_Unicorn
+ * Method:    _hook_add
+ * Signature: (JILunicorn/Hook;Ljava/lang/Object;JJI)J
+ */
+JNIEXPORT jlong JNICALL
+Java_unicorn_Unicorn__1hook_1add__JILunicorn_Hook_2Ljava_lang_Object_2JJI(
```
wtdcode, line 956: Why this signatures looks so weird?

wtdcode:
All break changes look good to me as Java binding is too old and buggy.

I have no big questions on this PR and thanks for this brilliant work!

samples/sample_arm64.c
wtdcode, line 1: Why do to make these changes to `samples/`? btw, the pac code looks taken from mine, could you add a link to the issue?

bindings/java/Makefile
nneonneo: I don't actually know! This is from the original Makefile. Honestly, the *right* solution would probably be to use Maven or Gradle to build a proper jar with the lib included, then put that jar on any classpath that needs Unicorn. If we want better packaging, I can look into doing that. (It might also be super cool to get this on Maven Central if there's interest, since that makes it easily available to any Java developer)

bindings/java/Makefile
nneonneo: Interesting, I hadn't even noticed. The old Makefile.build had `python`: https://github.com/unicorn-engine/unicorn/pull/1833/commits/93a5f9f3ef040b2f6ebedbfbf96c05e6e91c8ee7#diff-94ec5266cf434c0cb2f03d389e128624f0891519a56f84784479f13662f0e694L75.

Since `python3` seems to be universally used by the others, I'll fix this back.

bindings/java/testdep/hamcrest-2.2.jar
nneonneo: Yeah, there are. It's not really compatible with this Makefile setup, so what might be a better strategy would be to switch to Maven/Gradle (any preference?) to build artifacts and do testing, and then let them handle the dependency management.

bindings/java/Makefile
wtdcode: I’m more than happy to publish this to maven central! But this involves a few more efforts which may be out of scope of this PR:

1. Setup the namespace on maven central which we could do this and get whatever token or credentials.
2. Setup CI to build this bindings regularly. Since uc2 release, we did rust and csharp for this which you may refer.
3. For each release, send the built release jar __together with__ prebuilt libraries to maven central. This ensures that anyone can use the bindings without installing unicorn globally.

bindings/java/testdep/hamcrest-2.2.jar
wtdcode: I never used gradle honestly and I just know about maven so I will prefer maven.

bindings/java/unicorn_Unicorn.c
nneonneo: `_hook_add` is overloaded: there are zero, one and two-argument variants (for hooks like `UC_HOOK_INSN` and `UC_HOOK_TCG_OPCODE` which take extra arguments). The JNI method name needs to be mangled in this case. The names and the comment are all auto-generated by  `javac -h`.

wtdcode: Could you also document a bit (for examples, commands to generate this file etc) on how these files are generated? Or do I miss something?

The ultimate goal is that if some code is auto-generated, it should be reproducible.

samples/sample_arm64.c
nneonneo: One of the unit tests literally runs every sample compares the output with what is expected. I also wanted to add a PAC test to properly test ARM64 CP registers (since that's a particular API corner case). I ported all of the C tests to Java directly, and I have a little script that runs all the C tests, spits out their outputs, and generates the appropriate Java tests for comparison. It's extremely hacky because it handles several different `---`/`===` formats and still requires some manual post-processing, so I didn't include it in the patch. To support the script, I made the output from the samples a tiny bit more uniform.

I'll add a link to the issue for both the C and Java code.

bindings/java/unicorn_Unicorn.c
nneonneo: It's already part of the Makefile:

```
unicorn_Unicorn.h: unicorn/Unicorn.java
	$(JC) -h . $<
```

I can add a comment to the top of `unicorn_Unicorn.c` reminding people to keep the definitions in sync with the .h file.

bindings/java/Makefile
nneonneo: OK. At minimum, I'll look into adding a Maven POM to build the plugin automatically, and maybe look into hooking it up to the CI.

wtdcode: I could help on tedious Github CI setup because I’m pretty familiar with it… You can just integrate it to maven and that’s a very good start already.

wtdcode:
I will be soon on traveling so please expect another a few days absent. I will get back asap. Thanks for your patience. ;)

nneonneo:
No worries. I'll aim to get something working with Maven this week, and also look at integrating the (newly merged) reg2 API :)

nneonneo:
OK, the bindings are migrated to use Maven. The last thing on my TODO list is to switch to the reg2 API, which should be pretty straightforward.

bindings/java/Makefile
nneonneo: Maven implementation is done in dfdc8e7e8ebc07e54eef08bba784601a129c8474. Didn't touch the CI stuff; I'll leave that to you :)

nneonneo:
OK, that last commit ([763d041](https://github.com/unicorn-engine/unicorn/pull/1833/commits/763d04127c174c6479bf88ae7b0255f8ef052a82)) adds the reg2 API, which I think is the last issue that needed to be resolved. Should be good for a final review and hopefully merge :)

wtdcode:
Some last things which probably need your help:

1. I read through this: https://stackoverflow.com/questions/2937406/how-to-bundle-a-native-library-and-a-jni-library-inside-a-jar and https://github.com/adamheinrich/native-utils. Since we plan to push the package to maven central, we would like to bundle dynamic libraries into jar and load it in the static code block of the Unicorn class?.We prefer this approach for all bindings for ease of usage.
2. `javah` seems outdated and could you replace it with `javac -h`? I'm not 100% sure on this so I would be glad if you could help.
3. I'm thinking of removing `Makefile` totally so that we don't rely on `make`, which should enable it to compile on Windows. I see [exec plugin](https://www.mojohaus.org/exec-maven-plugin/usage.html) allows us to execute any command and thus we can just __compile unicorn libraries__ in `pom.xml` like other bindings did. This actually relates to my point 1 I believe.

At this moment, I'm experimenting with 1 and 3 to build dynamic libraries within `pom.xml` and bundle it to jar and load it in unicorn class. Could you have a look at 2?

nneonneo:
@wtdcode OK, it was easy to switch to `javac -h` so I have done so. I also made a fix to `const_generator` to only generate new data if there are changes, which prevents `mvn` from having to rebuild everything every time.

nneonneo:
Tests are failing, but I'm not sure it's my fault? The new test `test_x86_0xff_lcall` fails due to invalid instruction, which makes sense since the patch marks certain instructions as being invalid. Should the test itself be fixed?

wtdcode:
Ah sorry you are correct, the test should be fixed. xd

wtdcode:
I have mostly addressed the problem of building the binaries and now I'm investigating how to package everything together.

Meanwhile, I sent a request to Maven Central here:

https://issues.sonatype.org/browse/OSSRH-93144

nneonneo:
@wtdcode its been a while! Just checking: is there any plan to get this merged? Maybe we can merge it first and then worry about the build issues? It sounded like the build system itself was basically sorted out and we just wanted to get the package on Maven (which I fully support!). 

Merry Christmas and a happy new year!

wtdcode:
> @wtdcode its been a while! Just checking: is there any plan to get this merged? Maybe we can merge it first and then worry about the build issues? It sounded like the build system itself was basically sorted out and we just wanted to get the package on Maven (which I fully support!).
> 
> Merry Christmas and a happy new year!

Sure, I think that's a good start.

Regarding the build system, the reason why I failed to get it to work is that I lost all my progress (including some other work related to other issues) when I migrated my data, sorry.

Issue #1828, opened 2023-04-21
Breaking change without changing major version?


mrexodia:

mrexodia:
Just confirmed that restoring the old default behavior fixes the bug I mentioned here: https://github.com/unicorn-engine/unicorn/pull/1746#issuecomment-1517526868

```python
        def __ctl_w(ctl, nr):
            return ctl | (nr << 26) | (UC_CTL_IO_WRITE << 30)
        try:
            self._uc.ctl(__ctl_w(12, 1), 1)
        except UcError:
            pass
```

PhilippTakacs:
I'm not quit get which code is breaking. Can you provide a complete PoC which shows the issue?

wtdcode:
The reason I support #1746 is that, unicorn once used a bunch of dirty and hacky tricks and I was tired of fixing the side effects introduced by them. Therefore, my long-term goal is to remove as many as possible hacks from qemu system emulation (softmmu) by exposing suitable API.

Talking about the changes of PR, the motivation is that, unicorns try to maintain a `1:1` mapping as explained but on the top of system emulation. That is to say, some memory regions are probably not written, not executable. What's worse, some regions would be even re-mapped. To solve this, unicorn hacks `load_helper` and `write_helper` and even `mem_direct`, which obviously doesn't cover all use cases and wastes much of my debugging time.

Regarding your question, our next release will be 2.1.0 and this advancement should be enough to reflect such compatibility changes since our main API is not changed. If you have any questions, simply provide a PoC script and we are happy to fix them before the next release, though the date to be determined.

mrexodia:
Oh I also support the idea, it’s just that you are changing the default behavior that has been part of unicorn since the first version. Wouldn’t it be better to opt-in to such a drastic change and then at 3.x change it?

I’ll try to get a minimal poc working. Just a bit difficult because the setup is rather complex to work around unicorn bugs/oddities. It might take a while.

wtdcode:
> Oh I also support the idea, it’s just that you are changing the default behavior that has been part of unicorn since the first version. Wouldn’t it be better to opt-in to such a drastic change and then at 3.x change it?

We didn't change the default behavior. From the first version, unicorn offers a mixture of system emulation and `1:1` mapping, though not clearly stated anywhere. After this PR, we just make it a clear distinction by an extra control.

> 
[...]

PR #1834, opened 2023-05-09
load_helper only call cpu_loop_exit() when emulation is running

PhilippTakacs:
The load_helper is sometimes called from register writes. When the load fails check if emulation is running before jump out of the emulated code.

see #1828


wtdcode:
Merged, thanks!

PR #1835, opened 2023-05-12
Implement reg_{read,write}2 API

### LINKED ISSUE #1831: Add a uc_reg_size API
Currently, there's no way to know how much memory to allocate for `uc_reg_read` or `uc_reg_write`. For writing bindings or writing more "generic" code, it would be very useful to know how big a particular register is before performing a read/write operation.

This solves a few issues that I can see:
- This improves binding safety. Bindings can verify that they are using (or being passed) a large-enough buffer to read any given register.
- This reduces bugs when bindings don't support new registers. Bindings are not always updated to match the latest Unicorn code, and many bindings (e.g. Java, Python) fall back to reading 64-bit values on unknown registers. #1539 is an example of a bug caused by using a 64-bit read on a 128-bit register unknown to the binding: the read silently overflows the stack buffer, causing a mysterious crash. Having something like `uc_reg_size` would allow a binding to detect that the register was too large for the requested read.
- On big-endian *hosts*, doing something like `uint64_t *value; *(uint16_t *)value = ...; return *value;` will put the value in the *most-significant* bytes of the value. Ideally, Unicorn works equally well on big-endian as little-endian hosts.
- Bindings can adopt a more generic approach to reading or writing registers. Future larger SSE registers, for example, can be handled transparently by allocating a buffer of the appropriate size. This reduces the amount of special-casing in the bindings and therefore reduces maintenance burden.

Of course, the downside is maintaining a new `reg_size` for each arch. But, since we already have to maintain `reg_read` and `reg_write`, this should hopefully be an acceptable tradeoff.

I'm happy to take this and work on it; I'm putting this issue out to see if it would be worth doing in the first place (i.e. would be accepted into Unicorn).

nneonneo:
As suggested in #1831, this patch implements a "version 2" API for register read/write which takes a size parameter. Context and batch operations are all supported.

This patch also comes with some performance optimizations and some (slightly opinionated) refactoring. The performance improvement appears to be about 20-25% for a single read/write operation or about 10% for a batch operation of size 8, despite the addition of the size parameter.

As a side benefit, the register APIs now produce an appropriate error if an invalid (unknown) register is used.

Fixes #1831.


qemu/target/riscv/unicorn.c
```diff
@@ -76,539 +77,106 @@ static void riscv_release(void *ctx)
...
-    case UC_RISCV_REG_HTVAL:
-    case UC_RISCV_REG_HIP:
-    case UC_RISCV_REG_HTINST:
-    case UC_RISCV_REG_HGATP:
-    case UC_RISCV_REG_HTIMEDELTA:
-    case UC_RISCV_REG_HTIMEDELTAH: {
+    } else if (regid >= UC_RISCV_REG_F0 &&
+               regid <= UC_RISCV_REG_F31) { // "ft0".."ft31"
+        CHECK_REG_TYPE(uint64_t);
+        *(uint64_t *)value = env->fpr[regid - UC_RISCV_REG_F0];
```
nneonneo, line 100: note bug fix here (slight incompatibility): F registers should read all 64 bits even in 32-bit mode, as they can hold double-precision floating point values (doubles)

qemu/unicorn_common.h
```diff
@@ -128,4 +128,13 @@ static inline void uc_common_init(struct uc_struct* uc)
     if (!uc->release)
         uc->release = release_common;
 }
+
+#define CHECK_REG_TYPE(type) do {             \
+    if (unlikely(*size < sizeof(type))) {     \
+        return UC_ERR_NOMEM;                  \
```
wtdcode, line 134: This is not correct, I suggest `UC_ERR_ARG`

uc.c
```diff
@@ -211,7 +223,7 @@ static gint uc_exits_cmp(gconstpointer a, gconstpointer b, gpointer user_data)
     }
 }
 
-static uc_err uc_init(uc_engine *uc)
+static uc_err uc_init_engine(uc_engine *uc)
```
wtdcode, line 226: Any reason for these renames?

uc.c
```diff
@@ -1951,85 +2058,174 @@ static void find_context_reg_rw_function(uc_arch arch, uc_mode mode,
...
+
+UNICORN_EXPORT
+uc_err uc_context_reg_write_batch2(uc_context *ctx, int *regs,
+                                   const void *const *vals, size_t *sizes,
+                                   int count)
+{
+    reg_write_t reg_write = find_context_reg_rw(ctx->arch, ctx->mode).write;
+    void *env = ctx->data;
+    int mode = ctx->mode;
+    int setpc = 0;
```
wtdcode, line 2195: Why is `setpc` argument added?

uc.c
```diff
@@ -505,49 +519,158 @@ uc_err uc_close(uc_engine *uc)
...
+    return uc->reg_read(uc->cpu->env_ptr, uc->mode, regid, value, &size);
 }
 
 UNICORN_EXPORT
 uc_err uc_reg_write(uc_engine *uc, int regid, const void *value)
 {
     UC_INIT(uc);
-    return uc_reg_write_batch(uc, &regid, (void *const *)&value, 1);
+    int setpc = 0;
+    size_t size = (size_t)-1;
```
wtdcode, line 635: Personally I perfer using `0` as the default value because `(size_t)-1` implies the maximum value of the unsigned type, which is error-prone when writing `reg_read` code. For example, some contributors may simply write `memcpy(val, ptr, size)` someday without noticing that `size` can be `(size_t)-1`. However, this requires some extra code when checking the size but I think the tradeoff worths it.

uc.c
wtdcode: I see, for reducing code duplication for `break_translation_loop`, right?

wtdcode:
Look good to me in general though ideally your fixes and code format should be splitted into smaller commits. The work is pretty nice and complete!

However, we also need to port this to new bindings like python, rust and your java bindings. But this is out of this PR's scope and I don't have further concerns.

qemu/unicorn_common.h
nneonneo: Hmm, I wanted it to be distinct from "register not found" (which is UC_ERR_ARG) because it may hint the caller to try a different buffer size. I realize NOMEM is not quite accurate. Any other suggestions?

uc.c
nneonneo: I changed `<arch>_uc_init` to `uc_init_<arch>` via symbols.sh in d68013d7745fd2de0628de721cbe250df36619c9, mainly to make things a bit more uniform across architectures. Since that #define's uc_init, I had to change the name here to reduce confusion. It should not be a big issue given that these are static.

uc.c
nneonneo: Yes, it's to reduce code duplication - specifically, it lets us merge the handling of context and non-context registers, which substantially reduces the amount of per-arch code that needs to be maintained.

At some point, we'll probably need an "isContext" flag so we don't go touching CPU state in context operations, but that's an issue for a future PR.

uc.c
nneonneo: The precise semantics of `size` are that it specifies the size of the buffer on input, and the size that was actually written on output. It is completely reasonable that callers may allocate more memory than required, and they can check how much memory was actually used via the return value in `*size`. `memcpy(val, ptr, size)` is categorically wrong, because the size is just a maximum. I would call it `maxsize`, but because it gets set to the true size upon return, that would be slightly misleading.

`(size_t)-1` is for compatibility with the old API, which assumed that the buffer was unbounded. `(size_t)0` will require special handling and is semantically more convoluted. The most straightforward definition is that `(size_t)0` means a buffer of size zero; although this seems useless on the surface, it can be useful for e.g. poking registers that have side-effects (none in current architectures, but many ARM CP registers behave this way).

nneonneo:
> However, we also need to port this to new bindings like python, rust and your java bindings. But this is out of this PR's scope and I don't have further concerns.

I think that we would not expose these bindings to Python or Java directly (maybe Rust though). Rather, we would just have Python/Java always call the new APIs for safety, since they will always know how big their input buffers are.

wtdcode:
> > However, we also need to port this to new bindings like python, rust and your java bindings. But this is out of this PR's scope and I don't have further concerns.
> 
> I think that we would not expose these bindings to Python or Java directly (maybe Rust though). Rather, we would just have Python/Java always call the new APIs for safety, since they will always know how big their input buffers are.

Yes, that's my point. We should further update bindings to use the new API.

nneonneo:
Any news on this PR?

wtdcode:
> Any news on this PR?

Rushing on a conference deadline ;(

Will have a look once getting it done.

qemu/target/riscv/unicorn.c
wtdcode: This is fine!

uc.c
wtdcode: Sound reasonable!

qemu/unicorn_common.h
wtdcode: I suggest a new error type, what's your thoughts? Maybe `UC_ERR_REG`. Note to add a new error, you also need to update `uc_strerror`.

wtdcode:
I'm done with my deadlines and will soon start to review your two PRs (and probably other stale ones XD).

So far, for this one, I don't see any further problems. Thanks in advance!

nneonneo:
Rebased to latest dev, and changed `UC_ERR_NOMEM` to a new error `UC_ERR_OVERFLOW` - hope that's reasonable?

qemu/unicorn_common.h
nneonneo: Changed to a new error `UC_ERR_OVERFLOW`.

wtdcode:
Merged, thanks for your big contributions!

PR #1836, opened 2023-05-22
fix UC_MEM_WRITE_PROT callback

PhilippTakacs:
callbacks work on the physical address.

As far as I see this is the only case I have add this mistake, but would be nice when someone else could double check.


wtdcode:
Thanks!

PR #1838, opened 2023-05-24
Ignore page_collection_lock

tunz:
Unicorn is already ignoring page locks, so we also don't need to manage the page collection and its tree objects. It helps us to avoid redundant object allocations and speed up the execution 3x faster for my case.

### DIFF
--- a/qemu/accel/tcg/translate-all.c
+++ b/qemu/accel/tcg/translate-all.c
@@ -650,6 +650,7 @@ static gint tb_page_addr_cmp(gconstpointer ap, gconstpointer bp, gpointer udata)
 struct page_collection *
 page_collection_lock(struct uc_struct *uc, tb_page_addr_t start, tb_page_addr_t end)
 {
+#if 0
     struct page_collection *set = g_malloc(sizeof(*set));
     tb_page_addr_t index;
     PageDesc *pd;
@@ -664,9 +665,7 @@ page_collection_lock(struct uc_struct *uc, tb_page_addr_t start, tb_page_addr_t 
     assert_no_pages_locked();
 
  retry:
-#if 0
     g_tree_foreach(set->tree, page_entry_lock, NULL);
-#endif
 
     for (index = start; index <= end; index++) {
         TranslationBlock *tb;
@@ -677,9 +676,7 @@ page_collection_lock(struct uc_struct *uc, tb_page_addr_t start, tb_page_addr_t 
             continue;
         }
         if (page_trylock_add(uc, set, index << TARGET_PAGE_BITS)) {
-#if 0
             g_tree_foreach(set->tree, page_entry_unlock, NULL);
-#endif
             goto retry;
         }
         assert_page_locked(pd);
@@ -688,21 +685,24 @@ page_collection_lock(struct uc_struct *uc, tb_page_addr_t start, tb_page_addr_t 
                 (tb->page_addr[1] != -1 &&
                  page_trylock_add(uc, set, tb->page_addr[1]))) {
                 /* drop all locks, and reacquire in order */
-#if 0
                 g_tree_foreach(set->tree, page_entry_unlock, NULL);
-#endif
                 goto retry;
             }
         }
     }
     return set;
+#else
+    return NULL;
+#endif
 }
 
 void page_collection_unlock(struct page_collection *set)
 {
+#if 0
     /* entries are unlocked and freed via page_entry_destroy */
     g_tree_destroy(set->tree);
     g_free(set);
+#endif
 }
 
 static void page_lock_pair(struct uc_struct *uc, PageDesc **ret_p1, tb_page_addr_t phys1,
--- a/qemu/accel/tcg/translate-all.c
+++ b/qemu/accel/tcg/translate-all.c
@@ -522,6 +522,7 @@ static inline void page_unlock_tb(struct uc_struct *uc, const TranslationBlock *
     }
 }
 
+#if 0
 static inline struct page_entry *
 page_entry_new(PageDesc *pd, tb_page_addr_t index)
 {
@@ -542,7 +543,6 @@ static void page_entry_destroy(gpointer p)
     g_free(pe);
 }
 
-#if 0
 /* returns false on success */
 static bool page_entry_trylock(struct page_entry *pe)
 {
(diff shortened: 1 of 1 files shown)


qemu/accel/tcg/translate-all.c
```diff
@@ -677,9 +676,7 @@ page_collection_lock(struct uc_struct *uc, tb_page_addr_t start, tb_page_addr_t
             continue;
         }
         if (page_trylock_add(uc, set, index << TARGET_PAGE_BITS)) {
```
wtdcode, line 678: But you also remove these function calls, which have side effects obviously.

tunz: I think `page_trylock_add` also does nothing other than allocating memories. or did I miss anything?

tunz: Another side effect is `page_entry_destroy` call when destroying objects, but it's also same. `page_lock` is no-op in Unicorn, so I think it's a meaningless call.

wtdcode: `page_collection_lock` is used by `tb_invalidate_phys_range`, which is not non-ops in my experience.

wtdcode: `page_trylock_add` also inserts new pages, no?

wtdcode:
For the speedup, I suggest you sharing the backtrace and `pprof` reports if possible.

qemu/accel/tcg/translate-all.c
tunz: hmm interesting. I'm running it without `page_collection_lock`/`unlock`, but no problem yet. `tb_invalidate_phys_range` uses the page collection only for `page_collection_unlock` call.

`page_trylock_add` does not insert new pages. It creates a page entry and insert it to the tree, but it's not used anywhere. I guess it's only for managing order of locks.

tunz:
Here is a snippet of profiling. Profiled by WPR and tested on Windows.

Dev branch
```
- helper_le_stq_mmu_x86_64 (68.38%)
  - store_helper (67.94%)
    - notdirty_write (33.63%)
      - page_collection_lock_x86_64 (30.94%)
      - tb_invalidate_phys_page_fast_x86_64 (1.57%)
      - ...
    - page_collection_unlock_x86_64 (15.25%)
    - _free_base (9.64%)
    - helper_ret_stb_mmu_x86_64 (3.81%)
    - find_memory_region (1.35%)
    - ...
  - ...
- helper_ret_stb_mmu_x86_64 (7.17%)
  - store_helper (7.17%)
  - ...
- helper_le_stl_mmu_x86_64 (4.04%)
- helper_lookup_tb_ptr_x86_64 (2.47%)
- helper_le_ldq_mmu_x86_64 (1.79%)
- ...
```

With this fix:
```
- helper_le_stq_mmu_x86_64 (32.70%)
  - store_helper (31.35%)
    - store_helper <itself> (15.13%)
    - notdirty_write (10.81%)
      - tb_invalidate_phys_page_fast_x86_64 (7.57%)
      - ...
    - ...
- helper_lookup_tb_ptr_x86_64 (7.30%)
- helper_le_ldq_mmu_x86_64 (5.14%)
- helper_uc_tracecode (3.78%)
- helper_ret_stb_mmu_x86_64 (3.24%)
  - store_helper (2.97%)
  - ...
- ...
```

Percentage means a proportion of CPU usage. Execution time is reduced from 370s to 116s.

wtdcode:
> Here is a snippet of profiling. Profiled by WPR and tested on Windows.
> 
> Dev branch
> 
> ```
> - helper_le_stq_mmu_x86_64 (68.38%)
>   - store_helper (67.94%)
>     - notdirty_write (33.63%)

That's it and it matches my guess, this is slow because unicorn (qemu) has to ensure previous writes does invalidate any previous dirty code pages. This is specially handled due to Self Modifying Code (refer to qemu whitepaper if you wish).

>       - page_collection_lock_x86_64 (30.94%)
>         - _malloc_base (15.25%)
>         - g_tree_new_full (13.45%)
>         - page_find_alloc (1.35%)
>         - ...
>       - tb_invalidate_phys_page_fast_x86_64 (1.57%)
>       - ...
>     - page_collection_unlock_x86_64 (15.25%)
>     - _free_base (9.64%)
>     - helper_ret_stb_mmu_x86_64 (3.81%)
>     - find_memory_region (1.35%)
>     - ...
>   - ...
> - helper_ret_stb_mmu_x86_64 (7.17%)
>   - store_helper (7.17%)
>   - ...
> - helper_le_stl_mmu_x86_64 (4.04%)
> - helper_lookup_tb_ptr_x86_64 (2.47%)
> - helper_le_ldq_mmu_x86_64 (1.79%)
> - ...
> ```
> 
> With this fix:
> 
> ```
> - helper_le_stq_mmu_x86_64 (32.70%)
>   - store_helper (31.35%)
>     - store_helper <itself> (15.13%)
>     - notdirty_write (10.81%)
>       - tb_invalidate_phys_page_fast_x86_64 (7.57%)
>       - ...
>     - ...
> - helper_lookup_tb_ptr_x86_64 (7.30%)
> - helper_le_ldq_mmu_x86_64 (5.14%)
> - helper_uc_tracecode (3.78%)
> - helper_ret_stb_mmu_x86_64 (3.24%)
>   - store_helper (2.97%)
>   - ...
> - ...
> ```
> 
> Percentage means a proportion of CPU usage. Execution time is reduced from 370s to 116s.

I think a better solution is to check how qemu solves this because I feel that it shouldn't have too much overhead. Maybe we miss some core mechanism when manipulating qemu code.

tunz:
Oh, thanks for the hint. Now I see this problem: Unicorn always returns `false` for `cpu_physical_memory_get_dirty` and `true` for `cpu_physical_memory_is_clean`. So, `tlb_set_dirty` is never called in `notdirty_write`, and Unicorn always takes the slow path.

Then, does it sound ok to remove `TLB_NOTDIRTY` usages?

wtdcode:
> Oh, thanks for the hint. Now I see this problem: Unicorn always returns `false` for `cpu_physical_memory_get_dirty` and `true` for `cpu_physical_memory_is_clean`. So, `tlb_set_dirty` is never called in `notdirty_write`, and Unicorn always takes the slow path.

Thanks for your hint. This is an old known issue I notice when I developed Unicorn2. The root cause is a bit long story but keep it for short: for uc1 capability. However, now we have some ctl for flushing code cache so this could be improved in some way but I need to figure it out.

wtdcode:
You are correct, merged!

PR #1839, opened 2023-05-30
Avoid notdirty_write

tunz:
Attempt to avoid `notdirty_write` as much as possible. Please refer to #1838 for the context.

If there is no TB in the page, we can skip the slow path, so mark the TLB as `dirty` in such case. And `PageDesc` is allocated only when TB is generated, so we can apply the same logic for the case where `PageDesc` is not found.

If Unicorn generates a TB, we have to take the slow path for the page. In `tb_page_add`, if it's the first tb generated, `tlb_protect_code` will be called, and it will call `cpu_physical_memory_test_and_clear_dirty`. In this function, we reset the dirty flag so that Unicorn calls `notdirty_write` when memory is overwritten in the page.

Tested with `tests/regress/x86_self_modifying.py`.


wtdcode:
Superseded by #1970 and we have ported the fix. Closing.

Issue #1777, opened 2023-02-13
Unicorn doesn't support x86 MOVDIRI instruction?


smmalis37:
Attempting to run the instruction bytes `0f 38 f9 00` returns an INSN_INVALID error. I'd expect it to just run the instruction.

wtdcode:
Please check your CPUID to see if it's supported. Note you could always change CPU model.

smmalis37:
It is supported on my CPU. How do I change CPU model in the Rust bindings?

wtdcode:
> It is supported on my CPU. How do I change CPU model in the Rust bindings?

I mean, you should check if the instruction is supported or not for the emulated CPU.

smmalis37:
How do I control that? I don't see anything related to the emulated CPU model, I just call Unicorn::new() and write things into memory and set up my hooks.

smmalis37:
Did a bit more digging, are you referring to uc_ctl_set_cpu_model? It is not currently included in the Rust bindings, nor is the list of supported cpu models. In fact, I only see it called in the Go and Java bindings, and none of the others.

wtdcode:
> Did a bit more digging, are you referring to uc_ctl_set_cpu_model? It is not currently included in the Rust bindings, nor is the list of supported cpu models. In fact, I only see uc_ctl_set_cpu_model called in the Go and Java bindings, and none of the others.

Yes correct because Rust bindings is more of maintained by the community and I lack time to fix it...

PR #1847, opened 2023-07-12
Add CpuModel's for architectures to rust bindings

### LINKED ISSUE #1821: rust: expose ffi
the unicorn rust ffi is too high level for much customization, to enable an "opt-in" choice to more flexibility the `ffi` module should get re-exported public since there is no `unicorn-sys` create or equivalent etc.

lockbox:
This adds the enums for all the architectures to the rust bindings so that we can used `uc_set_cpu_model`. this also exposes the `ffi` module so that we can set the cpu model

Closes #1821 

And addresses concerns (for rust at least) for issue #1777


wtdcode:
Sorry for late. I just got on my desktop working. 

Thanks for your contribution! It's awesome.

Issue #1845, opened 2023-07-11
does unicorn have exec-only page?


HyperSine:
Hi, considering the following code:

```py
import unicorn

uc = unicorn.Uc(unicorn.UC_ARCH_X86, unicorn.UC_MODE_64)
uc.mem_map(0x400000, 0x1000, unicorn.UC_PROT_EXEC)

# mov al, byte ptr[rip]
# nop
# nop
# ...
uc.mem_write(0x400000, b'\x8a\x05\x00\x00\x00\x00'.ljust(0x1000, b'\x90'))

uc.emu_start(0x400000, 0x400006)    # run until first `nop`
# expect a UcError, but no error is raised

print('al = 0x{:02x}'.format(uc.reg_read(unicorn.x86_const.UC_X86_REG_AL)))
# it prints: al = 0x90
```

The instruction `mov al, byte ptr[rip]` would have read access to the byte at 0x400006 which resides in a page that is mapped with only  UC_PROT_EXEC flag. I would expect there is a UC_ERR_READ_PROT error but nothing happened. It seems UC_PROT_EXEC implies UC_PROT_READ. Is it by design? Or a bug?

PhilippTakacs:
This is a bug, because if you map two pages and change the code to `\x8a\x05\x00\x10\x00\x00` (`mov al, byte ptr[rip+0x1000]`) you get the expected `UC_ERR_READ_PROT`.

PhilippTakacs:
I have looked a bit more at it and this isn't that easy to solve. The problem that the load helper is only called once and the result is cached. Because the virtual mmu doesn't know about the UC_PROT_* permissions and defaults to the x86 default permissions. So the cache is stored as read/write/exec. So I don't think there is a good solution for this bug.

As a workaround you can register a mem_read callback to all exec only pages. An alternative workaround would be to switch to UC_TLB_VIRTUAL and register a tlb callback for the exec-only pages with which sets these pages exec only[0].

[0] Register a tlb callback shouldn't be neccesary, I'll look for a fix.

PhilippTakacs:
you can try following patch, than only switch to UC_TLB_VIRTUAL fix your problem:
```
diff --git a/qemu/softmmu/unicorn_vtlb.c b/qemu/softmmu/unicorn_vtlb.c
index 3b629bda..4ebade7e 100644
--- a/qemu/softmmu/unicorn_vtlb.c
+++ b/qemu/softmmu/unicorn_vtlb.c
[...]

PR #1848, opened 2023-07-17
vtlb only cache access permisions required for the operation

PhilippTakacs:
see #1845


wtdcode:
Thanks! I'm finally back on my desktop...

PR #1849, opened 2023-07-18
Fix rust compiler warnings

mlgiraud:
This PR fixes multiple `warning: private item shadows public glob re-export` warnings caused by an unnecessary use statement


wtdcode:
LGTM

PR #1850, opened 2023-07-18
Fix clippy warnings

mlgiraud:
This PR fixes a few clippy warnings caused by an outdated bitflags crate. Also adds two allow annotations for the raw pointer handles.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ exclude = [
 path = "bindings/rust/src/lib.rs"
 
 [dependencies]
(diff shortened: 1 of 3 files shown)


wtdcode:
LGTM

wtdcode:
But compilation fails, could you check it? It seems straightforward to fix.

mlgiraud:
Done, should compile now. Forgot to run tests locally. It was just a missing derive(Copy)

wtdcode:
Merged, thanks!

PR #1861, opened 2023-08-03
Backport qemu/qemu@75b208c

basavesh:
target/i386: fix operand order for PDEP and PEXT

For PDEP and PEXT, the mask is provided in the memory (mod+r/m) operand, and therefore is loaded in s->T0 by gen_ldst_modrm. The source is provided in the second source operand (VEX.vvvv) and therefore is loaded in s->T1. Fix the order in which they are passed to the helpers.


wtdcode:
Thanks!

PR #1866, opened 2023-08-11
fix: Make functions that deref ptr unsafe

mlgiraud:
Move the unsafe for functions that dereference a pointer that is passed as unsafe to function level. For an explanation why this was bad see here: https://rust-lang.github.io/rust-clippy/master/index.html#/not_unsafe_ptr_arg_deref


wtdcode:
LGTM and wait for CI but please separate `cargo fmt` in a standalone commit next time.

wtdcode:
Thansk and welcome!

PR #1867, opened 2023-08-14
fix: Wrap uc_hook to not expose ffi types in public api

mlgiraud:
The current api exposes raw pointers to the user. This change wraps the internal pointer type into a public facing type. This avoids direct use of the ffi for normal api use. Using the ffi directly is still possible.


wtdcode:
Looks good and thanks!

PR #1868, opened 2023-08-14
Fix/from handle unsafe

mlgiraud:
Hi, another small api change. Currently the TryFrom trait is used to convert a uc_handle (a raw pointer) into a Unicorn struct. This is inherently unsafe and should be marked as such. Since the TryFrom trait cannot be unsafe, i introduced a new method `from_handle` that is marked as unsafe. This way the user is responsible for ensuring the safety of the program (we cannot guarantee this, since the uc_handle pointer is supplied by the user).

### DIFF
--- a/bindings/rust/src/lib.rs
+++ b/bindings/rust/src/lib.rs
@@ -32,6 +32,28 @@
 extern crate alloc;
 extern crate std;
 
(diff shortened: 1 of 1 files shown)


wtdcode:
Thanks!

PR #1873, opened 2023-08-23
fix: Use correct addresses during memory cow

mlgiraud:
Title is mostly self explanatory. There was an issue where performing snapshots with a non one to one mapping of virtual to physical addresses, resulting in an assertion failure in unicorn. I assume this commit fixes the issue. @PhilippTakacs should check if this is correct now.

### DIFF
--- a/bindings/rust/src/lib.rs
+++ b/bindings/rust/src/lib.rs
@@ -35,7 +35,6 @@ extern crate std;
 use alloc::boxed::Box;
 use alloc::rc::Rc;
 use alloc::vec::Vec;
-use bitflags::Flags;
 use core::cell::UnsafeCell;
 use core::ptr;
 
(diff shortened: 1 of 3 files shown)


wtdcode:
I can understand the motivation and looks correct at first glance. Do you have a reproduction? So that I can make sure that the fix is correct.

mlgiraud:
I don't have a reproduction script since this occurred in our emulator for which i unfortunately cannot yet share the code. I will see if i can write something up that reproduces this.

mlgiraud:
I added a test that reproduces the issue. You can check by reverting the commit that fixes the issue.

wtdcode:
> I added a test that reproduces the issue. You can check by reverting the commit that fixes the issue.

I see. Will check and thanks!

PhilippTakacs:
LGTM

mlgiraud:
@wtdcode Can we merge this, or are there some changes that need to be made?

wtdcode:
Sorry just busy on something else and I will have a look today or tomorrow.

wtdcode:
I tested and it's the correct fix. Thanks and I appreciate your patience.

PR #1891, opened 2023-10-18
Created a CI to automatically attach builds to releases

marameref:
In this PR, I Created a workflow on release.yml that automatically creates a release which is triggered by tag push to remote/origin.
The workflow includes the steps to

Create a release,
Download artifacts locally
Attach the downloaded artifacts to the release
To trigger this, you can tag your repo using git tag and push the repo to production using git push origin ,
can be any tag starting with rc* eg rc0.09, rc0.01, rc1.2 etc


.github/workflows/release.yml
```diff
@@ -0,0 +1,51 @@
...
+
+      - name: Create Release
+        id: create_release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ github.ref }}
+          release_name: Release ${{ github.ref }}
+          draft: false
```
wtdcode, line 25: Please create a draft anyway because we will edit it in most cases

.github/workflows/release.yml
```diff
@@ -0,0 +1,51 @@
...
+          prerelease: false
```
wtdcode, line 26: Not sure if it's possible with Github Action, but it would be better `prerelease` is set to `true` when the tag contains `rc`.

wtdcode:
This workflow looks pretty promising. Before merging this, could you:

1. show a successful run of this workflow
2. rebase to dev branch

PR #1894, opened 2023-10-20
Implemented optional pointer for providing register size to operations

BitMaskMixer:
- This shall prevent buffer overflows during usage of too small variables for these functions.
- A macro has been added to check the correct register size - and currently used as POC for ARM 32 only
- The cpp tests are changed too, to demonstrate the usage.

- Breaking change: Bindings must reflect these changes


wtdcode:
Thanks for your interests. We did this already in dev branch via https://github.com/unicorn-engine/unicorn/commit/4055a5ab109c1d8d2da06515f3e117ca57faf179. Does that address your concern?

PR #1899, opened 2023-10-24
fix fxsave fpip value & provide unit tests

dglynos:
This is the same fix (and tests) that was merged in unicorn v1 (see https://github.com/unicorn-engine/unicorn/commit/63a445cbba18bf1313ac3699b5d25462b5d529f4/) now ported to unicorn v2. The fix makes sure that the fxsave x86/x64 instruction saves the instruction pointer (instead of zero) into the floating point context dump. This instruction is used in many modern polymorphic shellcodes.


wtdcode:
Thanks for your contribution. Could you rebase it to dev branch?

dglynos:
@wtdcode closing this pull request, issuing a new one based on the dev branch, see https://github.com/unicorn-engine/unicorn/pull/1900

Issue #1909, opened 2023-12-05
PowerPC "mfspr r3, DEC" emulation segmentation fault


dotCirill:
To get the fault you can change in `samples/sample_ppc.c` line `#define PPC_CODE "\x7F\x46\x1A\x14"` to `#define PPC_CODE "\x7c\x76\x02\xa6"`. This code (`mfspr r3, 0x16`) leads to SegFault.

Tested on master, dev branches. Builded with `cmake .. -DUNICORN_ARCH=ppc`.

```
>>> Tracing basic block at 0x10000, block size = 0x4
>>> Tracing instruction at 0x10000, instruction size = 0x4
Segmentation fault (core dumped)
```

The problem is in `cpu_ppc_load_decr` function in `qemu/hw/ppc/ppc.c`. It calles `_cpu_ppc_load_decr(env, env->tb_env->decr_next)`, but `tb_env` is NULL.

This function can be fixed either by returning just some value or  by calling `cpu_ppc_tb_init (CPUPPCState *env, uint32_t freq)`.

dotCirill:
The same issue with `#define PPC_CODE "\x7c\x0d\x42\xa6"` (`mfspr r0,TBUr`).

It crashes in `_cpu_ppc_load_tbu` because `tb_env` is not initialized.

wtdcode:
PR is highly welcome because I'm deadline-surfing these days.

PR #1910, opened 2023-12-06
Fix PPC32 fault when timer-spr access

dotCirill:
Issue #1909 

Access to TB, DEC (time-related) Special Purpose Registers led to crash.

`spr_read_decr`, `spr_write_decr` and others are changed to `spr_read_generic` and `spr_write_decr`.
So spr read/write won't call Qemu time-related functions.


wtdcode:
Thanks & welcome!

PR #1919, opened 2024-01-10
fix cow when using uc_mem_write

PhilippTakacs:
memory_cow expect the address and size to be aligned on pagesize.

### DIFF
--- a/qemu/softmmu/memory.c
+++ b/qemu/softmmu/memory.c
@@ -98,6 +98,9 @@ MemoryRegion *memory_cow(struct uc_struct *uc, MemoryRegion *current, hwaddr beg
     hwaddr current_offset;
     MemoryRegion *ram = g_new(MemoryRegion, 1);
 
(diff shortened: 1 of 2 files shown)


qemu/softmmu/memory.c
```diff
@@ -98,6 +98,9 @@ MemoryRegion *memory_cow(struct uc_struct *uc, MemoryRegion *current, hwaddr beg
     hwaddr current_offset;
     MemoryRegion *ram = g_new(MemoryRegion, 1);
 
+    assert((begin & ~TARGET_PAGE_MASK) == 0);
```
wtdcode, line 101: These checks look not necessary considering you have aligned them before calling the function?

PhilippTakacs: These debug checks are to avoid adding the bug again. In release mode the they are noops. But when some change the code and miss the alignment, these checks can help to find the bug.

wtdcode:
Merged, thanks!

PR #1920, opened 2024-01-16
fix uc_mem write with cow when using upper half of the address

PhilippTakacs:
I missed a bug in my last fix.

uc->target_page_align is a uint32_t. When the binary not will only invert the 32 bit of the value. Used this in a binary and operator with a uint64_t will case the upper 32bit of the address to be 0. Therefor the bug only appears when the upper 32bit of the address are used.

Now a local uint64_t variable is used for the alignemend and the test uses a not page alligned address which does not fit in 32bit.


wtdcode:
Here you go, thanks!

PR #1922, opened 2024-01-24
Backport from qemu : fix do_nonatomic_op_* vs signed operations

redoste:
This fix makes non-atomic versions of min/max operations aware of sign extension.

The following piece of RISC-V assembly was producing bogus results:
```asm
	la a0, .d
	li a2, 1

	amomin.w a5, a2, (a0)
	lw a6, 0(a0)
.d:
	.word -1
```

The expected result in `a6` is `-1` but without the fix the result is `1`


wtdcode:
Looks okay, simply wait for CI.

PR #1928, opened 2024-02-26
Feature: Bindings: add start end to block hook

mlgiraud:
The block hook was missing start and end parameters to limit the address range of the hook.


wtdcode:
Hello, thanks for your work but looks like you need to fix tests firstly.

mlgiraud:
Sorry for the delay. Had to write some proposals... Everything should be fixed now. I simply forgot to add the parameters in the rust tests.

wtdcode:
Cool, thanks!

PR #1929, opened 2024-03-06
Remove semicolons in python files

xclusivor:
Noticed an errant semicolon while looking at the x86 sample and was curious if there were more so just cleaned a few more up real quick.


wtdcode:
If i remember correctly, they are from previous py2 scripts.

wtdcode:
But you should target dev branch.

wtdcode:
Merged, thanks!

PR #1932, opened 2024-03-27
Rust conditional compilation

lockbox:
This is split into 2 commits
- apply default rust 2021 formatting to entire rust codebase
- conditionally include rust code based on the compilation bindings / features set when used in cargo.toml

### DIFF
--- a/bindings/rust/src/lib.rs
+++ b/bindings/rust/src/lib.rs
@@ -42,32 +42,72 @@ use libc::c_void;
 
 use ffi::uc_handle;
 
-pub use crate::arm::*;
-pub use crate::arm64::*;
-pub use crate::m68k::*;
-pub use crate::mips::*;
-pub use crate::ppc::*;
-pub use crate::riscv::*;
-pub use crate::s390x::*;
-pub use crate::sparc::*;
-pub use crate::tricore::*;
-pub use crate::unicorn_const::*;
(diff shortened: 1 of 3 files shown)


bindings/rust/src/lib.rs
```diff
@@ -746,6 +791,7 @@ impl<'a, D> Unicorn<'a, D> {
     }
 
     /// Add hook for x86 OUT instruction.
+    #[cfg(feature = "arch_x86")]
```
wtdcode, line 794: But where can users define these features?

lockbox: That's already built-in the the features in `Cargo.toml`, by default they are all enabled (via the `arch_all` feature enabling all of the arch features). Right now only the `build.rs` uses it when compiling unicorn, this just extends it so we aren't compiling rust code for c code that already got filtered out

lockbox: https://github.com/unicorn-engine/unicorn/blob/dev/Cargo.toml#L43

lockbox: (That's why the tests still work, by default all architectures are enabled)

wtdcode: Ohhhhhh, sry I was looking at master branch.

wtdcode:
Merged, thanks!

PR #1942, opened 2024-04-24
Add unicorn_py3 folder to setup.py

nganhkhoa:
New `unicorn_py3` folder is not installed when running `setup.py install`, making unicorn unable to be imported (`import unicorn`) when using Python 3, see error message attached. Double asterisk for recursiveness because inside the folder has `arch` folder also.

Without `unicorn_py3/*`
image

Without double asterisk
image


wtdcode:
I guess it's due to not using `pip install .`

PR #1945, opened 2024-04-27
Remove glib_compat

### LINKED ISSUE #1943: glib_compat breaks Qt Widgets with glib backend
When unicorn is linked in program that uses Qt Widgets (that uses glib internally),
the glib_compat inside breaks resulting binary (Qt wrongly uses glib_compat and crashes).

I removed glib_compat from unicorn completely and pointed it to system glib,
everythinks works OK now (Qt works and unicorn works).

Is it possible to prevent such library incompatibilities?
Maybe rename glib_compat functions to not collide with glib, or remove glib_compat completely (as I did).

Why is it there at all?

vlasta-labsky:
Fixes #1943


aquynh:
The original reason was to build Unicorn everywhere, including Windows. Do
you try there yet?

vlasta-labsky:
I haven't tried windows, it may took a while for me to prepare windows build environment (not windows user) ...
It looks that upstream glib supports win builds, so maybe it works just fine.
The issue #1943 that I want to solve is less likely to happen on windows - Qt for windows does not use glib,
and I think there are not many programs/libraries that use glib on windows.
The best option would be to eliminate glib dependency (even the system one) but that it beyond my capabilities.

wtdcode:
As suggested above, this breaks windows. See CI.

vlasta-labsky:
I looked at the CI build errors and they are related to missing pkgconfig and/or glib upstream packages - this patch introduces new dependency on upstream glib to replace glib_compat.

PR #1963, opened 2024-06-06
Optimize memory handling

PhilippTakacs:
By using the memory snapshots with real examples we found some bottlenecks. Therefor I have implemented some optimizations.

We found that the flatview creation is quite expensive (around 1/4 of all runtime including emulation) when you have a lot of memory regions. By permanently creating new regions with CoW we have a lot of regions and rebuild for each cow the flatview again. I have added a update function does only change affected ranges.

Also we found that the `find_ram_offset_last` optimization doesn't work after a restore. To fix this the context now includes also the `ramblock_freed` bool.

This also includes two smaller optimizations: only clear the affected tlb on memory_cow and save the last `ramblock` of the `ram_list`

It's currently only a draft, because we need to do a few more tests to see if the optimization works as expected.

### DIFF
--- a/qemu/exec.c
+++ b/qemu/exec.c
@@ -957,12 +957,10 @@ void flatview_add_to_dispatch(struct uc_struct *uc, FlatView *fv, MemoryRegionSe
 
 static ram_addr_t find_ram_offset_last(struct uc_struct *uc, ram_addr_t size)
 {
-    RAMBlock *block;
     ram_addr_t result = 0;
+    RAMBlock *block = uc->ram_list.last_block;
 
-    RAMBLOCK_FOREACH(block) {
-        result = MAX(block->offset + block->max_length, result);
-    }
+    result = block->offset + block->max_length;
 
     if (result + size > RAM_ADDR_MAX) {
         abort();
@@ -1076,18 +1074,26 @@ static void ram_block_add(struct uc_struct *uc, RAMBlock *new_block)
      * QLIST (which has an RCU-friendly variant) does not have insertion at
      * tail, so save the last element in last_block.
      */
-    RAMBLOCK_FOREACH(block) {
-        last_block = block;
-        if (block->max_length < new_block->max_length) {
-            break;
+    if (new_block->max_length > uc->target_page_size) {
+        RAMBLOCK_FOREACH(block) {
+            last_block = block;
+            if (block->max_length < new_block->max_length) {
+                break;
+            }
         }
+    } else {
+        last_block = uc->ram_list.last_block;
+        block = NULL;
     }
+
     if (block) {
         QLIST_INSERT_BEFORE_RCU(block, new_block, next);
     } else if (last_block) {
         QLIST_INSERT_AFTER_RCU(last_block, new_block, next);
+        uc->ram_list.last_block = new_block;
     } else { /* list is empty */
         QLIST_INSERT_HEAD_RCU(&uc->ram_list.blocks, new_block, next);
+        uc->ram_list.last_block = new_block;
     }
     uc->ram_list.mru_block = NULL;
 
--- a/qemu/exec.c
+++ b/qemu/exec.c
@@ -1074,7 +1074,7 @@ static void ram_block_add(struct uc_struct *uc, RAMBlock *new_block)
      * QLIST (which has an RCU-friendly variant) does not have insertion at
      * tail, so save the last element in last_block.
      */
-    if (new_block->max_length > uc->target_page_size) {
+    if (uc->ram_list.freed || new_block->max_length > uc->target_page_size) {
         RAMBLOCK_FOREACH(block) {
             last_block = block;
             if (block->max_length < new_block->max_length) {
@@ -1171,6 +1171,7 @@ void qemu_ram_free(struct uc_struct *uc, RAMBlock *block)
     QLIST_REMOVE_RCU(block, next);
     uc->ram_list.mru_block = NULL;
     uc->ram_list.freed = true;
+    uc->ram_list.last_block = NULL;
     /* Write list before version */
     //smp_wmb();
     // call_rcu(block, reclaim_ramblock, rcu);
--- a/qemu/exec.c
+++ b/qemu/exec.c
@@ -1455,6 +1455,13 @@ AddressSpaceDispatch *address_space_dispatch_new(struct uc_struct *uc, FlatView 
     return d;
 }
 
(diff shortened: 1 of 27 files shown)


PhilippTakacs:
I have additional added the flatview to the context. This way flatview doesn't need to be rebuild on `context_restore`.

I have tested and profiled this. It does in fact save us relevant time.

wtdcode:
> @wtdcode have you overseen this PR? It would be nice to get some feedback.

Sorry for not leaving a comment. I had a quick review on it and didn't find any obvious problems. I planned to have further debugging to have a better understanding before commenting further.

wtdcode:
I ported d01035767e47f69b4e545d1843dbaf08e6a74752 which generally allows unicorn to go fast `qemu_ld` path (not going through `store_helper` again) to have better writing performance. This shall have side effects on your `memory_cow` which relies on `store_helper`.

I have twinkled a bit on the code to ensure that when snapshot is enabled, no TLB is marked clean. However, I'm not sure if that works for you, you may have a look.

Regarding this PR, I locally tested and think it is good to go if you don't have more to add.

PhilippTakacs:
Thanks for the update. As far as I understand this patch allows faster writes, if the tlb cache entry exists. So for snapshots it might be enough to clear the tlb cache. This way the store_helper would be called again and everything works as expected. I'll test this.

> think it is good to go if you don't have more to add

I would like to rebase this to the dev branch to remove the "fixup" commits. Also because I haven't looked at this for a few weeks, I would like to test this again with the current HEAD. I'll ping you when I'm finished testing.

PhilippTakacs:
@wtdcode I have tested the code again and removed the fixup commits so for me this is ready to merge.

For the notdirty write I'm working on a optimization, but I'll open another PR for it.

wtdcode:
> @wtdcode I have tested the code again and removed the fixup commits so for me this is ready to merge.
> 
> For the notdirty write I'm working on a optimization, but I'll open another PR for it.

Will check today or tomorrow, thanks for your work!

uc.c
```diff
@@ -2418,6 +2431,11 @@ uc_err uc_context_restore(uc_engine *uc, uc_context *context)
             return ret;
         }
         uc_snapshot(uc);
+        uc->ram_list.freed = context->ramblock_freed;
```
wtdcode, line 2434: Sorry I should have noticed this earlier. With this change (and actually the early `snapshot_level` change), with `UC_CTL_CONTEXT_MEMORY`, the context is no longer safe to keep as a file (it contains pointers) and restored to another instance.

It is super non-trivial to adapt the memory context to a value type and thus I suggest adding a few documents in the header (`uc_context_save/restore`, `UC_CTL_CONTEXT_MEMORY`).

qemu/exec.c
```diff
@@ -1076,18 +1074,26 @@ static void ram_block_add(struct uc_struct *uc, RAMBlock *new_block)
      * QLIST (which has an RCU-friendly variant) does not have insertion at
      * tail, so save the last element in last_block.
      */
-    RAMBLOCK_FOREACH(block) {
-        last_block = block;
-        if (block->max_length < new_block->max_length) {
-            break;
+    if (uc->ram_list.freed || new_block->max_length > uc->target_page_size) {
```
wtdcode, line 1077: I fail to understand the reason of `new_block->max_length > uc->target_page_size` check. Why in such case you can't append ram blocks?

PhilippTakacs: The `ram_list` is ordered by the size of the ram block, so adding a block with `new_block->max_length > uc->target_page_size` would not necessary added at the end of the list. A future optimization could add an comparison of the max_length of the last block in the list and the new block.

wtdcode: Ohh, I see. Thanks for the explanation.

uc.c
```diff
@@ -2434,6 +2452,10 @@ uc_err uc_context_restore(uc_engine *uc, uc_context *context)
 UNICORN_EXPORT
 uc_err uc_context_free(uc_context *context)
 {
+    if (context->fv) {
+        free(context->fv->ranges);
+    }
+    g_free(context->fv);
```
wtdcode, line 2458: I think this shall move to the `if (context->fv)` branch.

qemu/softmmu/memory.c
```diff
@@ -196,19 +195,26 @@ MemoryRegion *memory_map_io(struct uc_struct *uc, ram_addr_t begin, size_t size,
...
         if (subregion->priority >= level) {
-            memory_region_del_subregion(mr, subregion);
+            memory_region_remove_subregion(mr, subregion);
             subregion->destructor(subregion);
             g_free(subregion);
-            mr->uc->memory_region_update_pending = true;
         }
     }
-    memory_region_transaction_commit(mr);
+  //  memory_region_transaction_commit(mr);
```
wtdcode, line 217: Is it safe to remove this commit?

uc.c
PhilippTakacs: Yes, we had this question already in the snapshots feature. I'll write some documentation about this.

qemu/softmmu/memory.c
PhilippTakacs: Yes. This function is called in two cases.

* unmap/free complete memory region (as part of destroy)
Then the parrent unmap will trigger the commit
* restore the memory state
Then the restore also restores the old flatview so no rebuild is nesecary

uc.c
PhilippTakacs: g_free() and free() with a NULL is a noop so it's not necesary. I can move it if you prefer it this way better.

wtdcode: Yes, please do so because it looks more straightforward.

qemu/softmmu/memory.c
wtdcode: I did a quick check, and it seems to make sense.

Could you also add these as comments here? So that I know the preconditions here if we are going to bump qemu version.

wtdcode:
Okay, I finished the review, and it's safe to go now with minor fixes and a few document updates. Thanks again for your brilliant work.

PhilippTakacs:
I have added the last fixes and rebased it to the dev branch.

The failed test doesn't look like it's a bug by my changes.

wtdcode:
> I have added the last fixes and rebased it to the dev branch.
> 
> The failed test doesn't look like it's a bug by my changes.

I will rerun and it seems Github Action bug.

wtdcode:
Thanks!

PR #1964, opened 2024-06-12
 Remove using of distutils

### LINKED ISSUE #1939: distutils deprecation
In Python 3.12, distutils is deprecated. Can you edit all instances of importing distutils into something that doesn't use distutils?

bet4it:
Resolve #1939


wtdcode:
Thanks!

PR #1966, opened 2024-06-22
Fix emulator detection

mrexodia:
References:
- https://gynvael.coldwind.pl/?id=268
- https://github.com/JonathanSalwan/Triton/issues/1131

This fixes the `bswap ax` detection as well as various other detections related to bad REX prefix handling. This bug is also still upstream in qemu, but they moved the code around quite a bit:

https://github.com/qemu/qemu/blob/ffeddb979400b1580ad28acbee09b6f971c3912d/target/i386/tcg/decode-new.c.inc#L2561-L2570

There are two fixes here:
- `bswap ax` now correctly sets `ax` to zero
- REX prefixes are correctly ignored when appropriate (they only apply when they are right next to the first byte of the instruction)

### DIFF
--- a/qemu/target/i386/translate.c
+++ b/qemu/target/i386/translate.c
@@ -4792,12 +4792,12 @@ static target_ulong disas_insn(DisasContext *s, CPUState *cpu)
 {
     TCGContext *tcg_ctx = s->uc->tcg_ctx;
     CPUX86State *env = cpu->env_ptr;
-    int b, prefixes;
+    int b, prefixes, prefix_count;
     int shift;
     MemOp ot, aflag, dflag;
     int modrm, reg, rm, mod, op, opreg, val;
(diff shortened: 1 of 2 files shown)


mrexodia:
The build errors on macos are also present on the main branch. No idea about the crash for MINGW64, help would be appreciated!

wtdcode:
MINGW64 errors is just random crashing which I can’t reproduce locally.

macOS is probably due to system upgraded.

wtdcode:
I will fix CI before 2.1.0. Let this PR get in firstly.

wtdcode:
You code accidentally trigger a long-live known mingw bug which causes random segfaults on Windows, which was finally fixed in a5fa5480490d37917e228976320e5b8a2f30b14d

PR #1975, opened 2024-07-14
Update github workflows

scribam:
First step to fix CI with:
- all warnings fixed
- "android cmake - gcc" fixed


wtdcode:
Sorry for the delay. Thanks!

PR #1977, opened 2024-07-16
gcc compiler warning fixes

BitMaskMixer:
The gcc compiler version 13.2.0 complains about:
- unused variables in two macros (due to casting to void* instead of void)
- enum-int-mismatch in a method parameter

This small PR fixes these issues so the build log is warning and error free for gcc - under linux.


wtdcode:
This should have been fixed as requested by aquynh. Probably I left our my local changes.

wtdcode:
Thanks!

Issue #1971, opened 2024-07-08
Instrumenting ARM code making use of hardfault handlers and especially CFSR and BFAR registers


r3-ck0:
Hi,

During a CTF I was debugging an ARM binary that was using hardfault handlers and I was having a really difficult time, which was mostly skill issues, but I came to a spot that I think might not be my own fault anymore.

I'm providing a script that might or might not give a seasoned unicorn veteran aneurysms, sorry in advance, I hope you don't live in the US and can afford treatment. The binary in question is here: https://github.com/DownUnderCTF/Challenges_2024_Public/tree/main/hardware/crash-landing

I can run the code up to the point where the hardfault handler is triggered. The binary is then supposed to copy the value from the flag into the R3 register, based on the value in BFAR. However, when I run the below code, CFSR is empty, although it should, according to the code, contain 0x8200.

Is this a bug? Is this something on QEMU? Or is this a skill issue on my side?

`
from __future__ import print_function
from unicorn import *
from unicorn.arm_const import *
from capstone import *
import struct

def pretty_print_registers(uc):
    # List of ARM registers for pretty printing
    registers = [
        UC_ARM_REG_R0, UC_ARM_REG_R1, UC_ARM_REG_R2, UC_ARM_REG_R3,
        UC_ARM_REG_R4, UC_ARM_REG_R5, UC_ARM_REG_R6, UC_ARM_REG_R7,
        UC_ARM_REG_R8, UC_ARM_REG_R9, UC_ARM_REG_R10, UC_ARM_REG_R11,
        UC_ARM_REG_R12, UC_ARM_REG_SP, UC_ARM_REG_LR, UC_ARM_REG_PC,
        UC_ARM_REG_CPSR
    ]
    
    # Register names for display purposes
    register_names = [
        "R0", "R1", "R2", "R3", "R4", "R5", "R6", "R7",
        "R8", "R9", "R10", "R11", "R12", "SP", "LR", "PC",
        "CPSR"
    ]
    
    print("Current Register States:")
    for reg, name in zip(registers, register_names):
        # Read the register value
        value = uc.reg_read(reg)
        # Print the register name and its value in hexadecimal format
        print(f"{name}: 0x{value:08X}")

# code to be emulated
ARM_CODE   = None

with open("./crash_landing.bin", "rb") as f:
    ARM_CODE = f.read()

# memory address where emulation starts
ADDRESS       = 0x8000000
[...]

PR #1978, opened 2024-07-17
Revive qemu logs - and extend the FAQ

BitMaskMixer:
The qemu logs can be quite useful to understand, why the emulation does (not) work.

As an example, there is an open issue I had looked in (because I was curious) and did not find an easy way to
see the qemu logging - till I understood there arent anything compiled in.

I could write that script in cpp code and debug it - but I would like to have a "quick" way to figure out whats going on
with pyhton scripts too.

This PR revives the qemu_log for a "special" build - where you might want to try to find the bug or whats going on.

Example: Take the issue https://github.com/unicorn-engine/unicorn/issues/1971 and try to figure out whats going on with this PR and the enabled logs. You will see a message like:

image

It's more like to help developers to print useful log messages so we can figure out what's going on faster.

In the future, the logging should be used more and more so we gain back knowledge about things pretty easly.

### DIFF
--- a/qemu/include/qemu/log.h
+++ b/qemu/include/qemu/log.h
@@ -30,20 +30,207 @@
 
 /* Logging functions: */
 
-/* log only if a bit is set on the current loglevel mask:
- * @mask: bit to check in the mask
- * @fmt: printf-style format string
- * @args: optional arguments for format string
- */
-#define qemu_log_mask(MASK, FMT, ...)
+/* To verbose logging, enable the next line. */
+//#define LOGGING_ENABLED // to enable logging
 
-/* log only if a bit is set on the current loglevel mask
- * and we are in the address range we care about:
- * @mask: bit to check in the mask
- * @addr: address to check in dfilter
- * @fmt: printf-style format string
- * @args: optional arguments for format string
- */
-#define qemu_log_mask_and_addr(MASK, ADDR, FMT, ...)
+#ifdef LOGGING_ENABLED
 
+#include <stdint.h>
+#include <string.h>
+#include <stdlib.h>
+#include <stdarg.h>
+#include <stdio.h>
+
+/**
+ * Reads the @p env_name and tries to parse the value into an uint32_t.
+ * @param env_name The environment variable name to parse.
+ * @return The parsed value.
+ ** 0 in case if the value can not be parsed (or is 0).
+ ** ULONG_MAX if the value is bigger than an uint32_t.
+ */
+static inline uint32_t read_and_parse_env(const char* env_name)
+{
+    uint32_t data = 0;
+    const char* env_data = getenv(env_name);
+
+    if (env_data != NULL) {
+        char buffer[11] = {0}; // 0xFFFF'FFFF\0
+        strncpy(buffer, env_data, sizeof(buffer) - 1);
+        data = (uint32_t)strtoul(buffer, NULL, 0);
+    }
+
+    return data;
+}
+
+/**
+ * Gets the log level by reading it once.
+ * @return The log level.
+ */
+static inline uint32_t get_log_level()
+{
+    static uint64_t log_level = UINT64_MAX;
+
+    if (log_level == UINT64_MAX) {
+        log_level = read_and_parse_env("UNICORN_LOG_LEVEL");
+    }
+
+    return (uint32_t)log_level;
+}
+
+/**
+ * Gets the log detail level by reading it once.
+ * @return The detail log level.
+ */
+static inline uint32_t get_log_detail_level()
+{
+    static uint64_t log_detail_level = UINT64_MAX;
+
+    if (log_detail_level == UINT64_MAX) {
+        log_detail_level = read_and_parse_env("UNICORN_LOG_DETAIL_LEVEL");
+    }
+
+    return (uint32_t)log_detail_level;
+}
+
+/**
+ * Checks if the @p log_level is active.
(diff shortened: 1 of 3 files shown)


wtdcode:
The idea is good and I planned to do this for a while. However, the logging functionality doesn't have a switch in cmake to be enabled. Meanwhile, there is a magic environment variables designed for this: `UNICORN_DEBUG`. We can do it like:

- Only build logging support when `NDEBUG` is not defined, i.e. debug builds.
- Only show debug output when `UNICORN_DEBUG` env is set.

So the logging will be only shown for developers.

BitMaskMixer:
Cool - good to know :)

---
> * Only build logging support when `NDEBUG` is not defined, i.e. debug builds.

I am not quite sure if NDEBUG is the right approach, as we assume now that only DEBUG builds 
are used to get the details.
The executing debug code might take another path which did not trigger the bug, 
so I wanted to have an independent way to enable the logs.

I wanted to add that to the CMakefiles, but - they are quite huge.
For my understanding, you can not pass preprocessor defines via -DXXX directly during the call to cmake, you must
add internal cmake-variables which caches the value first and add them to the desired target.

The target_compile_definitions is the way to add that to the right target I think, so I do not want to mix up the CMAKE_CXX_FLAGS and CMAKE_C_FLAGS flags directly.

---

> * Only show debug output when `UNICORN_DEBUG` env is set.

That's nice, but I prefer reading it once during instancing UC (or let it automatically be called when a log is processed). 
Not sure how fast the "getenv" call is, but sure, you are about to loose a lot of speed for 
unneeded repeated calls over. 

What do you think? Do we need a way to check again and again if the logging has been activated
 at some time during execution - which costs performance?
(Quite easy to implement - see below. Remove that static keyword)

For reading it once, storing it into a static variable would be a way to go I guess.

Something like:

```cpp
bool logging_enabled()
{
     static const char* env_data = getenv("UNICORN_DEBUG");
     
     return env_data != NULL;
}
```

Or something, which checks the pointer and try to parse the value behind the env_data. (more complex, not needed?)

BitMaskMixer:
@wtdcode Is there an easy way to execute the pipeline locally?
Like building up the needed docker containers - and execute the whole pipeline with them? 

I would dare to ask you to trigger it again if I could do that locally - but ... here we go.
Issues might be fixed, till the next compiler / tool complains ;-P

wtdcode:
Sorry for late.

> @wtdcode Is there an easy way to execute the pipeline locally? Like building up the needed docker containers - and execute the whole pipeline with them?
> 
> I would not dare to ask you to trigger it again if I could do that locally - but ... here we go. Issues might be fixed, till the next compiler / tool complains ;-P

Personally, I used https://github.com/nektos/act to do this.

qemu/include/qemu/log.h
```diff
@@ -30,20 +30,207 @@
 
 /* Logging functions: */
 
-/* log only if a bit is set on the current loglevel mask:
- * @mask: bit to check in the mask
- * @fmt: printf-style format string
- * @args: optional arguments for format string
+/* To verbose logging, enable the next line. */
+//#define LOGGING_ENABLED // to enable logging
```
wtdcode, line 34: We can add a cmake variable for this. If you are not familiar, I can do this for you.

qemu/tcg/tcg.c
```diff
@@ -3728,7 +3728,7 @@ int tcg_gen_code(TCGContext *s, TranslationBlock *tb)
     TCGOp *op;
 
 #ifndef NDEBUG
-    if (getenv("UNICORN_DEBUG")) {
+    if (is_log_level_active(CPU_LOG_TB_IN_ASM)) {
```
wtdcode, line 3731: is `CPU_LOG_TB_IN_ASM` really defined somewhere?

BitMaskMixer: Yes it is defined in the "log.h" file, where I have implemented the logging. 
If it's not defined, a compiler error would be generated.

Not sure if that is the right log level bit mask, as it's very very verbose. Feel free to change it, but I thougt it fits the most.

In the future, we might want to add Unicorn specific log levels (and macros) to distingiush between qemu und unicorn logs. (Just a quick popped up idea)

wtdcode: Ohhh yes, github folded the definitions.

BitMaskMixer:
> Personally, I used https://github.com/nektos/act to do this.

Yes I have seen that project too - but I did not used it before. 
A dockerfile which wraps all dependencies would be pretty nice, however, I am not sure how the (cross) compiling for other platforms will work, as you have MSVC and Apple builds going on.

Not familiar with the github logic or terminology, but are there "shareable and free" "git-runners" available, which might execute the pipeline for a PR automatically? 
The code is public anyway, so there is no issue with "leaking source" code I think.

wtdcode:
> > Personally, I used https://github.com/nektos/act to do this.
> 
> Yes I have seen that project too - but I did not used it before. A dockerfile which wraps all dependencies would be pretty nice, however, I am not sure how the (cross) compiling for other platforms will work, as you have MSVC and Apple builds going on.
> 
> Not familiar with the github logic or terminology, but are there "shareable and free" "git-runners" available, which might execute the pipeline (workflow) for a PR automatically? The code is public anyway, so there is no issue with "leaking source code" I think.

In that way, you need to fork it and enable actions in your forks.

wtdcode:
Merged, nice work!

PR #2000, opened 2024-09-14
Add py.typed to the Python binding package

omer54463:
It seems that the python bindings have great type hints, but due to the lack of a `py.typed` file, `mypy` doesn't use them.
This PR solves the issue.


omer54463:
I now realize I created a pull request into master, which is probably not what I should've done.
I'm not sure whether I should create a PR into dev or next or another branch. Any help would be appreciated :)

wtdcode:
Yes, this shall go do dev branch.

PR #2003, opened 2024-09-19
Misc. Python binding re-arrangements

elicn:
Highlights:
 - Miscellaneous re-arrangements following dual Python support
 - Add initial support for batch operations


wtdcode:
LGTM, here you go!

PR #2005, opened 2024-09-20
Add py.typed to the Python binding package

omer54463:
It seems that the python bindings have great type hints, but due to the lack of a py.typed file, mypy doesn't use them.
This PR solves the issue.

I previously created #2000 from my master into this repo's master instead of dev.


wtdcode:
Seems harmless, thanks!

PR #2011, opened 2024-09-25
Uc hook tcg improve

dotCirill:
Improve UC_HOOK_TCG_OPCODE feature.

Currently UC_HOOK_TCG_OPCODE is supported only for x86.

Such hooks can be usefull for implementing CMPLOG in unicorn-afl.


wtdcode:
The history seems weird, I think you should target dev branch.

dotCirill:
Yes, target should be dev branch.

wtdcode:
Looks good, thanks.

wtdcode:
I regret not checking the code rigorously, and unfortunately the PR has several flaws.

- `address` parameter is not correctly implemented. It should point to the instruction instead of the stale register value (you were referring to the _runtime_ value during _translation_ stage.
-  it is incorrect to convert `TCGv_i32` to `TCGv_i64` directly. It is UB and an extension instruction is needed.

But anyway, it's not your fault and thus I fixed them by myself already. Thanks for your contributions.

PR #2014, opened 2024-09-28
Add support for more Intel registers

elicn:
Highlights:
 - Added read and write support for CR8, though changes to this reg have no effect currently
 - Fixed a bug in which CR3 got updated whenever CR1 and CR2 were written
 - Initial support for AVX512
   - Added read and write support for XMM8-15 (all modes) and XMM16-31 (64-bit mode)
   - Added read and write support for YMM8-15 (all modes) and YMM16-31 (64-bit mode)
   - Added read and write support for ZMM0-31 (64-bit mode)
   - Writes to XMMn were tested to be reflected in YMMn and ZMMn and vice versa


wtdcode:
LGTM.

PR #2016, opened 2024-09-29
Support additional API on Python 3 bindings

elicn:
Highlights:
 - Redesign certain architecture-specific aspects to better leverage OOP
   - Code de-duplication
   - Easier to extend or support new architectures
 - Now fully support `reg_write_batch` and `reg_read_batch` API
 - Introduced coprocessor register accessors to AArch32 and AArch64, named `cpr_read` and `cpr_write`
 - 1 ARM regression test fixed
 - More comments, documentation and annotations

### DIFF
--- a/bindings/python/unicorn/unicorn_py3/arch/arm.py
+++ b/bindings/python/unicorn/unicorn_py3/arch/arm.py
@@ -50,39 +50,11 @@ class UcAArch32(Uc):
         """
 
         reg_class = (
-(UcAArch32.REG_RANGE_CP, UcRegCP),
+            (UcAArch32.REG_RANGE_CP, UcRegCP),
             (UcAArch32.REG_RANGE_Q, UcReg128)
         )
 
         return next((c for rng, c in reg_class if reg_id in rng), cls._DEFAULT_REGTYPE)
 
-    def reg_read(self, reg_id: int, aux: Any = None):
-        # select register class for special cases
-        reg_cls = UcAArch32.__select_reg_class(reg_id)
-
-        if reg_cls is None:
-            if reg_id == const.UC_ARM_REG_CP_REG:
-                return self._reg_read(reg_id, UcRegCP, *aux)
-
-            else:
-                # fallback to default reading method
-                return super().reg_read(reg_id, aux)
-
-        return self._reg_read(reg_id, reg_cls)
-
-    def reg_write(self, reg_id: int, value) -> None:
-        # select register class for special cases
-        reg_cls = UcAArch32.__select_reg_class(reg_id)
-
-        if reg_cls is None:
-            if reg_id == const.UC_ARM_REG_CP_REG:
-                self._reg_write(reg_id, UcRegCP, value)
-
-            else:
-                # fallback to default writing method
-                super().reg_write(reg_id, value)
-
-        else:
-            self._reg_write(reg_id, reg_cls, value)
 
 __all__ = ['UcRegCP', 'UcAArch32']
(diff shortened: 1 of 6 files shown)


bindings/python/unicorn/unicorn_py3/arch/arm.py
```diff
@@ -40,46 +40,71 @@ class UcAArch32(Uc):
...
+            is_64   : indicates whether this is a 64-bit register
 
-        return self._reg_read(reg_id, reg_cls)
+        Returns: value of coprocessor register
+        """
 
-    def reg_write(self, reg_id: int, value) -> None:
-        # select register class for special cases
-        reg_cls = UcAArch32.__select_reg_class(reg_id)
+        assert check_maxbits(coproc, 4)
```
wtdcode, line 77: Why not make `cpr_read/write` return `uc_err` and avoid `assert` here by returning `UC_ERR_ARG`? Terminating the whole process seems too aggressive.

bindings/python/unicorn/unicorn_py3/unicorn.py
```diff
@@ -1227,11 +1359,17 @@ def _do_reg_read_batch(self, reglist, vallist, count) -> int:
 
         return uclib.uc_context_reg_read_batch(self._context, reglist, vallist, count)
 
+    def _do_reg_write_batch(self, reglist, vallist, count) -> int:
```
wtdcode, line 1362: Duplicate definition of line 762?

bindings/python/unicorn/unicorn_py3/arch/arm.py
elicn: Generally speaking, exceptions are thrown when the arguments normally make sense but there is a specific problem this time (e.g. opening a file which happens to be inaccessible, etc.). In these cases the user may catch the exception and choose how to handle it.

Here the assertions come to block an invalid usage with incorrect arguments: it is not a call that may sometimes work and sometimes not -- rather it is invalid. There is no way to "handle" such case other than re-writing that piece of code.

bindings/python/unicorn/unicorn_py3/unicorn.py
elicn: They belong to two different classes (engine and context). The underlying methods are using the same names to allow seamless OOP.

wtdcode: Okay I see

bindings/python/unicorn/unicorn_py3/arch/intel.py
```diff
@@ -127,67 +129,42 @@ def __hook_insn_cpuid_cb(uc: Uc, key: int) -> int:
...
 
-            super().reg_write(reg_id, value)
-        else:
-            self._reg_write(reg_id, reg_cls, value)
-
-    def msr_read(self, msr_id: int) -> int:
-        return self._reg_read(const.UC_X86_REG_MSR, UcRegMSR, msr_id)
+        return self.reg_read(const.UC_X86_REG_MSR, msr_id)
 
     def msr_write(self, msr_id: int, value: int) -> None:
```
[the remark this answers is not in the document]
elicn: I am not following on this one. What should raise an exception? It marks this line:
```python
return self.reg_read(const.UC_X86_REG_MSR, msr_id)
```

bindings/python/unicorn/unicorn_py3/arch/arm.py
wtdcode: Okay, I also forget `AssertionError` can be caught so let it go.

PR #2022, opened 2024-10-06
Revamp Python regression tests suite

elicn:
Highlights:
 - All Python regression tests are now collectively managed and dispatched by a single centralized module called `regress.py`
   - Heavily modified most of the Python regression tests to adhere to `unittest` conventions
   - Added `regress.logger` to allow emitting logging output by log level
   - Adding new test is done simply by creating a new .py file in the same directory
   - Each test is standalone and may be ran directly
   - New test suite _should_ support Python 2, but this was not tested

Known failures:
 - Modifying x86 EFLAGS will not take effect unless re-writing EIP/RIP (`jumping.py`)
 - Mapping memory to high addresses will fail on use rather than on map (`mem_high_mapping.py`)
 - Unicorn will segfault when calling `mem_write` to write to a location that was mapped only as `UC_PROT_READ` (`hook_readonly_write_local.py`)
 - Wrong exceptions raised, root cause is yet to be determined (`mips_except.py`)


wtdcode:
Thanks!

PR #2023, opened 2024-10-07
Python: Fix assertion failure on ctl_flush_tb()

droe:
Fix an assertion failure on encoding a ctl with no arguments.
Unbreaks `ctl_flush_tb()`.


wtdcode:
Thanks!

PR #2026, opened 2024-10-11
Python binding setup refactoring + cibuildwheel workflow

Antelox:
- Moved test scripts into tests folder to make them handy for _pytest_ (this is just a preliminary step to actually porting current tests to pytest-compliant. However it's enough as of now since it makes the job to test if the wheel is ok or not...)
- Updated _CMakeLists.txt_ in order to set _CMAKE_OSX_ARCHITECTURES_ var
- Moved project package settings to the new TOML format: **pyproject.toml**
- Refactored **setup.py** to cleanup/improve the code and make it ready for _cibuildwheel_
- Updated _README.md_ with the package long description part
- Updated _MANIFEST.in_ with new paths
- Removed _setup.cfg_ since universal wheel building will be deprecated soon
- Replaced old _PyPI-publishing.yml_ workflow with brand-new one based on _cibuildwheel_: Leveraged the GitHub matrix strategy in order to execute more build types in parallel rather one-by-one. This is crucial for the Linux _aarch64_ builds since they will be built within Ubuntu x64 and _cibuildwheel_ uses _QEMU_ under the hood (hence slow!). The workflow is configured in a way that it builds for each platforms/architectures just python 3.8 wheel, and all the remaining python versions wheels only for a tag release (for a matter of workflow time execution of course)
- Removed old building scripts
- Removed _.gdb_history_ file
- Updated _.gitignore_ to exclude PyCharm-related files/folders
- Replaced macos-12 runner with macos-13 since it will be removed soon
- Select build type according to the presence of tag release


wtdcode:
Finally, we have some modern infrastructure for building wheels. I hope this fixes various wheels-related issues.

bet4it:
I am curious about the advantages of cibuildwheel compared to the previous packaging methods. Previously, on Linux, a single package could be universally compatible with multiple Python versions, and the API interface of Unicorn was not complicated. Why is it necessary to create a separate package for each Python version?

wtdcode:
> I am curious about the advantages of cibuildwheel compared to the previous packaging methods. Previously, on Linux, a single package could be universally compatible with multiple Python versions, and the API interface of Unicorn was not complicated. Why is it necessary to create a separate package for each Python version?

That's how python wheels are intended to work (actually the whole ecosystem of python wheel is a mess...). And most importantly, cibuildwheel saves me from wasting hours on fixing CI.

Regarding Python API changes, it is related to #1629

PR #2028, opened 2024-10-11
qemu/tcg: fix UC_HOOK_MEM_READ on aarch64.

### LINKED ISSUE #1908: UC_HOOK_MEM_READ only triggered once (x86 64 emulated on mac M1)
Hello,

### Description 
I'm facing a strange issue where my UC_HOOK_MEM_READ is *only called the first time* a memory read is encountered.
All subsequent reads do not trigger the callback.

### Setup

- Unicorn commit d4b92485
- Host: Mac M1 running macOS 14 (Sonoma)
- Emulated code: x86 64 

Unicorn has been compiled as per recommendations in the [documentation](https://github.com/unicorn-engine/unicorn/blob/master/docs/COMPILE.md). 
Unicorn python bindings were installed in a fresh python venv.

### Test case
```py
from unicorn import *

# mov rax, [0x10000]
# mov [0x10000], rax
code = b"\x48\xa1\x00\x00\x01\x00\x00\x00\x00\x00\x48\xa3\x00\x00\x01\x00\x00\x00\x00\x00"*10

# callback for tracing memory access (READ ony)
def hook_mem_access(uc, access, address, size, value, user_data):
    if access == UC_MEM_READ:
        print(f">>> Memory is being READ at {address:#x}, data size = {size}")
    else:
        print(f">>> Memory is being WRITE at {address:#x}, data size = {size}")

def test_x86_64_mem_read():
    print("Emulate x86 64-bit code")
    try:
    
        mu.hook_add(UC_HOOK_MEM_READ | UC_HOOK_MEM_WRITE, hook_mem_access)

        mu.mem_map(0x10000, 1024*8)
        mu.mem_write(0x10000, b"\xaa"*8)

        # write machine code to be emulated to memory
        mu.mem_map(0, 0x1000)
        mu.mem_write(0, code)


        # emulate machine code in infinite time
        mu.emu_start(0, len(code))

    except UcError as e:
        print("ERROR: %s" % e)


if __name__ == '__main__':
    test_x86_64_mem_read()

```

### Test case output
```
Emulate x86 64-bit code
>>> Memory is being READ at 0x10000, data size = 8
>>> Memory is being WRITE at 0x10000, data size = 8
>>> Memory is being WRITE at 0x10000, data size = 8
>>> Memory is being WRITE at 0x10000, data size = 8
>>> Memory is being WRITE at 0x10000, data size = 8
>>> Memory is being WRITE at 0x10000, data size = 8
>>> Memory is being WRITE at 0x10000, data size = 8
>>> Memory is being WRITE at 0x10000, data size = 8
>>> Memory is being WRITE at 0x10000, data size = 8
>>> Memory is being WRITE at 0x10000, data size = 8
>>> Memory is being WRITE at 0x10000, data size = 8
```

As can be seen above,  `hook_mem_access` is only called once for READ while it is expected to be called 10 times.
Strangely, this does not affect writes.

glennsec:
Directly jump into the slow path when there is any hookmem enabled. This fixes #1908.

### DIFF
--- a/qemu/include/tcg/tcg.h
+++ b/qemu/include/tcg/tcg.h
@@ -1577,4 +1577,11 @@ struct jit_code_entry {
 void uc_del_inline_hook(uc_engine *uc, struct hook *hk);
 void uc_add_inline_hook(uc_engine *uc, struct hook *hk, void** args, int args_len);
 
+static inline bool tcg_uc_has_hookmem(TCGContext *s)
+{
+    return HOOK_EXISTS(s->uc, UC_HOOK_MEM_READ) ||
+        HOOK_EXISTS(s->uc, UC_HOOK_MEM_READ_AFTER) ||
+        HOOK_EXISTS(s->uc, UC_HOOK_MEM_WRITE);
+}
+
 #endif /* TCG_H */
--- a/qemu/tcg/aarch64/tcg-target.inc.c
+++ b/qemu/tcg/aarch64/tcg-target.inc.c
@@ -1584,6 +1584,7 @@ static inline void tcg_out_adr(TCGContext *s, TCGReg rd, void *target)
 static inline bool has_hookmem(TCGContext *s)
 {
     return HOOK_EXISTS(s->uc, UC_HOOK_MEM_READ) ||
+        HOOK_EXISTS(s->uc, UC_HOOK_MEM_READ_AFTER) ||
         HOOK_EXISTS(s->uc, UC_HOOK_MEM_WRITE);
 }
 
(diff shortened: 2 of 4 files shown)


PhilippTakacs:
I think you missed at least `UC_HOOK_MEM_READ_AFTER`.

Also it would be nice to use HOOK_EXISTS_BOUNDED. This way the fast way is taken when no hook exists at the used address. But I don't know if this is possible at this place (you need the emulated physical address).

glennsec:
> I think you missed at least `UC_HOOK_MEM_READ_AFTER`.
> 
> Also it would be nice to use HOOK_EXISTS_BOUNDED. This way the fast way is taken when no hook exists at the used address. But I don't know if this is possible at this place (you need the emulated physical address).

Added check for `UC_HOOK_MEM_READ_AFTER`, including for i386. Factored out the code along the way. I am not sure either I can use `HOOK_EXISTS_BOUNDED()` as TCGContext.pc_start is set to the end of the BB at the point we call it for codegen.

droe:
fwiw, I'm using this fix as part of a bug fix stable branch on top of the `2.1.1` release.  It resolves #1908 for me on arm64 without any issues so far.

What is needed to get this merged?

wtdcode:
Sorry for the delay. Thanks for the contributions!

Issue #2029, opened 2024-10-12
unicorn 2.1.1: SMC doesn't work.


wallds:
```python
from unicorn import *
from unicorn.x86_const import *
from capstone import *
from capstone.x86_const import *
from capstone.x86 import *

CODE = b'\x48\x89\x05\x10\x00\x00\x00' # mov qword ptr [rip+0x10], rax
CODE += b'\x66\xc7\x05\x00\x00\x00\x00\x48\x05' # mov word ptr [rip], 0x0548;
CODE += b'\x8b\x80\x78\x56\x34\x12' # [orig] mov eax, dword ptr [rax + 0x12345678]; [after SMC] 480578563412 add rax, 0x12345678
CODE += b'\xf4' # hlt

def hook_code(uc: Uc, address, size, user_data):
    code = uc.mem_read(address, size)
    print(hex(address), size, code.hex(), '\t', end='')
    insn = list(cs.disasm(code, address, 1))
    if insn:
        insn = insn[0]
        print(insn.mnemonic, insn.op_str)
    else:
        print('[ERROR]: disasm')


def hook_mem_read(uc, access, address, size, value, user_data):
    print('[mem_read]:', access, hex(address), hex(value), size)


def hook_mem_write(uc, access, address, size, value, user_data):
    print('[mem_write]:', access, hex(address), hex(value), size)


def hook_mem_invalid(uc, access, address, size, value, user_data):
    print(f"[mem_invalid]:", hex(uc.reg_read(UC_X86_REG_RIP)), access, hex(address), size, value)
    return False


uc = Uc(UC_ARCH_X86, UC_MODE_64)  # unicorn 2.1.1
cs = Cs(CS_ARCH_X86, CS_MODE_64)  # capstone 5.0.3

CODE_BASE = 0x1000
STACK_BASE = 0x20000

uc.mem_map(CODE_BASE, 0x1000, UC_PROT_ALL)
uc.mem_map(STACK_BASE, 0x2000, UC_PROT_ALL)

uc.hook_add(UC_HOOK_CODE, hook_code)
if 'y' in input('add memory read/write hooks? (y/n) >>>').lower():
    uc.hook_add(UC_HOOK_MEM_READ, hook_mem_read)
    uc.hook_add(UC_HOOK_MEM_WRITE, hook_mem_write)
uc.hook_add(UC_HOOK_MEM_INVALID, hook_mem_invalid)

uc.mem_write(CODE_BASE, CODE)
[...]

PR #2031, opened 2024-10-14
Optimize Notdirty write

PhilippTakacs:
This enables `notdirty_write` when used with snapshots. To do this the tlb is cleared on a snapshot and the priority of the mr is compared with the snapshot level.

It also fixes a bug when the mmu doesn't map 1:1. Before this the memory region was looked up with the vaddr, but memory regions are mapped at the paddr.

It's currently a draft because I also want to improve the `MEM_HOOK` check. As you already noticed in #2029 there are some hooks missing in the check. I would also include a range check for the hooks so only skip the optimization when there is a hook on this address.

### DIFF
--- a/qemu/accel/tcg/cputlb.c
+++ b/qemu/accel/tcg/cputlb.c
@@ -1192,11 +1192,7 @@ static void notdirty_write(CPUState *cpu, vaddr mem_vaddr, unsigned size,
 {
     ram_addr_t ram_addr = mem_vaddr + iotlbentry->addr;
 
-    if (mr == NULL) {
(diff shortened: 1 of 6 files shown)


wtdcode:
Is it ready?

PhilippTakacs:
@wtdcode now it's ready

wtdcode:
> @wtdcode now it's ready

Cool, look pretty good to me and thanks a lot for adding tests. Would you like to amend your commit history?

PhilippTakacs:
The comment history is already in the way I would like it.

wtdcode:
Thanks!

PR #2034, opened 2024-10-17
Feat: Option to manually specify the build type in the workflow

Antelox:
- Fix upload-artifact actions in build wheel workflow
- Manually trigger workflow jobs to produce either Debug or Release builds


wtdcode:
Thanks!

PR #2036, opened 2024-10-18
Fix: Prevent Unnecessary Rebuilds Due to QEMU Config File Regeneration

JShweiri:
## Problem:
Currently, the Unicorn engine subproject is configured to regenerate QEMU's `config-host.h` and the architecture-specific `config-target.h` files every time CMake is run. This happens even if there haven't been any relevant changes to the project, leading to unnecessary rebuilds of the Unicorn library and any projects depending on it.

## Solution:
This PR introduces conditional regeneration of the QEMU config files. The logic now checks if the files already exist in the build directory. If they do, regeneration is skipped, saving valuable build time.

## Further Considerations:
While this PR addresses the most common case of unnecessary regeneration, there might be more subtle scenarios where we still want to regenerate the config files. Here are some potential approaches for future refinement:

- **Timestamp Comparison**: We could compare the timestamps of the input (e.g., `qemu/configure` script) and output (e.g., `config-host.h`) files. If the input file is newer, it suggests changes have been made that require regeneration.

- **Configuration Hashing**: A more robust approach would be to hash relevant configuration options and store the hash. If the configuration changes, the hash would change, triggering regeneration.

- **User-Controllable Option**: Provide a CMake option (e.g., `UNICORN_FORCE_REGEN`) to give users control over forcing regeneration when desired.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -343,99 +343,63 @@ else()
     set(TARGET_LIST "${TARGET_LIST} ")
 
     # GEN config-host.mak & target directories
-    execute_process(COMMAND sh ${CMAKE_CURRENT_SOURCE_DIR}/qemu/configure
-        --cc=${CMAKE_C_COMPILER}
-        ${EXTRA_CFLAGS}
-        ${TARGET_LIST}
-        WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
-    )
-    execute_process(COMMAND sh ${CMAKE_CURRENT_SOURCE_DIR}/qemu/scripts/create_config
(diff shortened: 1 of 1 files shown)


wtdcode:
Hey, thanks for your contribution but you shall send to our dev branch.

wtdcode:
I went through cmake documents and [ADDITIONAL_CLEAN_FILES](https://cmake.org/cmake/help/latest/prop_dir/ADDITIONAL_CLEAN_FILES.html#prop_dir:ADDITIONAL_CLEAN_FILES) seems to help easier regeneration. If we want to trigger a regeneration, we can simply `make clean`.

Could you also add the generated file to `ADDITIONAL_CLEAN_FILES`?

Issue #2035, opened 2024-10-18
Python Binding: Not yet support UC_HOOK_TLB_FILL?


DynamicLoader:
I'd installed version 2.1.1

When using hook_add, it just throw a exception. I digged into it and there is no impl of UC_HOOK_TLB_FILL.

```
        handlers: Mapping[int, Callable[[], Tuple]] = {
            uc.UC_HOOK_INTR               : __hook_intr,
            uc.UC_HOOK_INSN               : __hook_insn,
            uc.UC_HOOK_CODE               : __hook_code,
            uc.UC_HOOK_BLOCK              : __hook_code,
            uc.UC_HOOK_MEM_READ_UNMAPPED  : __hook_invalid_mem,
            uc.UC_HOOK_MEM_WRITE_UNMAPPED : __hook_invalid_mem,
            uc.UC_HOOK_MEM_FETCH_UNMAPPED : __hook_invalid_mem,
            uc.UC_HOOK_MEM_READ_PROT      : __hook_invalid_mem,
            uc.UC_HOOK_MEM_WRITE_PROT     : __hook_invalid_mem,
            uc.UC_HOOK_MEM_FETCH_PROT     : __hook_invalid_mem,
            uc.UC_HOOK_MEM_READ           : __hook_mem,
            uc.UC_HOOK_MEM_WRITE          : __hook_mem,
            uc.UC_HOOK_MEM_FETCH          : __hook_mem,
            # uc.UC_HOOK_MEM_READ_AFTER
            uc.UC_HOOK_INSN_INVALID       : __hook_invalid_insn,
            uc.UC_HOOK_EDGE_GENERATED     : __hook_edge_gen,
            uc.UC_HOOK_TCG_OPCODE         : __hook_tcg_opcode
        }
```

But the defination is in const...

wtdcode:
@elicn Would you like to add this? Or I can also help.

elicn:
Sure, let me look into this.

DynamicLoader:
Great to hear that! I'd compile it and try, but something just wrong....
I had set tlb mode to virtual and register the callback, like below:

```python
def hook_tlb_fill(uc : Uc, access, address, size, value, user_data):
    print(">>> TLB Fill at 0x%x, data size = %u" % (address, size))
    uc.ctl_flush_tb()

# <Some other init>

mu = Uc(UC_ARCH_RISCV, UC_MODE_64)
[...]

PR #2037, opened 2024-10-18
Hook tlb fill

elicn:
Highlights:
 - Add support for TLB fill hooks as requested in #2035 (_untested_)

Notes:
 - Requires sanity testing
 - Requires unit testing

### DIFF
--- a/bindings/python/unicorn/unicorn_py3/unicorn.py
+++ b/bindings/python/unicorn/unicorn_py3/unicorn.py
@@ -33,11 +33,11 @@ class UcTupledStruct(ctypes.Structure, Generic[VT]):
         return tuple(getattr(self, fname) for fname, *_ in self.__class__._fields_)  # type: ignore
 
 
-class _uc_mem_region(UcTupledStruct[MemRegionStruct]):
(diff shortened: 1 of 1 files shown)


wtdcode:
Thanks!

bindings/python/unicorn/unicorn_py3/unicorn.py
```diff
@@ -1092,6 +1104,13 @@ def __hook_tcg_op_cb(uc: Uc, address: int, arg1: int, arg2: int, size: int, key:
 
             return (__hook_tcg_op_cb, opcode, flags)
 
+        def __hok_tlb_fill():
+            @uccallback(self, HOOK_TLB_FILL_CFUNC)
+            def __hook_tlb_fill_cb(uc: Uc, vaddr: int, access: int, entry: ctypes._Pointer[uc_tlb_entry], key: int):
+                callback(uc, vaddr, access, entry.contents, user_data)
```
DynamicLoader, line 1110: should `return` it since the return value is bool, not void.

elicn: Fixed that in #2042

PR #2039, opened 2024-10-18
Modify python version check such that linters recognize it

omer54463:
This is sort of a follow up to #2005. Currently many type checkers have a hard time recognizing all the different ways you can compare versions with `sys.version_info` (it's a best-effort kind of thing).

I changed the comparison and made sure that:
1. It works for both python 3 and python 2.7.
2. Mypy and Pyright both understand the comparison and skip the python 2.7 implementation.

### DIFF
--- a/bindings/python/unicorn/unicorn.py
+++ b/bindings/python/unicorn/unicorn.py
@@ -8,7 +8,7 @@ from .unicorn_const import (
 
 __version__ = "%u.%u.%u" % (__MAJOR, __MINOR, __PATCH)
 
-if _sys.version_info.major == 2:
-    from .unicorn_py2 import *
-else:
+if _sys.version_info >= (3,):
     from .unicorn_py3 import *
+else:
+    from .unicorn_py2 import *
--- a/bindings/python/unicorn/unicorn_py2.py
+++ b/bindings/python/unicorn/unicorn_py2.py
@@ -23,8 +23,7 @@ ucsubclass = 0
 if not hasattr(sys.modules[__name__], "__file__"):
     __file__ = inspect.getfile(inspect.currentframe())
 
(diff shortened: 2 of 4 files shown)


elicn:
Personally I like less the half-tuple comparison (i.e. `(3, )`).
What wrong with the current method? I don't see any problems with MyPy or Pylint on my side.

bindings/python/unicorn/unicorn.py
```diff
@@ -8,7 +8,7 @@
 
 __version__ = "%u.%u.%u" % (__MAJOR, __MINOR, __PATCH)
 
-if _sys.version_info.major == 2:
-    from .unicorn_py2 import *
-else:
+if _sys.version_info >= (3,):
```
elicn, line 11: It could make sense if we were differentiating between different minor versions, but only comparing for major version this is less readable.

bindings/python/unicorn/unicorn_py2.py
```diff
@@ -23,8 +23,7 @@
 if not hasattr(sys.modules[__name__], "__file__"):
     __file__ = inspect.getfile(inspect.currentframe())
 
-_python2 = sys.version_info[0] < 3
-if _python2:
+if sys.version_info < (3,):
```
elicn, line 26: This piece of code is redundant altogether. Python 3 never gets here anyway.

tests/regress/init.py
```diff
@@ -9,7 +9,7 @@
 from unicorn.x86_const import *
 
 
-if sys.version_info.major == 2:
+if sys.version_info < (3,):
```
elicn, line 12: As mentioned above, this is less readable when comparing major version only.

wtdcode:
@omer54463 Is it intended to close this?

PR #2042, opened 2024-10-22
Fix TLB fill hook

elicn:
Have TLB fill hook return a `bool` as it should, following [this](https://github.com/unicorn-engine/unicorn/pull/2037/files#r1810834858) comment.


wtdcode:
Thanks!

PR #2049, opened 2024-11-02
Several bugfixes

tbodt:
Submitted all together because creating separate pull requests would be annoying


wtdcode:
LGTM, thanks and welcome!

Issue #1841, opened 2023-06-14
Access violation Exception on Visual Studio Debugger


hxm-cpp:
the call of uc_mem_map trigger an exception on a debugging session : 
```
>	a.exe!tcg_out_opc(TCGContext * s, int opc, int r, int rm, int x) Line 609	C
 	[Inline Frame] a.exe!tcg_out_push(TCGContext * reg, int) Line 1064	C
 	a.exe!tcg_target_qemu_prologue(TCGContext * s) Line 3679	C
 	a.exe!tcg_prologue_init_x86_64(TCGContext * s) Line 846	C
 	a.exe!tcg_exec_init_x86_64(uc_struct * uc, unsigned int tb_size) Line 1229	C
 	a.exe!machine_initialize(uc_struct * uc) Line 56	C
 	a.exe!uc_init(uc_struct * uc) Line 229	C
 	a.exe!uc_mem_map(uc_struct * uc, unsigned __int64 address, unsigned __int64 size, unsigned int perms) Line 1037	C
 	a.exe!main(int argc, char * * argv, char * * envp) Line 31	C++
 	[External Code]	
```
the lib was compiled in Release with debug infos on visual studio 2022 SDK 10.0.22621.0

it doesn't crash running standalone.

wtdcode:
Are you using the code from dev branch?

wtdcode:
It should be introduced by recent windows demand paging feature but I can’t reproduce it on my side. Do you have any special setup?
________________________________

ZehMatt:
I believe that this happens because a debugger is attached and in that case the debugger will be the first to see the exception before its passed to application, this is probably due to some hacky code introduced lately. I think you should be able to pass the exception to the application and it should continue running.

wtdcode:
Oh wait, ZehMatt is correct as now we only commit pages when it’s accessed in the first time which causes exceptions. However, the exceptions are handled safely and you can configure your debugger to ignore this type of exceptions and safely continue execution.

ZehMatt:
> Oh wait, ZehMatt is correct as now we only commit pages when it’s accessed in the first time which causes exceptions. However, the exceptions are handled safely and you can configure your debugger to ignore this type of exceptions and safely continue execution.

[...]

PR #2056, opened 2024-11-28
Fixing issues on Windows

wnxd:
Fixed the issue that when using only MEM_RESERVE to allocate memory under Windows, Segmentation fault may occur during read and write operations

### DIFF
--- a/qemu/accel/tcg/translate-all.c
+++ b/qemu/accel/tcg/translate-all.c
@@ -903,7 +903,7 @@ static LONG __fastcall code_gen_buffer_handler(PEXCEPTION_POINTERS ptr, struct u
 static inline void may_remove_handler(struct uc_struct *uc) {
     if (uc->seh_closure) {
         if (uc->seh_handle) {
-            RemoveVectoredExceptionHandler(uc->seh_handle);
+            RemoveVectoredContinueHandler(uc->seh_handle);
         }
         VirtualFree(uc->seh_closure, 0, MEM_RELEASE);
     }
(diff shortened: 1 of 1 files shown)


wtdcode:
Please read our FAQ first.

https://github.com/unicorn-engine/unicorn/wiki/FAQ#i-debug-my-application-but-soon-get-an-access-violation-inside-unicorn

#1841

wnxd:
I use cgo binding on Windows, and the same error #1841 appears when I compile and run release. It seems that it is not caused by the debugger attaching. After ida debugging and tracing, I found that VirtualAlloc only uses MEM_RESERVE to allocate memory, and it will return the address of the uninitialized segment.

wtdcode:
> I use cgo binding on Windows, and the same error #1841 appears when I compile and run release. It seems that it is not caused by the debugger attaching. After ida debugging and tracing, I found that VirtualAlloc only uses MEM_RESERVE to allocate memory, and it will return the address of the uninitialized segment.

That's what "reserve" means exactly: only allocating and committing the memory when being accessed.

For the IDA debugger, you need to send the exceptions to the program instead of handling them by IDA.

wtdcode:
Read my dev blog if you are interested in the technical details: https://blog.lazym.io/2024/09/22/mmap-and-Overcommit-memory-on-Windows/

wnxd:
I seem to understand that Unicorn uses veh to handle dynamic memory allocation

wnxd:
It is very likely that cgo caused veh to fail, so the cgo binding library of unicorn cannot be used at present

wtdcode:
cgo bindings lack maintenance for a while due to my lacking time but I have no idea why it would fail.

A temporary workaround is a compilation flag defined in `CMakeLists.txt` to control whether using `MEM_COMMIT`. If you wish, you could submit a PR for that.

wnxd:
Thank you very much, I will try to solve it first, if it doesn't work, I will use MEM_COMMIT

wnxd:
@wtdcode After I reviewed the Go source code and tested it, using AddVectoredContinueHandler seems to solve the early exit problem caused by Go's VectoredHandler

wtdcode:
I will close this because the fix is incorrect and not relevant.

If you have another patch for go bindings, feel free to fire another PR.

Issue #2059, opened 2024-12-04
RUST BINDINGS: uc_reg_{read,write}_batch for rust


hsa1as:
Hi,

I'd noticed that `uc_reg_read_batch` and `uc_reg_write_batch` do not have bindings in rust,

Can I open a PR to add these? 
My fix is here: https://github.com/unicorn-engine/unicorn/compare/dev...hsa1as:unicorn:dev

wtdcode:
Sure yes. You are highly welcome.
________________________________
From: Sai Ashwin ***@***.***>
Sent: Thursday, December 5, 2024 12:18:52 AM
To: unicorn-engine/unicorn ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [unicorn-engine/unicorn] RUST BINDINGS: uc_reg_{read,write}_batch for rust (Issue #2059)


Hi,

I'd noticed that uc_reg_read_batch and uc_reg_write_batch do not have bindings in rust,

Can I open a PR to add these?
My fix is here: dev...hsa1as:unicorn:dev<https://github.com/unicorn-engine/unicorn/compare/dev...hsa1as:unicorn:dev>

―

wtdcode:
Closing due to fixed already.

PR #2060, opened 2024-12-04
Rust Bindings: reg_read_batch and reg_write_batch 

hsa1as:
This is my proposed fix for adding reg_read_batch and reg_write_batch into the rust bindings. Issue #2059 

It's a bit janky as uc_reg_{read,write}_batch require the values to be passed as void** . Is there a better solution that I have overlooked?


wtdcode:
Thanks!

PR #2069, opened 2024-12-21
improve aarch64 feature gate in rust bindings

Sanae6:
Found out the arch_aarch64 feature flag doesn't actually enable the arm64 module, and it bothered me a bit. Changed arch_aarch64 to depend on arch_arm of requiring manually enabling both arch_arm and arch_aarch64 at the same time.


wtdcode:
Thanks!

PR #2072, opened 2024-12-26
Bindings/python workflow: Run regress tests + generic refactoring

Antelox:
- Bindings/python workflow
    - Added fullMode input in workflow_dispatch
    - Take decision whether to build either in debug or release mode and if to build for all python versions according to the commit message patterns
    - Set proper artifact names
    - Removed not needed steps
    - Compacted some steps in order to leverage more the matrix feature
    - Bumped cibuildwheel action to 2.22.0
    - Run actual regress tests in place of sample scripts
- Specify optional test install in pyproject.toml with proper requirements
- Derive package version from git tags
- Add GENERATORS env var support in setup.py to specify cmake generator and minor refactoring
- Minor cleanup/refactoring for the regress test suite
- Marked some regress tests with skipIf to skip them in case of old python versions
- Marked some failing regress tests to be checked with skipIf

### DIFF
--- a/tests/regress/arm_init_input_crash.py
+++ b/tests/regress/arm_init_input_crash.py
@@ -1,21 +1,21 @@
-#!/usr/bin/env python
 # Sample code for ARM of Unicorn. Nguyen Anh Quynh <aquynh@gmail.com>
 # Python sample ported by Loi Anh Tuan <loianhtuan@gmail.com>
-#
 
+
+import platform
 import regress
-
+import sys
+import unittest
 from unicorn import *
 from unicorn.arm_const import *
 
-
 # code to be emulated
 ARM_CODE = (
-    b"\x37\x00\xa0\xe3"		#  mov      r0, #0x37
-    b"\x03\x10\x42\xe0"		#  sub      r1, r2, r3
+    b"\x37\x00\xa0\xe3"  # mov      r0, #0x37
+    b"\x03\x10\x42\xe0"  # sub      r1, r2, r3
 )
 
-THUMB_CODE = b"\x83\xb0"    #  sub      sp, #0xc
+THUMB_CODE = b"\x83\xb0"  # sub      sp, #0xc
 
 # memory address where emulation starts
 ADDRESS = 0xF0000000
@@ -32,6 +32,7 @@ def hook_code(uc, address, size, user_data):
 
 
 class TestInitInputCrash(regress.RegressTest):
+    @unittest.skipIf(sys.platform == 'win32' or platform.machine().lower() not in ('x86_64', 'arm64'), 'TO BE CHECKED!')
     def test_arm(self):
         regress.logger.debug("Emulate ARM code")
 
@@ -42,7 +43,7 @@ class TestInitInputCrash(regress.RegressTest):
         mu.mem_map(ADDRESS, mem_size)
 
         stack_address = ADDRESS + mem_size
-        stack_size =  stack_address         # >>> here huge memory size
+        stack_size = stack_address  # >>> here huge memory size
         mu.mem_map(stack_address, stack_size)
 
         # write machine code to be emulated to memory
--- a/tests/regress/arm_memcpy_neon.py
+++ b/tests/regress/arm_memcpy_neon.py
@@ -1,44 +1,44 @@
-
 import regress
-
+import sys
+import unittest
 from unicorn import *
 from unicorn.arm_const import *
 
-
-SHELLCODE = bytes.fromhex(
-    '03 f0 8f e0'   # 0001F894        ADD         PC, PC, R3
-    '0d 07 21 f4'   # 0001F898        VLD1.8      {D0}, [R1]!
-    '0d 07 0c f4'   # 0001F89C        VST1.8      {D0}, [R12]!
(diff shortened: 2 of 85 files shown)


tests/regress/high_mem.py
```diff
@@ -23,8 +21,10 @@ def map_code_page(self, address, payload):
         self.uc.mem_map(address, 0x1000)
         self.uc.mem_write(address, payload)
 
+    @unittest.skipIf(sys.version_info < (3, 7), reason="requires python3.7 or higher")
     def test_virt_high_mapping(self):
-        """Mapping memory at high addresses should work when TLB mode
+        """
```
elicn, line 26: Not sure why this one was changed. PEP8 (and originally PEP 257) states that docstrings should start in the same lines with the quotes.

tests/regress/hook_add_crash.py
```diff
@@ -1,21 +1,15 @@
...
-def hook_mem_read_unmapped(mu, access, address, size, value, user_data):
-    pass
-
 class TestHook(regress.RegressTest):
     def test_excessive_hooks(self):
         mu = Uc(UC_ARCH_X86, UC_MODE_32)
 
         for _ in range(1337):
-            mu.hook_add(UC_HOOK_MEM_READ_UNMAPPED, hook_mem_read_unmapped)
+            mu.hook_add(UC_HOOK_MEM_READ_UNMAPPED, lambda *args, **kwargs: None)
```
elicn, line 12: Python linters strongly advise against `lambda` functions, and I am not sure there is a justification to deviate from this recommendation here. Also, this is not only a style change - but also functional: while the older implementation re-uses the same function, the new one creates a new `lambda` function for each iteration.

tests/regress/jumping.py
```diff
@@ -68,20 +66,20 @@ def hook_block(self, uc, address, size, _):
...
     def hook_code(self, uc, address, size, _):
         insn = uc.mem_read(address, size)
-        regress.logger.debug(">>> Tracing instruction at %#x : %s", address, binascii.hexlify(insn))
+        regress.logger.debug(">>> Tracing instruction at %#x : %s", address, insn.hex())
 
         regs = uc.reg_read_batch((
             UC_X86_REG_RAX, UC_X86_REG_RBX, UC_X86_REG_RCX, UC_X86_REG_RDX,
             UC_X86_REG_RSI, UC_X86_REG_RDI, UC_X86_REG_RBP, UC_X86_REG_RSP,
-            UC_X86_REG_R8,  UC_X86_REG_R9,  UC_X86_REG_R10, UC_X86_REG_R11,
+            UC_X86_REG_R8, UC_X86_REG_R9, UC_X86_REG_R10, UC_X86_REG_R11,
```
elicn, line 74: Yet another example for a case where blindly sticking to strict styling rules doesn't make things more readable.

tests/regress/regress.py
```diff
@@ -1,64 +1,30 @@
...
-        else:
-            tests = unittest.defaultTestLoader.loadTestsFromModule(module)
-            suite.addTests(tests)
-
-            logger.debug('found %d test cases in %s', tests.countTestCases(), mname)
-
-    logger.info('%d test cases were added', suite.countTestCases())
-
-    unittest.TextTestRunner().run(suite)
+logger.setLevel((os.getenv('REGRESS_LOG_LEVEL') or 'INFO').upper())
```
elicn, line 30: I guess this way the discovery happens automatically without the need of globing the files, however the deleted implementation provides some more info about the execution and failures to import tests. I am not sure whether we will still get the same details now.

tests/regress/arm_init_input_crash.py
```diff
@@ -32,6 +32,7 @@ def hook_code(uc, address, size, user_data):
 
 
 class TestInitInputCrash(regress.RegressTest):
+    @unittest.skipIf(sys.platform == 'win32' or platform.machine().lower() not in ('x86_64', 'arm64'), 'TO BE CHECKED!')
```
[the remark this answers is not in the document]
wtdcode: This is known failing on x86 targets and I need some time to investigate.

tests/regress/arm_memcpy_neon.py
```diff
@@ -1,52 +1,52 @@
...
-    '0d 07 0c f4'   # 0001F8CC        VST1.8      {D0}, [R12]!
-    '04 00 12 e3'   # 0001F8D0        TST         R2, #4
-    '04 30 91 14'   # 0001F8D4        LDRNE       R3, [R1],#4
-    '04 30 8c 14'   # 0001F8D8        STRNE       R3, [R12],#4
-    '82 2f b0 e1'   # 0001F8DC        MOVS        R2, R2,LSL#31
-    'b2 30 d1 20'   # 0001F8E0        LDRHCS      R3, [R1],#2
-    '00 10 d1 15'   # 0001F8E4        LDRBNE      R1, [R1]
-    'b2 30 cc 20'   # 0001F8E8        STRHCS      R3, [R12],#2
-    '00 10 cc 15'   # 0001F8EC        STRBNE      R1, [R12]
+SHELLCODE = (
```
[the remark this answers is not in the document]
wtdcode: This is for py2 i think.

PR #2084, opened 2025-01-06
Allow uc_ctl_set_page_size() for arm64

droe:
`uc_ctl_set_page_size()` for arm64 is currently disallowed with `UC_ERR_ARG` despite being implemented and functional.

Allow for arm64, and add test.

Additionally tested with a rather involved arm64 unicorn project that makes extensive use of paging in pages on demand, no issues.


wtdcode:
Thanks and welcome!

PR #2085, opened 2025-01-13
configure: update .h for clock_adjtime

clan:
man clock_adjtime: #include <sys/timex.h>
for glibc, if _GNU_SOURCE is defined, <time.h> will include <sys/timex.h> but not for musl, so replce <time.h> to <sys/timex.h>

PS it seems clock_adjtime is used nowhere?


wtdcode:
Thanks!

PR #2091, opened 2025-01-21
Fix UC_HOOK_MEM on arm32

xndcn:
See also #1354


wtdcode:
Sorry for late. Thanks!

PR #2095, opened 2025-01-25
arm64 python: Avoid nested class in insn hook for 10x hook and 10% overall speedup

droe:
Promote CpReg to a module-level class to address unnecessary performance reduction.

In a real-world use case tracing the emulation of real-world machine code, this change reduces time spent in CpReg namedtuple construction from 10% of overall time to below 1%, a ~ 10x speedup of the insn hook itself, or a ~ 10% overall speedup.  Measured using cProfile, python 3.13.

### DIFF
--- a/.github/workflows/build-uc2.yml
+++ b/.github/workflows/build-uc2.yml
@@ -479,7 +479,7 @@ jobs:
             artifact: 'ubuntu-cmake-aarch64.7z',
             archiver: '7z a',
             generators: 'Ninja',
-            distro: ubuntu20.04
+            distro: ubuntu22.04
           }
           - {
             os: ubuntu-latest,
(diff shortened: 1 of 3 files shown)


droe:
Build failures look unrelated to this change.

wtdcode:
No idea why it suddenly fails. Let me rerun it firstly.

droe:
Failures still look unrelated to this PR.  Similar failures also happen in other PR branches.  Also, it looks like GCC is crashing, and I'd be very surprised if this Python-only change would cause a GCC crash.

wtdcode:
The crashes result from Github Action upgrading default ubuntu distribution (kernel upgrades). It does not relate to your changes and I just lack time to fix it. Please wait a bit sorry.

wtdcode:
Thanks!

PR #2099, opened 2025-02-11
fix(arm): correct write to ARM coprocessor

amaanq:
### Problem

Any attempt to write to UC_ARM_REG_C1_C0_2 ([The ARM Coprocessor Access Control Register](https://developer.arm.com/documentation/ddi0344/k/system-control-coprocessor/system-control-coprocessor-registers/c1--coprocessor-access-control-register)) fails with UC_ERR_ARG.
The specific switch case handling writes to this register was commented out since 2021, but by default, the error code was initialized to UC_REG_OK, so there was no error returned until #1835, where this was changed to be initialized to UC_REG_ERR_ARG. As a result, any write to UC_ARM_REG_C1_C0_2 returned an error.

### Solution

In this PR, I've uncommented and properly implement the CPACR register write with appropriate size checking. I've also added a unit test to verify that writes no longer fail - before this PR, the test does indeed fail.

### DIFF
--- a/qemu/target/arm/unicorn_arm.c
+++ b/qemu/target/arm/unicorn_arm.c
@@ -440,10 +440,10 @@ uc_err reg_write(void *_env, int mode, unsigned int regid, const void *value,
             env->regs[15] = (*(uint32_t *)value & ~1);
             *setpc = 1;
             break;
-            // case UC_ARM_REG_C1_C0_2:
-            //     env->cp15.c1_coproc = *(int32_t *)value;
-            //     break;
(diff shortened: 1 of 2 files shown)


wtdcode:
Looks like commenting out by accident? Btw, you should target dev branch.

amaanq:
Done, thanks

wtdcode:
Will merge after I get CI fixed. Please wait a bit.

wtdcode:
Merged, thanks!

PR #2100, opened 2025-02-13
Python bindings: Refreshing workflow + regress tests

Antelox:
- Switched to the `ubuntu-24.04-arm` runner
- Bumped Windows runner to `windows-2022` and `Visual Studio 17 2022` GENERATORS
- Minor changes about checks in workflow jobs
- Fixed the pagefile job (even though should not be really needed)
- Refreshed the TO BE CHECKED regress tests to either update or remove the skip conditions
- Added a test to check if the created sdist archive is ok

### DIFF
--- a/.github/workflows/build-wheels-publish.yml
+++ b/.github/workflows/build-wheels-publish.yml
@@ -45,7 +45,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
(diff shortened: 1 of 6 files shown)


wtdcode:
LGTM, waiting for CI.

Thanks!

wtdcode:
Thanks, merged and triggering a full release CI.

PR #2106, opened 2025-02-15
GitHub Workflow: Updating jobs with new runners + fixes for Linux and windows x86 jobs

Antelox:
- Switched from windows-2019 to windows-2022
- Switched to ubuntu-24.04-arm image for Linux aarch64 jobs
- Fixed Linux x86 job. It was failing but not reported
- Switched from Visual Studio 16 2019 generators to Visual Studio 17 2022
- Un-commented Windows MINGW32 static and shared jobs
- Generic clean-up

### DIFF
--- a/.github/workflows/Crate-publishing.yml
+++ b/.github/workflows/Crate-publishing.yml
@@ -26,58 +26,58 @@ jobs:
       matrix:
         config:
           - {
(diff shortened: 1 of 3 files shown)


wtdcode:
Thanks, let's have 2.1.3.

PR #2107, opened 2025-02-15
GitHub Workflow: Fix aarch64 build bug introduced with previous refactoring #2106

Antelox:
- Remove a slash for aarch64 job that resulted to be a leftover from previous refactoring
- More cleanup


wtdcode:
Thanks!

Issue #2108, opened 2025-02-15
Unicorn 2.1.3


wtdcode:
There will be another complementary release 2.1.3 to address a few distributions issues.

wtdcode:
cc @Antelox

Antelox:
Hi @NickCao can you show me use case please?

Antelox:
Meanwhile https://github.com/unicorn-engine/unicorn/pull/2112

NickCao:
> Hi [@NickCao](https://github.com/NickCao) can you show me use case please?

The primary use case is for distribution packaging. Such as in https://github.com/NixOS/nixpkgs/blob/6a41cfea0cc3736b9e18099c61e564a9622d275d/pkgs/development/python-modules/geoarrow-pandas/default.nix#L30

wtdcode:
> > Hi [@NickCao](https://github.com/NickCao) can you show me use case please?
> 
> The primary use case is for distribution packaging. Such as in https://github.com/NixOS/nixpkgs/blob/6a41cfea0cc3736b9e18099c61e564a9622d275d/pkgs/development/python-modules/geoarrow-pandas/default.nix#L30

Does current master head `56ba3476e5f541f9f1b1943a916a86d86ece3092` works for nixpkgs?

NickCao:
> > > Hi [@NickCao](https://github.com/NickCao) can you show me use case please?
> > 
> > 
> > The primary use case is for distribution packaging. Such as in https://github.com/NixOS/nixpkgs/blob/6a41cfea0cc3736b9e18099c61e564a9622d275d/pkgs/development/python-modules/geoarrow-pandas/default.nix#L30
> 
> Does current master head `56ba3476e5f541f9f1b1943a916a86d86ece3092` works for nixpkgs?

It works, thanks!

PR #2112, opened 2025-02-18
Python bindings: Switch to setuptools-scm for versioning

Antelox:
- Switch from versioningit to setuptools-scm
- Disable building of tests via DUNICORN_BUILD_TESTS var

Discussion: https://github.com/unicorn-engine/unicorn/issues/2108#issuecomment-2664163420


wtdcode:
Thanks!

PR #2118, opened 2025-02-20
Fix: Potential Vulnerability in Cloned Function

tabudz:
**Description**
This PR fixes a security vulnerability in address_space_translate_for_iotlb() that was cloned from qemu but did not receive the security patch. The original issue was reported and fixed under https://github.com/qemu/qemu/commit/418ade7849ce7641c0f7333718caf5091a02fd4c.
This PR applies the same patch to eliminate the vulnerability.

**References**
https://nvd.nist.gov/vuln/detail/CVE-2022-35414
https://github.com/qemu/qemu/commit/418ade7849ce7641c0f7333718caf5091a02fd4c

### DIFF
--- a/qemu/exec.c
+++ b/qemu/exec.c
@@ -487,7 +487,7 @@ MemoryRegion *flatview_translate(struct uc_struct *uc, FlatView *fv, hwaddr addr
 
 /* Called from RCU critical section */
 MemoryRegionSection *
-address_space_translate_for_iotlb(CPUState *cpu, int asidx, hwaddr addr,
+address_space_translate_for_iotlb(CPUState *cpu, int asidx, hwaddr orig_addr,
                                   hwaddr *xlat, hwaddr *plen,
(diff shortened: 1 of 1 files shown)


wtdcode:
The backport is failing CI. Could you adapt it?

tabudz:
What do you think went wrong here? I applied the same fix as in the original commit.

wtdcode:
TARGET_PAGE_MASK is not constant in Unicorn and you need to adapt it accordingly.

tabudz:
But TARGET_PAGE_MASK appears multiple times in the same file, doesn't it?

PhilippTakacs:
Yes but TARGET_PAGE_MASK is defined to "uc->init_target_page->mask" and in this function uc is not defined. So to fix it you need `struct uc_struct *uc = cpu->uc;`. Have you even tried to compile this before you opened the PR?

PR #2121, opened 2025-02-26
add uc_mem_read_virtual

PhilippTakacs:
New api to read from a vaddr. When using the MMU it's useful to direct read from virtual addresses.

Needs some tests.

I'm not sure about a write_virtual() function. Could also be useful, but is a lot more complex to implement, because of snapshots.

A virtual to physical translate function would be also nice, but currently the `tlb_fill()` function has no probe argument.


wtdcode:
I have dreamed of this API for quite a while because `uc_mem_read/write` only deals with physical addresses and super inconvenient with MMU. Thanks for prototyping!

It would be cool to even have translation API as suggested, but many QEMU MMU fill implementation has serious side effects. 

Why does "write_vritual" not work with snapshots? What's the current `uc_mem_write` behavior? Maybe just keep the same behavior.

Lastly please target dev branch instead, the master branch is preparing for 2.1.3 and this change can't go into 2.1.3 obviously =).

PhilippTakacs:
> Lastly please target dev

Sorry my bad.

> Why does "write_vritual" not work with snapshots?

It would work, it's just more code and maybe duplicated code, because of the snapshot handling. With a simple virtual to physical API it would be more simple.

> It would be cool to even have translation API as suggested, but many QEMU MMU fill implementation has serious side effects.

In the current QEMU version is `tlb_fill()` replaced by `tlb_fill_align()`. This function has a probe parameter, which would be the key for a virtual to physical translation API.

wtdcode:
CI is failing at this moment and I lack time to fix it. Will do later this week.

PhilippTakacs:
> CI is failing at this moment

There were also some bugs in my changes which causes the CI to fail.

> but many QEMU MMU fill implementation has serious side effects

I have added a testcase (on x86) and read the MMU implementations to verify this would work. As far as I see it works on every MMU expect for SPARC. Would it be OK to just document this issue?

About the write: Would it be OK to just have a read implementation at the moment?

PhilippTakacs:
> What's the current uc_mem_write behavior? Maybe just keep the same behavior.

Currently uc_mem_write does COW when memory snapshots are enabled. I would also expect this from a `uc_mem_write_virtual` function.

The problem is: This is quite complicated code and with the given `tlb_vaddr_to_host` it would be more complicated to implement for a `uc_mem_write_virtual` function. Given the fact that I have two times missed some cases in the `uc_mem_write` function, I would try to avoid implement this without a simple way to translate virtual to physical addresses.

wtdcode:
I just revisited the `tlb_vaddr_to_host` and noticed the `probe` parameter, which should avoid the side effects. Sorry I forget the functionality here yesterday without further checking the code. I also saw documents in SPARC and that's okay we can document it and return something like `UC_ERR_ARG`.

Btw the `tlb_fill` function indeed has the `probe` argument, I assume we can expose the translation API directly? Maybe just expose `tlb_vaddr_to_host`? In this case, we can avoid code deduplication by using `tlb_vaddr_to_host` to convert vaddr to phyaddr and use `uc_mem_read/write` to do memory operations.

```c
typedef struct CPUClass {
    // ...
    bool (*tlb_fill)(CPUState *cpu, vaddr address, int size,
                     MMUAccessType access_type, int mmu_idx,
                     bool probe, uintptr_t retaddr);
    // ...
}
```

PhilippTakacs:
So I have added a `uc_virtual_to_physical` function. The problem I had was: I wanted to (ab)use a function and only add some code to extract the physical address (like I have done with `tlb_vaddr_to_host`). I have now just added a `tlb_vaddr_to_paddr`.

I'll try to implement a `uc_mem_write_virtual` next week.

PhilippTakacs:
The `uc_mem_write_virtual` function should work, but I haven't had time to test it. I'll write some tests and a bit more documentation. But I have not that much time at the moment.

PhilippTakacs:
@wtdcode write_virtual is also there and tested.

wtdcode:
> @wtdcode write_virtual is also there and tested.

Thanks for the brilliant work. I can review only a bit later as I'm having an urgent deadline on April 10.

uc.c
```diff
@@ -773,6 +773,136 @@ static bool check_mem_area(uc_engine *uc, uint64_t address, size_t size)
     return (count == size);
 }
 
+uc_err uc_virtual_to_physical(uc_engine *uc, uint64_t address, uc_prot prot,
```
wtdcode, line 776: Do we really need prot here? Is there possible that different protections lead to different physical addresses? If the protection is only used for access check, is it trivial to skip such checks and only "translate" addresses for whatever protection? If that is not trivial, we can keep it as it is.

qemu/unicorn_common.h
```diff
@@ -28,6 +29,68 @@ static inline bool cpu_physical_mem_write(AddressSpace *as, hwaddr addr,
...
+{
+    MMUAccessType access_type;
+    void *hostptr;
+    int mmu_idx = cpu_mmu_index(uc->cpu->env_ptr, false);
+
+    /*
+     * Only page aligned access is allowed,
+     * because tlb_fill() might change the mappings
+     */
+    assert((addr & TARGET_PAGE_MASK) == ((addr + len - 1) & TARGET_PAGE_MASK));
```
wtdcode, line 42: this shall return `UC_ERR_ARG` or more appropriate errors instead of assertions.

include/unicorn/unicorn.h
```diff
@@ -933,6 +933,46 @@ uc_err uc_mem_write(uc_engine *uc, uint64_t address, const void *bytes,
 UNICORN_EXPORT
 uc_err uc_mem_read(uc_engine *uc, uint64_t address, void *bytes, size_t size);
 
+/*
+ Read a range of bytes in memory after mmu translation.
+
+ @uc:      handle returned by uc_open()
+ @address: starting virtual memory address of bytes to get.
+ @prot:    The access type for the tlb lookup
```
wtdcode, line 941: As reviewed previously, if `@prot` is essential, please document a bit more here. Like how different protections could affect the behavior. For example, what does `uc_mem_read_virtual` with `prot = UC_PROT_WRITE` imply?

uc.c
PhilippTakacs: The prot is to have the mmu access rights checked. You might want to check if the corresponding address is write accessible. To check this you can call `uc_virtual_to_physical(uc, addr, UC_PROT_WRITE, &paddr)` and check the return code.

Also it's necessary to have prot for a potential `tlb_fill` call.

uc.c
```diff
@@ -773,6 +773,136 @@ static bool check_mem_area(uc_engine *uc, uint64_t address, size_t size)
...
+            return UC_ERR_FETCH_PROT;
+        }
+    }
+
+    restore_jit_state(uc);
+    return UC_ERR_OK;
+}
+
+UNICORN_EXPORT
+uc_err uc_mem_read_virtual(uc_engine *uc, uint64_t address, uc_prot prot,
```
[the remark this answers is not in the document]
PhilippTakacs: Sounds good, I'll adopt this.

uc.c
wtdcode: But for “translation” itself, it has nothing to do with permissions, right? My point is that, from the semantics of the API, protection seems not necessary here though I understand tlb_fill requires this.

wtdcode: For example, assume a memory region located at vaddr 0x2000 with paddr 0x1000 but with PROT_NONE, i.e. no one can access it. But from the perspective of emulator, the mapping is there exactly and we should be able to translate addresses no matter the protections.

However, I do understand that qemu code might not design for such cases. Therefore, if you think it is not feasible to translate addresses without protection, we can leave it as it is but with better docs.

qemu/unicorn_common.h
PhilippTakacs: I use asserts for potential unicorn bugs and return errors for user bugs. This internal function is designed to only be called within one page. This is implemented in `uc_mem_read_virtual`. Returning an error might cause a user to believe it's a bug in there code.

I'll write a comment about this.

wtdcode: Yeah, but addr is from user inputs, no? `uc_mem_read_virtual` seems using it directly? My concern is that this might cause the assertion too easy to trigger.

PhilippTakacs: Yes addr is (more or less) from user input, but len is enforced to be maximum the rest of the page. What this assert checks is that the full len is on the same page. So the assert is in exactly what `uc_mem_read_virtual` enforces. When we add other callers, they must handle this internal api the same way.

So I don't see how this assert triggers to early.

PhilippTakacs:
@wtdcode is there anything missing for this getting merged?

wtdcode:
> @wtdcode is there anything missing for this getting merged?

Hello Philipp, sorry for late because I'm on academic travel. I will give another review shortly.

wtdcode:
Thanks for your consistent contributions =3.

PR #2122, opened 2025-03-03
Fix Python 3.8 to 3.11

Arusekk:
Current logic requires old setuptools on py3.8 up to py3.11. So if new setuptools is installed, importing unicorn always fails with ImportError on these Python versions.

It now tries to use old setuptools on py3.8, but falls back gracefully to not using anything.
So it works on py3.9+ fully, and on py3.8 it also works but with more limited path discovery (I think it is enough, since no one complained yet).

Fixes: [0c34496a395eac567dc24eb6574550938a52b126](https://github.com/unicorn-engine/unicorn/pull/2016/commits/0c34496a395eac567dc24eb6574550938a52b126) ("Modify canonicals import")
Ref: https://github.com/Gallopsled/pwntools/pull/2556

### DIFF
--- a/bindings/python/unicorn/unicorn_py3/unicorn.py
+++ b/bindings/python/unicorn/unicorn_py3/unicorn.py
@@ -114,14 +114,17 @@ def __load_uc_lib() -> ctypes.CDLL:
     # - global load
     # - python's lib directory
 
-    if sys.version_info.minor >= 12:
+    if sys.version_info >= (3, 9):
         from importlib import resources
 
         canonicals = resources.files('unicorn') / 'lib'
     else:
-        import pkg_resources
+        try:
+            import pkg_resources
 
-        canonicals = pkg_resources.resource_filename('unicorn', 'lib')
+            canonicals = pkg_resources.resource_filename('unicorn', 'lib')
+        except ImportError:
+            canonicals = None
 
     lib_locations = [
         os.getenv('LIBUNICORN_PATH'),


wtdcode:
Thanks for your patch but this actually surprises me as our wheels are tested for all these versions. How to reproduce it? So that we can add this to our CI too. Is it related to outdated (or newer?) setuptools? Is it possible that we should have more precise dependency on setuptools?

FYI, our CI testing wheels on all python versions: https://github.com/unicorn-engine/unicorn/actions/runs/13309809343

cc @Antelox

Arusekk:
What I did:
```sh
# set up UV with virtualenv
uv pip install unicorn  # selects the wheel but does not install setuptools at all
python -c 'import unicorn'
```

UV is an alternative implementation of pip, which does not install even pip into the venv.

The reason this worked in CI is because precisely in python3.12 the stdlib `venv` module [stopped installing setuptools][cpy]. This was probably the cause 3.12 was chosen by the PR author as the split point instead of 3.9.
So it should not normally happen, in virtual environments or otherwise; this bug is low-impact. But this means that setuptools (with deprecated `pkg_resources` module) is currently a de-facto requirement for unicorn with py3.11 and below, so this bug will surface on systems where unicorn is a distro package. This PR lifts that restriction.

[cpy]: https://github.com/python/cpython/issues/95299

Arusekk:
I just saw that Unicorn CI uses `virtualenv` rather than `venv`, but the same applies: https://github.com/pypa/virtualenv/pull/2558

wtdcode:
> What I did:
> 
> ```shell
> # set up UV with virtualenv
> uv pip install unicorn  # selects the wheel but does not install setuptools at all
> python -c 'import unicorn'
> ```
> 
> UV is an alternative implementation of pip, which does not install even pip into the venv.
> 
> The reason this worked in CI is because precisely in python3.12 the stdlib `venv` module [stopped installing setuptools](https://github.com/python/cpython/issues/95299). This was probably the cause 3.12 was chosen by the PR author as the split point instead of 3.9. So it should not normally happen, in virtual environments or otherwise; this bug is low-impact. But this means that setuptools (with deprecated `pkg_resources` module) is currently a de-facto requirement for unicorn with py3.11 and below, so this bug will surface on systems where unicorn is a distro package. This PR lifts that restriction.

Thanks for good explanation. I think our excellent contributor @Antelox is looking into it.

wtdcode:
Before that I need to fix CI firstly, btw.

Antelox:
Hi @Arusekk please take a look at the PR I created right now. I' sort of took over from this one to resolve an old issue as well. Thanks for this!

PR #2125, opened 2025-03-06
Snapshot use after free

PhilippTakacs:
Fixes use after free on snapshots restore.

Also some cleanup in the unit tests. Mostly forgot to call `uc_close`.

The sanitizer also finds some stack overflow in `test_x86_smc_add` and `test_x86_smc_mem_hook`. I'm currently don't understand the error.

In `test_context_snapshot` of `test_mem` there is an Memory leak which I'm currently debug.

### DIFF
--- a/qemu/softmmu/memory.c
+++ b/qemu/softmmu/memory.c
@@ -28,6 +28,7 @@
 
 void memory_region_transaction_begin(void);
 static void memory_region_transaction_commit(MemoryRegion *mr);
+static void memory_region_destructor_container(MemoryRegion *mr);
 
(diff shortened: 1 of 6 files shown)


PhilippTakacs:
> In `test_context_snapshot` of `test_mem` there is an Memory leak which I'm currently debug.

This is also fixed now.

wtdcode:
`The sanitizer also finds some stack overflow in test_x86_smc_add and test_x86_smc_mem_hook. I'm currently don't understand the error.`

Noticed this and will fix too.

PR #2139, opened 2025-03-30
Fix physical address truncation on 32-bit systems with addressing extensions

ExhoAR22:
While experimenting with Unicorn to emulate an ARMv7 system that heavily utilizes LPAE, I noticed Unicorn would translate the virtual addresses properly. However, the address got truncated to 32-bits somewhere in the pipeline, which rendered my emulator inoperable.

I tracked down the problem to a couple of types in the `CPUTLBEntry` struct and surrounding functions. I've changed them from `target_ulong` to `hwaddr` (hopefully I didn't miss any).

This change seems to work for my setup, I haven't tried other architectures (such as x86 with PAE) but if my understanding of the code is correct, the change should be architecture-angostic.

When running `cargo test`, all tests seem to pass. However, I have not added new ones for this scenario.

### DIFF
--- a/qemu/accel/tcg/cputlb.c
+++ b/qemu/accel/tcg/cputlb.c
@@ -1469,7 +1469,7 @@ load_helper(CPUArchState *env, target_ulong addr, TCGMemOpIdx oi,
     uintptr_t index = tlb_index(env, mmu_idx, addr);
     CPUTLBEntry *entry = tlb_entry(env, mmu_idx, addr);
     target_ulong tlb_addr = code_read ? entry->addr_code : entry->addr_read;
-    target_ulong paddr;
+    hwaddr paddr;
     const size_t tlb_off = code_read ?
         offsetof(CPUTLBEntry, addr_code) : offsetof(CPUTLBEntry, addr_read);
     const MMUAccessType access_type =
@@ -2090,7 +2090,7 @@ store_helper(CPUArchState *env, target_ulong addr, uint64_t val,
     uintptr_t index = tlb_index(env, mmu_idx, addr);
     CPUTLBEntry *entry = tlb_entry(env, mmu_idx, addr);
     target_ulong tlb_addr = tlb_addr_write(entry);
-    target_ulong paddr;
+    hwaddr paddr;
     const size_t tlb_off = offsetof(CPUTLBEntry, addr_write);
     unsigned a_bits = get_alignment_bits(get_memop(oi));
     void *haddr;
--- a/qemu/include/exec/cpu-defs.h
+++ b/qemu/include/exec/cpu-defs.h
@@ -112,7 +112,7 @@ typedef struct CPUTLBEntry {
             target_ulong addr_read;
             target_ulong addr_write;
             target_ulong addr_code;
-            target_ulong paddr;
+            hwaddr paddr;
             /* Addend to virtual address to get host address.  IO accesses
                use the corresponding iotlb value.  */
             uintptr_t addend;
--- a/qemu/include/exec/exec-all.h
+++ b/qemu/include/exec/exec-all.h
@@ -477,7 +477,7 @@ address_space_translate_for_iotlb(CPUState *cpu, int asidx, hwaddr addr,
 hwaddr memory_region_section_get_iotlb(CPUState *cpu,
                                        MemoryRegionSection *section);
 
-static inline bool uc_mem_hook_installed(struct uc_struct *uc, target_ulong paddr)
+static inline bool uc_mem_hook_installed(struct uc_struct *uc, hwaddr paddr)
 {
     if (HOOK_EXISTS_BOUNDED(uc, UC_HOOK_MEM_FETCH_UNMAPPED, paddr))
         return true;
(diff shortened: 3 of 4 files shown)


wtdcode:
Looks good to me. Let's see if this broken any unit test in C.

wtdcode:
Could you also add a small test to here: https://github.com/unicorn-engine/unicorn/blob/master/tests/unit/test_arm.c ?

ExhoAR22:
> Could you also add a small test to here: https://github.com/unicorn-engine/unicorn/blob/master/tests/unit/test_arm.c ?

Doing that would mean writing code that initializes the MMU and creates a page table.

Since there aren't any existing tests that utilize the MMU for ARM that would mean writing them from scratch, and while I don't mind doing that, I won't have the time to do it until the weekend.

Does this small fix really necessitate a test?

PhilippTakacs:
As far as I see this can also be tested with `UC_TLB_VIRTUAL`. With this testing this is a lot simpler.

wtdcode:
> > Could you also add a small test to here: https://github.com/unicorn-engine/unicorn/blob/master/tests/unit/test_arm.c ?
> 
> Doing that would mean writing code that initializes the MMU and creates a page table.
> 
> Since there aren't any existing tests that utilize the MMU for ARM that would mean writing them from scratch, and while I don't mind doing that, I won't have the time to do it until the weekend.

It’s okay and takes your time.

> 
> Does this small fix really necessitate a test?

The bug was introduced in #1746 and Unicorn is too complex so that even I could make mistakes and ignore even such obvious errors. Therefore, writing a test can ensure your use case is guaranteed to work in our future releases, both beneficial to your project and Unciorn engine in a long run.

That said, I sincerely appreciate if you could write a test case or if you are really too busy, we can go as it is, but at a risk of breaking the code in the future. Take your time and feel free to make a choice, I’m not pushing you anyway.

wtdcode:
Considering you are using rust, it's also okay to add to https://github.com/unicorn-engine/unicorn/blob/master/tests/rust-tests/main.rs if you prefer to add a test. Our CI will run these tests for every commit.

PhilippTakacs:
Here is a test for this using `UC_TLB_VIRTUAL`:

```
diff --git a/tests/unit/test_arm.c b/tests/unit/test_arm.c
index 538a91c0..dda6f2be 100644
--- a/tests/unit/test_arm.c
+++ b/tests/unit/test_arm.c
@@ -956,6 +956,42 @@ static void test_arm_cp15_c1_c0_2(void)
     OK(uc_close(uc));
 }
 
+static bool test_arm_v7_lpae_hook_tlb(uc_engine *uc, uint64_t addr,
+                                    uc_mem_type type, uc_tlb_entry *result,
+                                    void *user_data)
+{
+    result->paddr = addr + 0x100000000;
+    result->perms = UC_PROT_ALL;
+    return 1;
+}
+
+static void test_arm_v7_lpae_hook_fetch(uc_engine *uc, uc_mem_type type,
+                                 uint64_t address, int size, uint64_t value,
+                                 void *user_data)
+{
+    TEST_CHECK(address == 0x100000000);
+}
+
+static void test_arm_v7_lpae(void)
+{
+    uc_engine *uc;
+    uc_hook hook_fetch, hook_tlb;
+    char code[] = "\x00\xf0\x20\xe3"; // nop
+    OK(uc_open(UC_ARCH_ARM, UC_MODE_ARM, &uc));
+    OK(uc_ctl_set_cpu_model(uc, UC_CPU_ARM_CORTEX_A7));
+
+
+    OK(uc_ctl_tlb_mode(uc, UC_TLB_VIRTUAL));
+    OK(uc_hook_add(uc, &hook_tlb, UC_HOOK_TLB_FILL, test_arm_v7_lpae_hook_tlb, NULL, 1, 0));
+    OK(uc_hook_add(uc, &hook_fetch, UC_HOOK_MEM_FETCH, test_arm_v7_lpae_hook_fetch, NULL, 1, 0));
+
+    OK(uc_mem_map(uc, 0x100001000, 0x1000, UC_PROT_ALL));
+    OK(uc_mem_write(uc, 0x100001000, code, sizeof(code)));
+    OK(uc_emu_start(uc, 0x1000, 0x1000 + sizeof(code) - 1, 0, 0));
+
+    OK(uc_close(uc));
+}
+
 TEST_LIST = {{"test_arm_nop", test_arm_nop},
              {"test_arm_thumb_sub", test_arm_thumb_sub},
              {"test_armeb_sub", test_armeb_sub},
@@ -985,4 +1021,5 @@ TEST_LIST = {{"test_arm_nop", test_arm_nop},
              {"test_arm_tcg_opcode_cmp", test_arm_tcg_opcode_cmp},
              {"test_arm_thumb_tcg_opcode_cmn", test_arm_thumb_tcg_opcode_cmn},
              {"test_arm_cp15_c1_c0_2", test_arm_cp15_c1_c0_2},
+             {"test_arm_v7_lpae", test_arm_v7_lpae},
              {NULL, NULL}};
```

ExhoAR22:
Thanks @PhilippTakacs, I added your patch!
Wasn't familiar with this feature, it'll probably be quite useful to me in the future.

wtdcode:
Thanks for both of you =).

PR #2140, opened 2025-04-01
refactor(lib): mark pointers as const where possible

amaanq:
This PR just marks some parameters that are pointers as const pointers - I noticed that clang-tidy was warning me that these could be marked as const, and it makes sense to me.


wtdcode:
Make sense to me. Sorry I missed this PR.

PR #2142, opened 2025-04-03
bindings: ruby: fix version identifier to 2.1.3

anthraxx:
It seems when bumping 2.1.2 and 2.1.3 some places have been forgotten. This actually results in a 2.1.1 gem file when building distro packages as a result.


wtdcode:
Also updated here: https://github.com/unicorn-engine/unicorn/wiki/Developers so that I won't forget again =(

PR #2155, opened 2025-04-12
feat(arm): add an `ESR` register

### LINKED ISSUE #2153: [ARM32] Hook for "svc" always returns interrupt number 2

amaanq:
- Closes #2153

### Problem

When emulating ARM code, one cannot extract the syndrome of a given interrupt (for calls such as `SVC`, `HVC`, etc.). ARMv7 doesn't have an ESR (exception syndrome register) like AArch64 does, but it should be possible for consumers to get this information.

### Solution

I've added a `UC_ARM_REG_ESR` similar to the AArch64 registers. This register can read from and write to the `syndrome` at any given point. A test has been added to verify that reading from this syndrome value works.


include/unicorn/arm.h
```diff
@@ -211,6 +211,7 @@ typedef enum uc_arm_reg {
     UC_ARM_REG_XPSR_NZCVQ,
     UC_ARM_REG_XPSR_G,
     UC_ARM_REG_XPSR_NZCVQG,
+    UC_ARM_REG_ESR,
```
wtdcode, line 214: We shall always append enums or the values of some enums like `UC_ARM_REG_CP_REG` will be changed.

Also please add a small comment here to indicate it is a pseudo register.

amaanq: done

wtdcode:
LGTM. Just wait for CI.

Thanks =)

PR #1918, opened 2024-01-09
Add Renesas RH850 architecture support

virtualabs:
Hello,

This PR adds support for the Renesas RH850 architecture and has been tested against different firmware files designed for the Renesas F1KM-S4 development board.

This implementation supports all instructions for the V850e3v2 CPU used in the Renesas RH850 architecture.

PR #2156, opened 2025-04-12
Update RH850 code in staging

amaanq:
This PR updates the RH850 code in the staging branch - it's effectively my way of saying "I've reviewed the code", but there were a lot of changes to be made, so I decided make those changes, since https://github.com/unicorn-engine/unicorn/pull/1918 is a bit old and I didn't want to burden the author with dozens of comments :^).

The main highlights are:

- Cleaned up main RH850 header
- Removing unnecessary code that isn't used by unicorn
- Ensured that code hooks work (only block hook support was present)
- Ensured usage of the instance-specific `tcg_ctx` for all register manipulation, rather than static variables
- Adjusted naming convention where it was inconsistent
- Cleaned up the test

@virtualabs if you have the time, it'd be great if you can take a look at these changes. If they look good to you, I'd consider RH850 ready. One question I had though is, why name it RH850 and not V850? From the little searching I did online, the RH850 cores are the same as the V850e3v5, so it'd make more sense to name this "V850" as that's the architecture, and then RH850/V850e3v5 is one cpu "model" that can be implemented (then, in theory, we could also support emulating v850e1/e2 chips as well).


virtualabs:
I've had a quick look at the changes and it seems OK (but I'm not a unicorn engine expert). Maybe I should try this version with some piece of code I have to see if it still works as expected, but for now I don't have access to this code and won't be able to test it before a few days.

Regarding the naming, yes RH850 uses a V850e3v5 CPU and renaming it to V850 seems the best option (but as a consequence, every reference to rh850 in the code need to be renamed as well, how tedious). Originally, I've worked on this implementation in the context of emulating a RH850 SoC thus explaining the name (my mistake).

wtdcode:
Renaming usually is the easiest part doing the port =)

PR #2159, opened 2025-04-12
fix(rust): watch all source and header files

amaanq:
### Problem

If a C source file is edited, and then subsequently a Rust build/test run is trigger, the code is not recompiled. This is because rustc doesn't know to rerun the build script when these C files change.

### Solution

In this PR, I've added support for emitting the rust instruction to rerun the build script when any C files change, instead of having to have to trigger it manually.


wtdcode:
Thanks!

PR #2160, opened 2025-04-12
feat(rust): improve ARM CP register ergonomics

amaanq:
This is a follow-up to #2138

### Problem

The Rust structs for ARM/ARM64 CP registers were named `RegisterARM(64)_CP` - this isn't conventional. Ideally, it should be `RegisterARM64CP`. Additionally, reading from the CP register was broken on ARM64, and neither ARM nor ARM64 had a write variant.

### Solution

I've added a step in the `build.rs` to remove underscores in the register struct names sans the prefix/suffix, and have added write variants of the `coproc` functions. They also now take in a register as the argument, so users can manipulate which coprocessor register to read from. Lastly, I added some helper methods on these CP register structs to make it more ergonomic to construct them.


wtdcode:
Thanks!

PR #2162, opened 2025-04-13
fix(ppc): avoid using static TCG variables

amaanq:
### Problem

A lot of variables in `qemu/target/ppc/translate.c` use global TCG variables, which is unexpected as these variables should be a part of the instance-specific TCG context.

### Solution

In this PR, I've moved these global PPC TCG variables to the TCGContext struct.


wtdcode:
CI fails =)

amaanq:
The android CI failure is pretty weird..

PR #1903, opened 2023-11-01
Add Backend Support for LoongArch64

tyssjhx:
Add backend support for LoongArch64, fixing the conflicts with tcg function tcg_out_dupi_vec. The codes are tested by the unit tests on a LoongArch and a x86 machine.

PR #2164, opened 2025-04-14
Loongarch port

wtdcode:
This builds and passes all tests on 3A5000.

Supersedes #1903


wtdcode:
Note: both gcc and clang pass all tests =)

PR #2167, opened 2025-04-15
S390x registers

### LINKED ISSUE #2166: [S390X] Missing support for floating-point registers
While testing a [pwndbg PR](https://github.com/pwndbg/pwndbg/pull/2873), it appears that s390x [doesn't support floating-point registers](https://github.com/unicorn-engine/unicorn/blob/f8c6db950420d2498700245269d0b647697c5666/qemu/target/s390x/unicorn.c#L57-L85) despite there being constants for them defined. (I'm using `unicorn            2.1.3`, not that I think it matters)

```python
#!/usr/bin/env python3
from unicorn import *
from unicorn.s390x_const import *

print(unicorn.uc_version())

emu = Uc(UC_ARCH_S390X, UC_MODE_BIG_ENDIAN)

emu.reg_write(UC_S390X_REG_F0, 0x22)
assert emu.reg_read(UC_S390X_REG_F0) == 0x22

```

This fails with:
```
(2, 1, 33620991)
WARNING: Your register accessing on id 17 is deprecated and will get UC_ERR_ARG in the future release (2.2.0) because the accessing is either no-op or not defined. If you believe the register should be implemented or there is a bug, please submit an issue to https://github.com/unicorn-engine/unicorn. Set UC_IGNORE_REG_BREAK=1 to ignore this warning.
WARNING: Your register accessing on id 17 is deprecated and will get UC_ERR_ARG in the future release (2.2.0) because the accessing is either no-op or not defined. If you believe the register should be implemented or there is a bug, please submit an issue to https://github.com/unicorn-engine/unicorn. Set UC_IGNORE_REG_BREAK=1 to ignore this warning.
  File "/tmp/unicorn-testing/test.py", line 10, in <module>
    assert emu.reg_read(UC_S390X_REG_F0) == 0x22
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

wtdcode:
close #2166


wtdcode:
Rust is supposed to be `bindgen` now?

PR #2169, opened 2025-04-21
Add from_handle_with_data method

Evian-Zhang:
This makes it easier to do some FFI stuff. And it is consistent with `new` and `new_with_data`.


Evian-Zhang:
BTW, the original rust code does not seem to be formatted yet. 🤔 Maybe someone could just `cargo fmt` it via a commit without a PR.

wtdcode:
Thanks!

PR #2170, opened 2025-04-22
Make Unicorn struct clone-able

Evian-Zhang:
In some situations, we hold a mutable reference to the user-defined data stored in `Unicorn` struct by `get_data_mut`, and we also need to call some mutable method like `reg_write` at the same time. However, currently Rust does not support inter-procedure field-granularity borrow checker (see [this blog](https://smallcultfollowing.com/babysteps/blog/2025/02/25/view-types-redux/) for more details), as a result, this makes it impossible to do this.

Since the `Unicorn` struct holds an `Rc` of `UnicornInner`, we can just cheaply clone the `Unicorn` to get rid of this.

In this PR, I use `perfect_derive` instead of simple `derive`, since Rust's derive will require the generic parameter `D` also implement `Clone`, which is unnecessary (See [this blog](https://smallcultfollowing.com/babysteps/blog/2022/04/12/implied-bounds-and-perfect-derive/) for more details).

### DIFF
--- a/bindings/rust/Cargo.toml
+++ b/bindings/rust/Cargo.toml
@@ -16,6 +16,7 @@ workspace = true
 
 [dependencies]
 unicorn-engine-sys = { version = "2.1.3", path = "sys", features = [], default-features = false }
+perfect-derive = "0.1.5"
 
 [build-dependencies]
 bindgen = "0.71.1"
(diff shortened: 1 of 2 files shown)


wtdcode:
Is it trivial doable like:

```rust
impl<D: Clone> Clone for Unicorn {
...
}
```

Evian-Zhang:
Yes you are right. Removed the `perfect_derive` and implemented it manually.

wtdcode:
Thanks =)

PR #2176, opened 2025-04-29
Add edge generated callback

Evian-Zhang:
I've currently completed a rough version of unicornafl to add ability to cache TBs across forking (that code change is relatively large and thus requires further cleanup in a few days). This change is necessary for that feature.


wtdcode:
Cool!

PR #2180, opened 2025-05-06
Added CFG Check and standalone flag for .NET binding

AdvDebug:
Added code for CFG check (windows) for the .NET binding since unicorn doesn't support CFG. and added the standalone flag to make it easier to build and work with the library.


AdvDebug:
> You shall target dev branch =)

Oh right didn't notice 😅 delete this and i will make another one

PR #2182, opened 2025-05-07
Added CFG check and standalone flag for .NET binding

AdvDebug:
Added CFG check on instance creation since unicorn doesn't support it and added "--standalone" flag to make it easier to work with the library.


wtdcode:
Thanks & welcome!

PR #2185, opened 2025-05-10
Shall we not check stop_request in hook loops?

Evian-Zhang:
`uc->stop_request` is set by `uc_emu_stop()`. Currently, there is a check for this flag in hook loops: if the flag is true after a hook, then the loop is break out. This behavior is undocumented, and to be more specific, the behavior is:

If user called `uc_emu_stop()` in a hook callback, then:

* The following callbacks with the same type at the same moment will be skipped
* For instructions executed later until the TB end, all types of hook will only execute the first callback.

This behavior is strange, and makes users lose control of instructions executed before TB ends.

Since users have called emu stop, and a TB will not be so large, not checking the `stop_request` may not involve much overhead?

### DIFF
--- a/qemu/accel/tcg/cputlb.c
+++ b/qemu/accel/tcg/cputlb.c
@@ -1529,10 +1529,6 @@ load_helper(CPUArchState *env, target_ulong addr, TCGMemOpIdx oi,
                                            ((uc_cb_eventmem_t)hook->callback)(uc, UC_MEM_FETCH_UNMAPPED, paddr, size, 0, hook->user_data));
                     if (handled)
                         break;
-
-                    // the last callback may already asked to stop emulation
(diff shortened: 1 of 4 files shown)


Evian-Zhang:
I'm totally confused. All I changed is the true branch for `uc->stop_request`, which is only set by `uc_emu_stop()`. The failed CI is `test_arm_thumb_ite`, which does not involve any `uc_emu_stop()` call. Why it failed🤔🤔🤔

wtdcode:
> For instructions executed later until the TB end, all types of hook will only execute the first callback.

This is not correct because uc will exit emulation immediately by `check_exit_request`. I assume the CI failure has told you about this.

Evian-Zhang:
OK I understand. Thanks :)

PR #2197, opened 2025-06-08
fix incorrect ret of trampoline `closure` in `alloc_code_gen_buffer`

HyperSine:
In x86 windows (NOT x64), the second argument of `AddVectoredExceptionHandler` is a function pointer that uses stdcall calling conv:

```c
// from winnt.h
#define NTAPI __stdcall
...
...
typedef LONG (NTAPI *PVECTORED_EXCEPTION_HANDLER)(
    struct _EXCEPTION_POINTERS *ExceptionInfo
    );
```

The trampoline `closure`, which is passed to `AddVectoredExceptionHandler`, is responsible to clean up args on stack. So the last instruction of `closure` should be `ret 4` instead of just `ret`, otherwise it would cause stack unbalance and you may get crash randomly.


wtdcode:
Thanks

PR #2199, opened 2025-06-08
Minor fixes for vmem apis

wtdcode:
cc @PhilippTakacs


uc.c
```diff
@@ -803,6 +803,8 @@ uc_err uc_vmem_translate(uc_engine *uc, uint64_t address, uc_prot prot,
             return UC_ERR_WRITE_PROT;
         case UC_PROT_EXEC:
             return UC_ERR_FETCH_PROT;
+        default:
+            return UC_ERR_ARG;
```
PhilippTakacs, line 807: This is not necessary, because the code checks `prot` for exact one protocol. But it doesn't harm.

PR #2217, opened 2025-07-17
Update to zig 0.14.0

RoadRoller01:
I have just edited build.zig.zon,
and added .zig-cache to .gitignore

Discussion:
if you have unicorn installed system wide and tried to build the project, it will
panic about not finding the unicorn/build dir.

if you haven't unicorn installed system wide it will panic about not finding the unicorn library instead of building it

its seems that you could force it to build by running `zig build cmake`
but it will error with 
```shell
error: unsupported linker arg: --dependency-file=CMakeFiles/unicorn.dir/link.d
```

I prefer dropping down cmake and use the zig build system it seems that someone have ported unicorn to it. [here](https://github.com/vesim987/unizorn/blob/master/build.zig)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -88,6 +88,7 @@ rust_build
 [Oo]bj/
 packages/
 cmocka/
+.zig-cache/
 zig-cache/
 zig-out/
 .cache
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -1,5 +1,11 @@
 .{
-    .name = "unicorn",
+    .name = .unicorn,
     .version = "2.1.3",
-    .paths = .{""},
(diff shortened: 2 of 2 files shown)


wtdcode:
> I prefer dropping down cmake and use the zig build system it seems that someone have ported unicorn to it. [here](https://github.com/vesim987/unizorn/blob/master/build.zig)

I agree with this and contributions are highly welcome.

vesim987:
I can upstream my build.zig if you want, iirc building for Linux and MacOS was working, windows build failed due to missing some compiler built in

wtdcode:
> I can upstream my build.zig if you want, iirc building for Linux and MacOS was working, windows build failed due to missing some compiler built in

We accept the improvements for bindings as long as CI is green in most case.

PR #2219, opened 2025-07-22
glib_compat/glib_compact: Clear the buffer in g_hash_table_resize

MarsDoge:
There has a hidden danger in g_hash_table_resize function:

The g_hash_table_resize, it seems to resize the has table, and write the old value to the new hash table. It seems look for the empty items in the new_nodes, and if the new_nodes is not be cleared, it maybe naver found the empty items. So I added a line under the g_new0 function, and I think clearing new_nodes every time may improve the performance.

Added a memset to clear the new_nodes to fix this problem.


MarsDoge:
@wtdcode please review it, thanks !

PR #2220, opened 2025-07-28
Revert "glib_compat/glib_compact: Clear the buffer in g_hash_table_resize"

wtdcode:
Reverts unicorn-engine/unicorn#2219

As requested by @PhilippTakacs , this fix might be redundant.


PhilippTakacs:
As already explained: the `g_new0` function returns already clean memory. This is also required, because when `new_nodes` wouldn't be all zero the hashtable wouldn't work correct after resize.

For the failed test I would guess this is because of a to low timeout.

PR #2223, opened 2025-07-29
Python bindings: Use ABI3 wheels

Antelox:
- Enable ABI wheel building
- Rework the workflow in order to build ABI3 wheels for all python versions without any separations
- Bump cibuildwheel to v2.23.3

### DIFF
--- a/.github/workflows/build-wheels-publish.yml
+++ b/.github/workflows/build-wheels-publish.yml
@@ -11,14 +11,6 @@ on:
         options:
           - '0'
           - '1'
-      fullMode:
-        description: 'Full Mode'
-        required: false
-        default: ''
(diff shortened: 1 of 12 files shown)


.github/workflows/build-wheels-publish.yml
```diff
@@ -11,14 +11,6 @@ on:
         options:
           - '0'
           - '1'
-      fullMode:
```
wtdcode, line 14: Why does this get removed?

Antelox: Because it’s no more needed, now all python versions will be tested

wtdcode:
I see how to fix the error.

PR #2225, opened 2025-08-05
Remove ninja installation from macos runners

scribam:
Remove the following warning in GitHub Actions with macOS runners:

> ninja 1.13.1 is already installed and up-to-date. To reinstall 1.13.1, run: brew reinstall ninja


wtdcode:
Thanks!

PR #2226, opened 2025-08-05
Bump cmake minimum required version to 3.10

scribam:
Fix the following warning when using CMake 4.0:
```
CMake Deprecation Warning at CMakeLists.txt:4 (cmake_minimum_required):
  Compatibility with CMake < 3.10 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 # CMake setup for Unicorn 2.
 # By Huitao Chen & Nguyen Anh Quynh, 2019-2020
 
-cmake_minimum_required(VERSION 3.5)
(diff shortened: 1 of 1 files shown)


wtdcode:
I can't recall why we stick to this old version but is it possible to do the `<min>...<max>` way?

scribam:
> I can't recall why we stick to this old version but is it possible to do the `<min>...<max>` way?

Yes, done

wtdcode:
Cool, thanks

PR #2230, opened 2025-08-16
Improve support for MSVC ARM64

scribam:
This do not allow to build Unicorn with MSVC ARM64 as this seems unsupported by qemu currently but it fixes few errors found in CMakelists.txt.
Using only `CMAKE_SIZEOF_VOID_P EQUAL 8` is not enough to detect x86_64 anymore as ARM64 returns the same value

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -117,16 +117,26 @@ if (UNICORN_LOGGING)
     add_compile_options(-DUNICORN_LOGGING)
 endif()
 
-# Some distributions on some rare architecures don't auto link atomic for us and
+# Some distributions on some rare architectures don't auto link atomic for us and
 # we do this manually by adding flags.
 set(ATOMIC_LINKAGE_FIX FALSE)
 if(MSVC)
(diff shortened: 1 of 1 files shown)


wtdcode:
Even you make the compiler right, currently it is not possible to really run unicorn on windows arm64 natively as we have inline assembly. I was working on this but lack machines to test.

wtdcode:
See the blockers here: https://github.com/unicorn-engine/unicorn/blob/master/qemu/accel/tcg/translate-all.c#L930

Another workaround is that we allocate static translation buffers as qemu does on arm64.

scribam:
There is a runner available on GitHub Actions: windows-11-arm (https://github.blog/changelog/2025-08-07-arm64-hosted-runners-for-public-repositories-are-now-generally-available/)

Clearly not the best solution for development but maybe enough for tests.

I don't plan to look at the assembly as it is beyond my skills so feel free to close this PR if you think it does not bring anything useful for now

wtdcode:
Thanks for letting me know and I will leave this PR open for furture arm64 support work.

PR #2242, opened 2025-09-26
basic arm32 instruction hooking

appleflyerv3:
add basic arm32 instruction hooking support for the wfi instruction


bindings/python/unicorn/arm_const.py
```diff
@@ -191,3 +191,7 @@
 UC_ARM_REG_SL = 76
 UC_ARM_REG_FP = 77
 UC_ARM_REG_IP = 78
+
+UC_ARM_INS_INVALID = 0
```
wtdcode, line 195: We have a const_generator.py for this and you should update that instead.

qemu/target/arm/op_helper.c
```diff
@@ -284,6 +284,34 @@ void HELPER(wfi)(CPUARMState *env, uint32_t insn_len)
 {
     CPUState *cs = env_cpu(env);
     int target_el = check_wfx_trap(env, false);
+    uc_engine *uc = env->uc;
+    struct hook * hook;
+    bool skip_wfi = false;
+    bool synced = false;
```
wtdcode, line 290: This variable seems redundant here?

appleflyerv3: sorry which variable?

wtdcode: Oh my bad, it’s okay.

qemu/target/arm/op_helper.c
```diff
@@ -284,6 +284,34 @@ void HELPER(wfi)(CPUARMState *env, uint32_t insn_len)
...
+            uintptr_t pc = GETPC();
+            if (!synced && !uc->skip_sync_pc_on_exit && pc) {
+                cpu_restore_state(uc->cpu, pc, false);
+                synced = true;
+            }
+            JIT_CALLBACK_GUARD_VAR(skip_wfi, ((uc_cb_insn_wfi_t)hook->callback)(env->uc, hook->user_data));
+        }
+
+        // the last callback may already asked to stop emulation
+        if (env->uc->stop_request)
```
wtdcode, line 308: This branch seems not making sense here. If there are pending uc_emu_stop, we should actually break out instead of returning.

Is it possible to also add unit tests btw?

bindings/python/unicorn/arm_const.py
appleflyerv3: should i just run it for python? other bindings do not have support for this feature yet...

wtdcode: At least they should have these enums.

qemu/target/arm/op_helper.c
appleflyerv3: fixed return -> break and added unit tests, resolved.

qemu/target/arm/unicorn_aarch64.c
```diff
@@ -440,6 +440,16 @@ static int arm64_cpus_init(struct uc_struct *uc, const char *cpu_model)
     return 0;
 }
 
+static bool arm64_insn_hook_validate(uint32_t insn_enum)
+{
+    if (insn_enum != UC_ARM64_INS_WFI && insn_enum != UC_ARM64_INS_MRS &&
+        insn_enum != UC_ARM64_INS_MSR && insn_enum != UC_ARM64_INS_SYS &&
+        insn_enum != UC_ARM64_INS_SYSL) {
```
wtdcode, line 447: why `SYSL`?

wtdcode:
Thanks and welcome!

PR #2249, opened 2025-10-22
Do not build ABI3 wheel with no-GIL interpreters

clin1234:


wtdcode:
Sorry for late. Why do you need these changes? As far as I can tell, our bindings shall work with free-threaded runtime?

clin1234:
In the `setup.py`, it requests to build a wheel complying with the Limited API. For free-threaded interpreters, it isn't compatible yet

wtdcode:
My claim is exactly that we are supposed to be compatible with free-threaded interpreters. Could you elaborate why we are not compatible with them?

clin1234:
https://py-free-threading.github.io/porting-extensions/#limited-api-support

wtdcode:
> https://py-free-threading.github.io/porting-extensions/#limited-api-support

We do not call _any_ CPython C API.

PR #2250, opened 2025-10-23
uc_mem_map fix for win64

Ayushkr3:
With reference to this issue
https://github.com/unicorn-engine/unicorn/issues/1841
Fix to the error Exception thrown: Access violation writing location [address]. on Win64 in Visual studio 2019 and and 2022

**Before:**
image

**All tests passed**


wtdcode:
Read our FAQ and we do not accept such “Fix”.

Instead, you should pass exceptions back to applications.

PR #2251, opened 2025-10-25
build: fix __int128_t redefinition with Clang on MSYS2

moleium:
Fixes a build failure on Windows when using the MSYS2 clang64 toolchain.

The compiler already defines `__int128_t` which caused a typedef redefinition error. This PR updates the preprocessor directive in `qemu/include/qemu/int128.h` to avoid this redefinition for all Clang compilers.

```sh
 unicorn/qemu/include/qemu/int128.h:150:16: error: typedef redefinition with different types ('Int128' (aka 'struct Int128') vs '__int128')
  150 | typedef Int128 __int128_t;
      |                ^
1 error generated.
```

### DIFF
--- a/qemu/include/qemu/int128.h
+++ b/qemu/include/qemu/int128.h
@@ -146,7 +146,7 @@ static inline Int128 bswap128(Int128 a)
 #else /* !CONFIG_INT128 */
 
 typedef struct Int128 Int128;
-#if !(defined(_MSC_VER) && defined(__clang__))
+#if !defined(__clang__)
 typedef Int128 __int128_t;
 #endif
 


moleium:
Gentle bump on this. Let me know if any changes are needed.

qemu/include/qemu/int128.h
```diff
@@ -146,7 +146,7 @@ static inline Int128 bswap128(Int128 a)
 #else /* !CONFIG_INT128 */
 
 typedef struct Int128 Int128;
-#if !(defined(_MSC_VER) && defined(__clang__))
+#if !defined(__clang__) && !(defined(_MSC_VER) && defined(__clang__))
```
PhilippTakacs: This always evaluate to false.

PhilippTakacs: Sorry I have missread the change, forget the previous comment.

But it still looks a bit overshooting, is there a way to check for MSYS2?

moleium: `__int128` is a feature provided by the compiler (clang) not the OS environment.
we shouldn't guard a native type behind a distribution flag when the compiler supports it globally regardless of where it's running

PhilippTakacs: Sorry I just had a quick look at the code, you are right with the argument about checking independent of the OS.

But looking at your code this check reduces to `!defined(__clang__)`, because if clang is defined the first check fails and if clang is not defined `(defined(_MSC_VER) && defined(__clang__)` is always false and get inverted.

So a better change would be just check for `!defined(__clang__)`.

wtdcode: This looks exactly equivalent to `!defined(__clang__)`.

wtdcode:
btw you should send to the dev branch.

qemu/include/qemu/int128.h
moleium: Correct. I will fix this.

moleium:
Thanks for feedback. I simplified the macro and changed the base branch to `dev`.

README.md
```diff
@@ -4,6 +4,8 @@ Unicorn Engine
 [![pypi downloads](https://pepy.tech/badge/unicorn)](https://pepy.tech/project/unicorn)
 [![Fuzzing Status](https://oss-fuzz-build-logs.storage.googleapis.com/badges/unicorn.svg)](https://bugs.chromium.org/p/oss-fuzz/issues/list?sort=-opened&can=1&q=proj:unicorn)
 
+**LOOKING FOR CONTRIBUTORS!** See [this](https://github.com/unicorn-engine/unicorn/issues/2237).
```
wtdcode: These changes seem not related.

moleium: Apologies, It seems an upstream commit got tangled in the git history during the previous rebase. I've done a hard reset.

wtdcode: Well... Could you pull the dev branch again? I just merged the commit to fix CI.

moleium: Done.

wtdcode:
Thanks!

PR #2259, opened 2025-11-11
ARM PAC FAQ fix: Read HCR_EL2 instead of clobbering with SCTLR_EL1

droe:
The sample code for setting up PAC is missing the read register call for `HCR_EL2`.  This leads to clobbering `HCR_EL2` with whatever the bits in `SCTLR_EL1` were.

Fix by adding the missing register read call.


wtdcode:
Thanks!

PR #2266, opened 2025-11-23
Fix JAVA binding compilation

gheorghe-crihan:
Apparently, the pointer there is a leftover from the previous implementation. Fails to build on OSX.

A side note: the Java regression test fails for the m68k engine, so another PR created #2267, SR register bug that has been fixed in #2161 and as a result the test needs fixing in turn.


wtdcode:
Hi! Sorry for late response. Could you send this to our dev branch?

gheorghe-crihan:
> Hi! Sorry for late response. Could you send this to our dev branch?

@wtdcode, here, #2284.

PR #2268, opened 2025-11-26
arm: Add missing hflag update after writing xpsr

wuetj:
This adds a missing hflag update after updating the xpsr.
Currently when using an Arm Cortex-M CPU, which does not feature `ARM_FEATURE_M_SECURITY`, the `EXC_RETURN` interrupt is not triggered.
After some debugging it turns out the `v7m_handler_mode` condition here is never true:
https://github.com/unicorn-engine/unicorn/blob/c24c9ebe773ce6fbecb0e39f68ffb23b7326b17f/qemu/target/arm/translate.c#L827-L830
`v7m_handler_mode` is evaluated based on the `HANDLER` flag which is set when rebuilding hflags:
https://github.com/unicorn-engine/unicorn/blob/c24c9ebe773ce6fbecb0e39f68ffb23b7326b17f/qemu/target/arm/helper.c#L11592-L11594
Since setting IPSR using `uc_reg_write` does not rebuild hflags, qemu doesn't realize that it's currently in an exception and returning from an exception causes a hard fault.
This fixes this by rebuilding hflags after updating xPSR.
This doesn't matter for processors with the security feature since the condition is always true in that case. I've updated the test to catch this.

### DIFF
--- a/tests/unit/test_arm.c
+++ b/tests/unit/test_arm.c
@@ -286,8 +286,8 @@ static void test_arm_m_exc_return(void)
     int r_sp = 0x8000;
     uc_hook hook;
 
-    uc_common_setup(&uc, UC_ARCH_ARM, UC_MODE_THUMB | UC_MODE_MCLASS, code,
-                    sizeof(code) - 1, UC_CPU_ARM_CORTEX_A15);
+    uc_common_setup(&uc, UC_ARCH_ARM, UC_MODE_THUMB, code,
+                    sizeof(code) - 1, UC_CPU_ARM_CORTEX_M7);
     OK(uc_mem_map(uc, r_sp - 0x1000, 0x1000, UC_PROT_ALL));
     OK(uc_hook_add(uc, &hook, UC_HOOK_INTR,
                    test_arm_m_exc_return_hook_interrupt, NULL, 0, 0));
--- a/qemu/target/arm/unicorn_arm.c
+++ b/qemu/target/arm/unicorn_arm.c
@@ -154,6 +154,7 @@ static void v7m_msr_xpsr(CPUARMState *env, uint32_t mask, uint32_t reg,
     }
 
     xpsr_write(env, val, xpsrmask);
+    arm_rebuild_hflags(env);
 }
 
 static uc_err read_cp_reg(CPUARMState *env, uc_arm_cp_reg *cp)


tests/unit/test_arm.c
```diff
@@ -286,8 +286,8 @@ static void test_arm_m_exc_return(void)
     int r_sp = 0x8000;
     uc_hook hook;
 
-    uc_common_setup(&uc, UC_ARCH_ARM, UC_MODE_THUMB | UC_MODE_MCLASS, code,
-                    sizeof(code) - 1, UC_CPU_ARM_CORTEX_A15);
+    uc_common_setup(&uc, UC_ARCH_ARM, UC_MODE_THUMB, code,
```
PhilippTakacs, line 289: Are you sure about removing the `UC_MODE_MCLASS`? Because without this the test fails with `UC_ERR_FETCH_UNMAPPED`.

tests/unit/test_arm.c
```diff
@@ -286,8 +286,8 @@ static void test_arm_m_exc_return(void)
...
+                    sizeof(code) - 1, UC_CPU_ARM_CORTEX_M7);
```
PhilippTakacs, line 290: Can you create a second test instead changing the used CPU? Maybe add a CPU parameter to the function and write two tests simply calling `test_arm_m_exc_return`.

tests/unit/test_arm.c
wuetj: `UC_MODE_MCLASS` is deprecated and will simply override the cpu model with `UC_CPU_ARM_CORTEX_M33` upon starting emulation.
https://github.com/unicorn-engine/unicorn/blob/c24c9ebe773ce6fbecb0e39f68ffb23b7326b17f/qemu/target/arm/cpu.c#L2113-L2114
Failing the test is expected without the second commit in this PR. Using the M7 instead of M33 reveals the bug I addressed in the PR.

tests/unit/test_arm.c
wuetj: I'm not sure why `UC_CPU_ARM_CORTEX_A15` is being used here since it's not an M-class CPU. `UC_MODE_MCLASS` overwrites the cpu model with `UC_CPU_ARM_CORTEX_M33`, so this value is ignored anyways.
I could create two separate test cases for `UC_CPU_ARM_CORTEX_M33` (with M security feature) and `UC_CPU_ARM_CORTEX_M4` (no M security feature) if necessary.

tests/unit/test_arm.c
PhilippTakacs: I have tested with `A7` instead of `M7`, sorry.

tests/unit/test_arm.c
PhilippTakacs: This are two different code path so we should have two tests. One for `M33` and one for this bug.

wuetj: The check for `EXC_RETURN` is simply always `true` if the CPU has security extension, since in that case the `EXC_RETURN` code can also be triggered in thread mode. 
https://github.com/unicorn-engine/unicorn/blob/c24c9ebe773ce6fbecb0e39f68ffb23b7326b17f/qemu/target/arm/translate.c#L807-L811
If `EXC_RETURN` works for the M7, it will always work for the M33.

I can add tests for both the M7 and M33 though, if you think it's necessary.

wuetj:
@wtdcode Is there anything preventing this from being merged? What are your thoughts on the test case?

wtdcode:
Sorry for being late. The code looks good to me.

PR #2274, opened 2025-12-17
add uc_hook_set_user_data

PhilippTakacs:
This allows to change the user_data for hooks. This way it's simpler to adopt the hooks for current needs. I.e. change the page table entries for the tlb hook.

### DIFF
--- a/include/unicorn/unicorn.h
+++ b/include/unicorn/unicorn.h
@@ -1150,6 +1150,21 @@ UNICORN_EXPORT
 uc_err uc_hook_del(uc_engine *uc, uc_hook hh);
 
 /*
+ change the user data from a hook callback.
+ This change the user-defined data for a given hook.
+ NOTE: It's undefinde behavior when called on a hook which was not initialized
+ by uc_hook_add or deleted by uc_hook_delete
+ @hh: handle returned by uc_hook_add()
+ @user_data: user-defined data. This will be passed to callback function in its
+      last argument @user_data
+
+ @return UC_ERR_OK on success, or UC_ERR_ARGS when the hook was block or code
+ hook
+*/
+UNICORN_EXPORT
+uc_err uc_hook_set_user_data(uc_hook hh, void *user_data);
+
+/*
  Variables to control which state should be stored in the context.
  Defaults to UC_CTL_CONTEXT_CPU. The options are used in a bitfield
  so to enable more then one content the binary or of the required
--- a/include/unicorn/unicorn.h
+++ b/include/unicorn/unicorn.h
@@ -1154,6 +1154,7 @@ uc_err uc_hook_del(uc_engine *uc, uc_hook hh);
  This change the user-defined data for a given hook.
  NOTE: It's undefinde behavior when called on a hook which was not initialized
  by uc_hook_add or deleted by uc_hook_delete
+ @uc: handle returned by uc_open()
  @hh: handle returned by uc_hook_add()
  @user_data: user-defined data. This will be passed to callback function in its
       last argument @user_data
@@ -1162,7 +1163,7 @@ uc_err uc_hook_del(uc_engine *uc, uc_hook hh);
  hook
 */
 UNICORN_EXPORT
-uc_err uc_hook_set_user_data(uc_hook hh, void *user_data);
+uc_err uc_hook_set_user_data(uc_engine *uc, uc_hook hh, void *user_data);
 
 /*
  Variables to control which state should be stored in the context.
--- a/include/unicorn/unicorn.h
+++ b/include/unicorn/unicorn.h
@@ -1159,8 +1159,8 @@ uc_err uc_hook_del(uc_engine *uc, uc_hook hh);
  @user_data: user-defined data. This will be passed to callback function in its
       last argument @user_data
 
- @return UC_ERR_OK on success, or UC_ERR_ARGS when the hook was block or code
- hook
+ @return UC_ERR_OK on success, or UC_ERR_ARG when the hook was block or code
+ hook and emulation is runnings
 */
 UNICORN_EXPORT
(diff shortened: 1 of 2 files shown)


wtdcode:
LGTM.

PhilippTakacs:
> LGTM.

Shall I rebase this to current dev to rerun the CI or is there anything else blocking the merge?

uc.c
```diff
@@ -2060,6 +2060,17 @@ uc_err uc_hook_del(uc_engine *uc, uc_hook hh)
     return UC_ERR_OK;
 }
 
+UNICORN_EXPORT
+uc_err uc_hook_set_user_data(uc_hook hh, void *user_data)
+{
+    struct hook *hook = (struct hook *)hh;
+    if (hook->type == UC_HOOK_BLOCK || hook->type == UC_HOOK_CODE) {
```
wtdcode, line 2067: Wait why do we exclude these two types?

PhilippTakacs: Because the callbacks are directly called by the tb with the user_data pointer written to the tb. It would require to rebuild the tb to allow change the user_data.

wtdcode: Oh that's correct. Can we inline the pointer to the hook struct in the tb instead? In that case, the modification to `uc_hook` would take effect as well.

PhilippTakacs: No because there is no wrapper function in unicorn for these hooks. The cb function is directly called. I could write a wrapper function which just use the hook_struct. This might affect performance which I would like to avoid in the block case.

wtdcode: Oh okay, I see. I forgot the inline optimization.

A wrapper would hurt performance quite a lot. Spare me some time to think about if we have better ways.

PhilippTakacs: What could be done is only allow to change the user_data of code/block hooks while unicorn isn't running and then clear the affected tb cache. This would only affect performance when you change the user data.

PhilippTakacs: A I missed that I have designed the API without an uc_engine, so I can't test if the emulation is running. So I would say it's ok to just disallow changing user data for these hooks. What do you think?

PhilippTakacs:
@wtdcode can you merge this PR or is there something missing/blocking?

uc.c
wtdcode: What do you think of registering the data pointers as TCG registers? Like how we sync `env->pc`, we could keep the data pointer updated every time before calling the inline callbacks. This imposes the minimum overhead (a pointer read).

Not a compulsory change, and we can keep the current behavior for simplicity.

wtdcode: Another point: according to https://github.com/unicorn-engine/unicorn/blob/c24c9ebe773ce6fbecb0e39f68ffb23b7326b17f/qemu/include/tcg/tcg-op.h#L50

It is still possible to change the data of the code/block hooks if there are more than 1 hook. Maybe it still makes sense to have the API with `uc_engine` and clear tb cache instead?

PhilippTakacs: Ok let's split this up:

1. changing the hook data while emulation is not running:

This is quite simple just add a uc_engine and clear the affected tb cache. I'll add this.

2. changing the hook data while emulation is running

This is quite tricky.

The code you linked is for tb generation, so it is not usable during emulation. Btw. I think the `==` is wrong and should be a `>=`.

Syncing the userdata like `env->pc` just adds a register read. The problem I see is that the user_data doesn't change this often and changing the user_data of code/block hooks during emulation is not the feature everyone needs. Also this adds complex code. So because I believe this is rarely used I would try to avoid changing the tb generation code.

The third option is to handle this is handle it like self modifying code (see #2271 and #2279), so clear the tb cache and break the translation loop. This is also quite complex and has some quirks on using it (mostly block rerunning).

So all in all I would say the best way to handle this is not to allow changing the user data of code/block hooks while emulation is running.

wtdcode: The logic is correct by using == 1, i.e., if there is exactly one code hook we do inline optimisations where changing data is not possible. When we have more hooks, the code here generates a uc_tracecode helper that iterates all hooks so that changing data is rather simple.

If uc is passed with hooks, as far as I can tell, the only invalid case is only having 1 code hook due to inline optimisation . Or would you like to enable inlining for all cases?

wtdcode: For a bit more context, not enabling inlining for more than 1 hooks was due to a bad hack in tcg_optimize leading to segfault at that time. Since I removed the hack recently, we could extend inlining to any number of hooks.

PhilippTakacs: I somehow misinterpreted what this code is doing, thanks for explaining.

wtdcode: Also I would also recommend not allowing changing data during emulation.

PhilippTakacs: So I guess this can be merged now?

include/unicorn/unicorn.h
```diff
@@ -1149,6 +1149,22 @@ uc_err uc_hook_add(uc_engine *uc, uc_hook *hh, int type, void *callback,
...
+ change the user data from a hook callback.
+ This change the user-defined data for a given hook.
+ NOTE: It's undefinde behavior when called on a hook which was not initialized
+ by uc_hook_add or deleted by uc_hook_delete
+ @uc: handle returned by uc_open()
+ @hh: handle returned by uc_hook_add()
+ @user_data: user-defined data. This will be passed to callback function in its
+      last argument @user_data
+
+ @return UC_ERR_OK on success, or UC_ERR_ARGS when the hook was block or code
```
wtdcode: The documents are not updated here (return error during emulation) and `UC_ERR_ARGS` seems a typo.

uc.c
wtdcode: yeah, look good now except the small typo and documents above.

PR #2276, opened 2025-12-20
Avoid memory corruption in arm context save/restore code

### LINKED ISSUE #2195: uc_context_save/uc_context_restore (still) does not work properly for some ARM CPUs
`uc_context_save()` stores dynamically allocated pointers which prevents the context to be restored into another `uc_engine` and leads to double frees on `uc_close()` and other memory problems.

The issue has already been discussed in #1635 . It seems it has not been resolved properly for all ARM CPUs.

Code to reproduce (simplified version of `test_arm_context_save`):
```
#include <unicorn/unicorn.h>

void main() {
    uc_engine *uc;
    uc_engine *uc2;
    uc_context *ctx;

    uc_open(UC_ARCH_ARM, UC_MODE_THUMB, &uc);
    uc_ctl_set_cpu_model(uc, UC_CPU_ARM_CORTEX_M4);

    uc_context_alloc(uc, &ctx);
    uc_context_save(uc, ctx);

    uc_open(UC_ARCH_ARM, UC_MODE_THUMB, &uc2);
    uc_ctl_set_cpu_model(uc2, UC_CPU_ARM_CORTEX_M4);

    uc_context_restore(uc2, ctx);
    uc_context_free(ctx);
    uc_close(uc); // Frees some memory which will also be freed by uc_close(uc2)
    uc_close(uc2); // free(): double free detected in tcache 2
}
```

The affected CPUs:
```
UC_CPU_ARM_CORTEX_M3
UC_CPU_ARM_CORTEX_M4
UC_CPU_ARM_CORTEX_M7
UC_CPU_ARM_CORTEX_M33
UC_CPU_ARM_CORTEX_R5
UC_CPU_ARM_CORTEX_R5F
```

For `UC_CPU_ARM_CORTEX_M33` the pertinent pointers are in `pmsav8` and `sau`, for the rest they are in `pmsav7`.

droe:
Addresses multiple issues in the arm context save/restore code:
-   When restoring from context, do not clobber pointers in CPUARMState with data from the context buffer.  This causes use-after-free bugs usually manifesting as double-free crashes.
-   When saving to context, do not copy pointers from CPUARMState to the context buffer, to avoid leaking pointers to the context.  Instead, zero the respective area of the context buffer, to ensure that reg_read and reg_write can access fields outside of the copied range.  Specifically, env->uc needs to be NULL when accessed for a context.
-   When restoring to a CPU with a different nr than the source CPU, copy min(nr, ctx_nr) instead of nothing at all.
-   Add regression tests checking all CPU models for arm and arm64 for trivial double-free crashes.

Fixes #2195

### DIFF
--- a/qemu/target/arm/unicorn_arm.c
+++ b/qemu/target/arm/unicorn_arm.c
@@ -630,9 +630,18 @@ static int arm_cpus_init(struct uc_struct *uc, const char *cpu_model)
     return 0;
 }
 
+// The strategy is to copy the top part of CPUARMState that does not contain
+// any pointers into the context buffer, and zero-fill the remainder.  For all
+// fields in the pointer range, data that needs to be saved to the context must
+// be manually appended to the end of the buffer.  We need to zero the pointer
+// portion in the context buffer, because reg_read/reg_write are accessing some
+// of these fields when accessing registers in a context.
+#define CONTEXT_COPYSIZE offsetof(CPUARMState, pmsav7)
+#define CONTEXT_ZEROSIZE (sizeof(CPUARMState) - CONTEXT_COPYSIZE)
+
 static size_t uc_arm_context_size(struct uc_struct *uc)
 {
-    size_t ret = offsetof(CPUARMState, cpu_watchpoint);
+    size_t ret = CONTEXT_COPYSIZE + CONTEXT_ZEROSIZE;
     ARMCPU *cpu = (ARMCPU *)uc->cpu;
     CPUARMState *env = (CPUARMState *)&cpu->env;
     uint32_t nr;
@@ -708,8 +717,10 @@ static uc_err uc_arm_context_save(struct uc_struct *uc, uc_context *context)
         p += sizeof(uint32_t) * nr;                                            \
     }
     p = context->data;
-    memcpy(p, uc->cpu->env_ptr, uc->cpu_context_size);
-    p += uc->cpu_context_size;
+    memcpy(p, uc->cpu->env_ptr, CONTEXT_COPYSIZE);
+    p += CONTEXT_COPYSIZE;
+    memset(p, 0, CONTEXT_ZEROSIZE);
+    p += CONTEXT_ZEROSIZE;
 
     nr = cpu->pmsav7_dregion;
     ARM_ENV_SAVE(env->pmsav7.drbar)
@@ -739,8 +750,8 @@ static uc_err uc_arm_context_restore(struct uc_struct *uc, uc_context *context)
     ctx_nr = *(uint32_t *)p;                                                   \
     if (ctx_nr != 0) {                                                         \
         p += sizeof(uint32_t);                                                 \
-        if (field && ctx_nr == nr) {                                           \
-            memcpy(field, p, sizeof(uint32_t) * ctx_nr);                       \
+        if (field && nr != 0) {                                                \
+            memcpy(field, p, sizeof(uint32_t) * (ctx_nr > nr ? nr : ctx_nr));  \
         }                                                                      \
         p += sizeof(uint32_t) * ctx_nr;                                        \
     } else {                                                                   \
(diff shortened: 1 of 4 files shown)


qemu/target/arm/unicorn_arm.c
```diff
@@ -765,8 +776,6 @@ static uc_err uc_arm_context_restore(struct uc_struct *uc, uc_context *context)
     ARM_ENV_RESTORE(env->sau.rlar)
 
 #undef ARM_ENV_RESTORE
-    // Overwrite uc to our uc
-    env->uc = uc;
```
wtdcode, line 769: Is this removed by design?

wtdcode:
LGTM except ` Specifically, env->uc needs to be NULL when accessed for a context.`. Could you elaborate the reason?

droe:
The reason is that AArch32 `reg_write` wants to write to `env->uc->thumb` when setting `pc`, and it checks if `uc` is NULL to check if it is operating on a live uc or a context, see:
https://github.com/unicorn-engine/unicorn/blob/d4076e1e74e47976effbd349ee132d6b1b51940b/qemu/target/arm/unicorn_arm.c#L440-L443

`reg_write` writing to `uc->thumb` looks like a potential bug to me, but I did not want to investigate and touch that in this PR.  Maybe this PR should address that too, what do you think?

wtdcode:
> The reason is that AArch32 `reg_write` wants to write to `env->uc->thumb` when setting `pc`, and it checks if `uc` is NULL to check if it is operating on a live uc or a context, see:
> 
> https://github.com/unicorn-engine/unicorn/blob/d4076e1e74e47976effbd349ee132d6b1b51940b/qemu/target/arm/unicorn_arm.c#L440-L443
> 
> `reg_write` writing to `uc->thumb` looks like a potential bug to me, but I did not want to investigate and touch that in this PR. Maybe this PR should address that too, what do you think?

This points to https://github.com/unicorn-engine/unicorn/commit/a903fa182be77ec7f0ef07e58015bfcd8782e33e

Would you mind rebasing your branch to the latest dev so that we can have CI to run?

droe:
Rebased on latest `dev`.

wtdcode:
mingw test failure is Github Action bug.

qemu/target/arm/unicorn_arm.c
droe: Yes.  We no longer clobber `env->uc` during restore, hence we do not need to fix `env->uc` to the correct `uc` after clobbering it with a potentially stale or mismatching pointer from the context.

droe:
> This points to https://github.com/unicorn-engine/unicorn/commit/a903fa182be77ec7f0ef07e58015bfcd8782e33e

I am not sure why thumb is a property of uc and not part of the CPU state, but thinking about it some more, the behaviour makes sense.  I think this PR can go in as is (unless there are any concerns of course).

droe:
Rebased on top of latest `dev` and resolved merge conflict.

qemu/target/arm/unicorn_arm.c
```diff
@@ -630,9 +630,18 @@ static int arm_cpus_init(struct uc_struct *uc, const char *cpu_model)
     return 0;
 }
 
+// The strategy is to copy the top part of CPUARMState that does not contain
+// any pointers into the context buffer, and zero-fill the remainder.  For all
+// fields in the pointer range, data that needs to be saved to the context must
+// be manually appended to the end of the buffer.  We need to zero the pointer
+// portion in the context buffer, because reg_read/reg_write are accessing some
+// of these fields when accessing registers in a context.
+#define CONTEXT_COPYSIZE offsetof(CPUARMState, pmsav7)
```
wtdcode, line 640: Back reviewing this. `pmsav7` seems to be missing a few model fields like `pmsav8`, causing regression.

droe: Good catch, thanks.  Fixed and rebased to latest `dev`.

PR #2282, opened 2026-01-15
add uc_emu_run

PhilippTakacs:
This function allows to run from the current instruction pointer. Can be used when you allready manage all registers, so you don't need to read the instruction pointer and start unicorn with it.

### DIFF
--- a/include/unicorn/unicorn.h
+++ b/include/unicorn/unicorn.h
@@ -1094,6 +1094,27 @@ uc_err uc_emu_start(uc_engine *uc, uint64_t begin, uint64_t until,
                     uint64_t timeout, size_t count);
 
 /*
+ Emulate machine code starting from the current instruction pointer
+
+ @uc: handle returned by uc_open()
+ @until: address where emulation stops (i.e. when this address is hit)
+ @timeout: duration to emulate the code (in microseconds). When this value is 0,
+        we will emulate the code in infinite time, until the code is finished.
+ @count: the number of instructions to be emulated. When this value is 0,
+        we will emulate all the code available, until the code is finished.
+
+ NOTE: The internal states of the engine is guranteed to be correct if and only
+       if uc_emu_start returns without any errors or errors have been handled in
+       the callbacks.
+
+ @return UC_ERR_OK on success, or other value on failure (refer to uc_err enum
+   for detailed error).
+*/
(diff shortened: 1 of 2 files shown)


include/unicorn/unicorn.h
```diff
@@ -1047,6 +1047,27 @@ UNICORN_EXPORT
...
+
+ NOTE: The internal states of the engine is guranteed to be correct if and only
+       if uc_emu_start returns without any errors or errors have been handled in
+       the callbacks.
+
+ @return UC_ERR_OK on success, or other value on failure (refer to uc_err enum
+   for detailed error).
+*/
+UNICORN_EXPORT
+uc_err uc_emu_run(uc_engine *uc, uint64_t until, uint64_t timeout,
```
wtdcode, line 1114: I like the idea but I feel that the state after `uc_emu_start` is not always safe to start immediately, like those after errors.

`uc_emu_start` with a pc write shall reset states in most cases, but `uc_emu_run` does not. Do you think that introduces potential issues?

PhilippTakacs: Might be, I need to think about it a bit. But I have a few questions:

* Which error states are not save to start immediately from?
* does a pc write restore from a context does have the same effect?

wtdcode: > Which error states are not save to start immediately from?

I can not really come up with one so I think this PR is mostly safe to go but let me review others firstly.

> does a pc write restore from a context does have the same effect?

Writing a pc is always making the emulator to a clean state, at least I design it in this way.

PhilippTakacs:
An alternative way would be to implement this the other way around. So read the pc and call uc_emu_start. This way the pc write is done every time and state is reseted.

wtdcode:
> An alternative way would be to implement this the other way around. So read the pc and call uc_emu_start. This way the pc write is done every time and state is reseted.

But then that is not really worth a new API because users could do that by themselves.

PhilippTakacs:
I have added this to have a simpler API, not to avoid the reg_write. I just run into a bug because I forgot to update my local pc variable. I believe I'm not the only one who depends on a correct set pc register and just reads the pc to pass it to uc_emu_start. I would say having a simple to use API for this cases is worth it.

PR #2285, opened 2026-01-20
fix: emulator CPU state sync on hook only should sync pc, and fix pc syncing on mmio_map callback

appleflyerv3:


wtdcode:
@appleflyerv3 have you solved your issue with this PR? I'm a bit losing the context.

appleflyerv3:
yes its fixed in this PR, tested and works

appleflyerv3:
pc is always synced properly when using hooks whereby it doesn't affect anything other than the pc on sync, and pc syncing has been added for mmio map

appleflyerv3:
you may also observe the added CI tests in this PR, and they pass on my PR but not on latest 2.1.5-dev

PhilippTakacs:
@wtdcode is there anything missing for this to be merged?

There is a merge conflict with #2344, but as far as I see this pr is a better fix for #2309

PR #2286, opened 2026-01-26
Add Windows ARM64 build support

expend20:
The changes include:
* ARM64 architecture detection in CMakeLists.txt
* ARM64 Windows SEH trampoline in translate-all.c
* Native ARM64 setjmp/longjmp assembly wrappers
* MSVC compatibility fixes for aarch64 TCG backend
* Disabled x86-specific intrinsics on ARM64

### DIFF
--- a/qemu/util/setjmp-wrapper-win32-arm64.asm
+++ b/qemu/util/setjmp-wrapper-win32-arm64.asm
@@ -1,41 +1,26 @@
-; setjmp/longjmp wrapper for Windows ARM64 (MSVC/armasm64)
+; setjmp/longjmp for Windows ARM64 (MSVC armasm64)
 ;
-; This is a native assembly implementation that doesn't create a stack frame,
-; avoiding the stack corruption issues that occur with a C wrapper.
+; Based on musl libc aarch64 setjmp/longjmp implementation.
+; https://git.musl-libc.org/cgit/musl/tree/src/setjmp/aarch64/setjmp.s
+; https://git.musl-libc.org/cgit/musl/tree/src/setjmp/aarch64/longjmp.s
 ;
-; On ARM64 Windows, wrapping setjmp in a C function doesn't work because:
-; 1. The C wrapper creates its own stack frame
-; 2. setjmp saves the state inside the wrapper
-; 3. After setjmp returns, the wrapper's stack frame is released
-; 4. Other code runs and reuses that stack space
-; 5. When longjmp is called, it tries to return to corrupted stack
+; Custom implementations are needed because the CRT longjmp calls
+; RtlUnwind for stack unwinding, which crashes when JIT-generated code
+; frames (with no SEH unwind metadata) are on the stack.
+; See setjmp-wrapper-win32-arm64.md for details.
 ;
-; This assembly implementation:
-; 1. Saves all callee-saved registers directly to jmp_buf
-; 2. Sets Frame=0 to disable stack unwinding
-; 3. Returns 0 directly without calling CRT
-; 4. longjmp restores registers and returns to original caller
-;
-; jmp_buf layout for ARM64 Windows (per MSVC _JUMP_BUFFER):
-;   Offset 0x00: Frame (set to 0 to disable unwinding)
-;   Offset 0x08: Reserved
-;   Offset 0x10: X19
-;   Offset 0x18: X20
-;   Offset 0x20: X21
-;   Offset 0x28: X22
-;   Offset 0x30: X23
-;   Offset 0x38: X24
-;   Offset 0x40: X25
-;   Offset 0x48: X26
-;   Offset 0x50: X27
-;   Offset 0x58: X28
-;   Offset 0x60: Fp (X29)
-;   Offset 0x68: Lr (X30)
-;   Offset 0x70: Sp
-;   Offset 0x78: Fpcr (4 bytes) + Fpsr (4 bytes) = 8 bytes combined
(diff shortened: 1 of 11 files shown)


wtdcode:
Hi! Thanks for your PR. Have you checked https://github.com/unicorn-engine/unicorn/pull/2230 ?

qemu/util/setjmp-wrapper-win32-arm64.asm
```diff
@@ -0,0 +1,124 @@
+; setjmp/longjmp wrapper for Windows ARM64 (MSVC/armasm64)
```
wtdcode: Hmm, why do we need this wrapper? This looks AI generated, is there any case to illustrate this?

wtdcode:
btw, you are supposed to send to our dev branch.

qemu/util/setjmp-wrapper-win32-arm64.asm
expend20: I wasn't able to make it work withot this wrapper

TLDR

# Why setjmp/longjmp wrappers are needed on Windows ARM64

## Background

Unicorn's QEMU TCG uses `setjmp`/`longjmp` to exit from JIT-generated code back
to the CPU execution loop (e.g. on HLT, exceptions, or memory faults). On
Windows ARM64, both the CRT `setjmp` and `longjmp` are unsuitable for this
because they interact with Windows Structured Exception Handling (SEH) stack
unwinding, which cannot traverse JIT-generated code frames that lack unwind
metadata.

## Why `_setjmp_wrapper` is needed

The CRT `_setjmp` on ARM64 saves the caller's frame pointer into `jmp_buf`,
which `longjmp` later uses to drive `RtlUnwind`. The wrapper instead stores
`Frame=0` in the `jmp_buf`, which signals "no stack unwinding needed."

Without it: **linker error** — `os-win32.h` declares
`extern int _setjmp_wrapper(jmp_buf)` and the `setjmp` macro expands to it, so
all three callsites (`cpu-exec.c`, `translate-all.c`, `translate.c`) reference
this symbol.

## Why `_longjmp_wrapper` is needed

Even with a correct `jmp_buf` (Frame=0), the CRT `longjmp` still calls
`__longjmp_internal` → `RtlUnwind` → `RtlUnwindEx`, which attempts to walk the
stack. When JIT-generated code is on the stack (no SEH unwind info), this fails
with exception `0xC00000FF`.

The wrapper bypasses the CRT entirely — it restores callee-saved registers
directly from the `jmp_buf` with `ldp`/`ldr` instructions and `ret`s to the
saved return address.

Without it: **runtime crash** — confirmed via debugger:

```
ntdll!RtlRaiseStatus          (exception 0xC00000FF)
ntdll!RtlUnwindEx              ← stack unwinding fails here
ntdll!RtlUnwind
VCRUNTIME140!__longjmp_internal
VCRUNTIME140!longjmp
unicorn!cpu_loop_exit_x86_64
unicorn!helper_hlt_x86_64
0x000001b480000184             ← JIT code (no unwind info)
```

46 of 54 x86 tests crash with exit code `0xC00000FF`.

## Why x64 doesn't need a longjmp wrapper

On x64, `_setjmp` accepts a second parameter (frame pointer). Passing `NULL`
disables stack unwinding in `longjmp`. ARM64's CRT has no equivalent mechanism,
so both wrappers are required.

expend20:
> Hi! Thanks for your PR. 

Hey, what's up?

> Have you checked #2230 ?

not really, but that's clearly not sufficient to make it work

> btw, you are supposed to send to our dev branch.

rebased, I see Rust bindings are gone, so am I supposed to open two pair PRs (second one to the unicorn-engine-rs) simultaneosly? or should I wait until you approve this one?

PS: I've also done AVX/AVX2 port https://github.com/unicorn-engine/unicorn/commit/6546bd5c450f06d49ef4326df067ca1b0b64c5d5 is it something you interested in having in dev banch as well?

qemu/util/setjmp-wrapper-win32-arm64.asm
wtdcode: If that is really the case, I would suggest taking implementation from https://git.musl-libc.org/cgit/musl/tree/src/setjmp/aarch64

wtdcode:
> > Hi! Thanks for your PR.
> 
> Hey, what's up?
> 
> > Have you checked #2230 ?
> 
> not really, but that's clearly not sufficient to make it work
> 
> > btw, you are supposed to send to our dev branch.
> 
> rebased, I see Rust bindings are gone, so am I supposed to open two pair PRs (second one to the unicorn-engine-rs) simultaneosly? or should I wait until you approve this one?

You can open both.

> 
> PS: I've also done AVX/AVX2 port [6546bd5](https://github.com/unicorn-engine/unicorn/commit/6546bd5c450f06d49ef4326df067ca1b0b64c5d5) is it something you interested in having in dev banch as well?

Depending on how you port the code (like are you porting them from future qemu releases?).

qemu/util/setjmp-wrapper-win32-arm64.asm
expend20: thanks! switched to musl's implementation

expend20:
> Depending on how you port the code (like are you porting them from future qemu releases?).

yeah, exactly like that from latest Qemu repo. Is there any other way?

expend20:
added CI build for windows-arm64 static/shared too

expend20:
hi @wtdcode, is there anything else I can do to facilitate this merge?

PR #2288, opened 2026-02-06
Added missing ppc_const import to python __init__.py bindings

Veryyes:
`__init__.py` in the python bindings was missing the import for `ppc_const`. This PR adds that to be consistent with all the other imports


wtdcode:
Thanks & welcome!

PR #2289, opened 2026-02-10
PPC: Re-enable timebase (TB) registers with 512MHz frequency

Theldus:
Hi,   
I'm building a [user-space AIX emulator](https://github.com/Theldus/aix-user) and noticed that AIX's libc implementation of `clock_gettime(CLOCK_MONOTONIC)` uses the PowerPC mftb/mftbu instructions to read the timebase for monotonic timing.

Currently, these registers always return 0 because PR #1910 stubbed them out to prevent crashes from uninitialized tb_env. This causes CLOCK_MONOTONIC to always return zero'ed values.

This PR fixes this by initializing tb_env during CPU reset with cpu_ppc_tb_init() and re-enabling timebase read/write functions (spr_read_tbl/tbu).

Tested with: AIX binaries + `test_ppc`.

### DIFF
--- a/qemu/target/ppc/translate_init.inc.c
+++ b/qemu/target/ppc/translate_init.inc.c
@@ -205,7 +205,6 @@ static void spr_write_decr(DisasContext *ctx, int sprn, int gprn)
 
 /* SPR common to all non-embedded PowerPC, except 601 */
 /* Time base */
-#if 0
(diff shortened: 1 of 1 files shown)


wtdcode:
Looks good but you should send to our dev branch. Thanks in advance.

Theldus:
Oh, sorry for the confusion, I've sent it again on PR #2296.
Thanks for the clarification.

Issue #2290, opened 2026-02-10
code_gen_buffer_handler calculates size incorrectly under some circumstance


randomdude:
Hi all. I recently hit the following problem with Unicorn. Unfortunately I can't share a repro case, as the emulated code is from a third party product. I'm making the bugreport as verbose as possible in an attempt to compensate for this.

I saw this when I tried to run Unicorn from Python:

```
Traceback (most recent call last):
  File "D:\code\unicorn\repro\foo.py", line 71, in <module>
    run()
  File "D:\code\unicorn\repro\foo.py", line 67, in run
    unicornEng.emu_start(RIP, RIP + 0x10, count=1)
  File "D:\code\unicorn\repro\venv\lib\site-packages\unicorn\unicorn_py3\unicorn.py", line 763, in emu_start
    status = uclib.uc_emu_start(self._uch, begin, until, timeout, count)
OSError: exception: access violation writing 0x0000019DBFC33000
```

This is on a **Win10 amd64** system, using the current `pip`-packaged sources. I also verified this happens on the latest git head (`c24c9ebe`).

I built a debug version of Unicorn, and saw that `VirtualProtect` was failing when called from `code_gen_buffer_handler` under a specific circumstance. In `translate-all.c` we can see the invocation in question (line 893):

```
if (VirtualAlloc(base, size, MEM_COMMIT, PAGE_EXECUTE_READWRITE)) {
```

This failed:

```
0:000> r rax
rax=0000000000000000
0:000> !gle
LastErrorValue: (Win32) 0x1e7 (487) - Attempt to access invalid address.
LastStatusValue: (NTSTATUS) 0xc0000018 - {Conflicting Address Range}  The specified address range conflicts with the address space.
```

The arguments that `VirtualProtect` is called with seem incorrect - specifically, the size:
```
0:000> dv
    size = 0x400000
    base_end = 0x000001f2`40033000 "--- memory read error at address 0x000001f2`40033000 ---"
    right = 0x000001f2`40000000 "???"
    base = 0x000001f2`3fc33000 "--- memory read error at address 0x000001f2`3fc33000 ---"
    left = 0x000001f2`00000000 "USWVATAUAVAWH???"
```

`VirtualProtect` is being invoked with a start of `base` and size of `size` (so  `0x000001f2_3fc33000` and `0x400000`). This results in `000001f2_40033000`, which is greater than the `right` value of `0x000001f2_40000000`, which the reservation was previously made according to. And so, the call fails.

[...]

PR #2293, opened 2026-02-11
Correct size calculation in code_gen_buffer_handler

randomdude:
See issue https://github.com/unicorn-engine/unicorn/issues/2290


wtdcode:
Nice catch and I confirmed the fix is correct.

PR #2294, opened 2026-02-12
Enable PPC64 support

freeqaz:
## Summary

- Remove the "currently unsupported" qualifier from `UC_MODE_PPC64` and enable PPC64 emulation
- Change the default PPC64 CPU model from POWER10 to 970 v2.2 for broader compatibility, and fix the cpu_model bounds check in the else branch
- Add comprehensive PPC64 tests (C unit tests, Python binding test, Rust binding test) and a PPC64 sample

## Details

PPC64 mode was declared but non-functional — initializing with `UC_MODE_PPC64` would select the POWER10 CPU model, which has dependencies that prevented it from working correctly. Switching the default to 970 v2.2 resolves this.

The C test suite adds 8 PPC64 tests covering:
- Integer add (small and 64-bit large values)
- Floating-point add
- System call interrupt
- Condition register read/write
- 64-bit load (`ld`) and store (`std`)
- SPR/timebase register access

All existing tests (PPC32, x86, ARM, ARM64, ctl, mem) continue to pass.

## Files changed

- `include/unicorn/unicorn.h` — Update `UC_MODE_PPC64` comment
- `qemu/target/ppc/translate_init.inc.c` — Default CPU model + bounds check fix
- `tests/unit/test_ppc.c` — 8 new PPC64 unit tests
- `samples/sample_ppc.c` — PPC64 sample
- `bindings/python/tests/test_ppc.py` — Python PPC64 test
- `bindings/rust/unicorn-engine/src/tests/ppc.rs` — Rust PPC64 test

## Test plan

- [x] PPC32 tests pass
- [x] PPC64 tests pass (8 new tests)
- [x] x86 tests pass
- [x] ARM/ARM64 tests pass
- [x] ctl and mem tests pass
- [x] Build succeeds with all architectures enabled

Heads up: I used LLMs to help me write this code. I've tried to clean up to slop, but if you notice anything off please share feedback.

### DIFF
--- a/.github/workflows/build-wheels-publish.yml
+++ b/.github/workflows/build-wheels-publish.yml
@@ -134,7 +134,7 @@ jobs:
           python3 -m pip install -U pip wheel
           python3 -m wheel tags --python-tag='py2' --abi-tag=none wheelhouse/*cp37*.whl
 
-      - uses: LizardByte/actions/actions/setup_python@v2025.715.25226
+      - uses: LizardByte/actions/actions/setup_python@v2026.212.22356
         if: (runner.os == 'Linux' && (matrix.arch == 'x86_64' || matrix.arch == 'aarch64') && matrix.cibw_build == 'cp*-manylinux*') || (runner.os == 'macOS' && matrix.arch == 'arm64') || (runner.os == 'Windows' && matrix.arch == 'AMD64')
         with:
           python-version: 2.7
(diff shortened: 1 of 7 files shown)


wtdcode:
You are having CI failures because you should send to our dev branch.

Scottcjn:
Good to see PPC64 support landing. The 970 v2.2 CPU model choice is solid — that is the G5 processor.

We have real 970 (Power Mac G5 Dual 2.0GHz) and POWER8 S824 hardware available. Happy to validate emulation accuracy against physical hardware if that would be useful — comparing instruction behavior, cache semantics, and AltiVec/VSX results between Unicorn output and bare-metal execution.

PR #2296, opened 2026-02-17
PPC: Re-enable timebase (TB) registers with 512MHz frequency

Theldus:
Hi,   
I'm building a [user-space AIX emulator](https://github.com/Theldus/aix-user) and noticed that AIX's libc implementation of `clock_gettime(CLOCK_MONOTONIC)` uses the PowerPC mftb/mftbu instructions to read the timebase for monotonic timing.

Currently, these registers always return 0 because PR #1910 stubbed them out to prevent crashes from uninitialized tb_env. This causes CLOCK_MONOTONIC to always return zero'ed values.

This PR fixes this by initializing tb_env during CPU reset with cpu_ppc_tb_init() and re-enabling timebase read/write functions (spr_read_tbl/tbu).

Tested with: AIX binaries + `test_ppc`.

Small note: This is a re-send of PR #2289, but to the right branch, sorry for the mistake.

### DIFF
--- a/qemu/target/ppc/translate_init.inc.c
+++ b/qemu/target/ppc/translate_init.inc.c
@@ -205,7 +205,6 @@ static void spr_write_decr(DisasContext *ctx, int sprn, int gprn)
 
 /* SPR common to all non-embedded PowerPC, except 601 */
 /* Time base */
-#if 0
 static void spr_read_tbl(DisasContext *ctx, int gprn, int sprn)
 {
     TCGContext *tcg_ctx = ctx->uc->tcg_ctx;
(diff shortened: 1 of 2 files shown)


wtdcode:
LGTM. Waiting for CI.

wtdcode:
Though the PR could be merged as it is, would you mind a small test to `tests/unit` illustrate your case?

Theldus:
Hi,
I've added the test `test_ppc32_spr_mftb()`: it reads the upper and lower value, do some busy-wait, and read it again, if the values changed, the timer is working.

I have intentionally not checked for wrap arounds since the idea is to just check if the timer is increasing/changing the values.

wtdcode:
Thanks and welcome!

Issue #1955, opened 2024-05-13
Execution of xgetbv instruction and setting up of the XCR register.


Nitr0-G:
Hello, guys! 
How can I emulate the xgetbv instruction? Judging by the latest updates, there is support for it, but I have not found anywhere examples of configuring the XCR register for it. I wrote the code below, but apparently I still do not have its execution and get: `uc_emu_start(uc, ctx.m_ExecuteFromRip, ctx.m_ImageEnd, 0, 0); returned with UC_ERR_INSN_INVALID when xgetbv executing`

My code for configuring the XCR register:
```
	DWORD_PTR XCR0 = _xgetbv(0);

	uc_x86_msr MsrIa32VmxCr0Fixed0;
	MsrIa32VmxCr0Fixed0.rid = (uint32_t)Msr::kIa32VmxCr0Fixed0;
	MsrIa32VmxCr0Fixed0.value = XCR0;

	uc_x86_msr MsrIa32VmxCr0Fixed1;
	MsrIa32VmxCr0Fixed0.rid = (uint32_t)Msr::kIa32VmxCr0Fixed1;
	MsrIa32VmxCr0Fixed0.value = XCR0;

	uc_reg_write(m_uc, UC_X86_REG_MSR, &MsrIa32VmxCr0Fixed0);
	uc_reg_write(m_uc, UC_X86_REG_MSR, &MsrIa32VmxCr0Fixed1);
```

Verison of UC: I'm using unicorn from the master branch.

Nitr0-G:
poc

```
#define X86_CODE64                                                             \
    "\x41\xBC\x3B\xB0\x28\x2A\x49\x0F\xC9\x90\x4D\x0F\xAD\xCF\x49\x87\xFD\x90" \
    "\x48\x81\xD2\x8A\xCE\x77\x35\x48\xF7\xD9\x4D\x29\xF4\x49\x81\xC9\xF6\x8A" \
    "\xC6\x53\x4D\x87\xED\x48\x0F\xAD\xD2\x49\xF7\xD4\x48\xF7\xE1\x4D\x19\xC5" \
    "\x4D\x89\xC5\x48\xF7\xD6\x41\xB8\x4F\x8D\x6B\x59\x4D\x87\xD0\x68\x6A\x1E" \
    "\x09\x3C\x59\x0F\x01\xD0"
    
// memory address where emulation starts
#define ADDRESS 0x1000000

static void test_x86_64(void)
{
[...]

PR #2299, opened 2026-02-23
Expose x86 XCR0 register

riptl:
Expose the XCR0 register for x86 to allow users to control XSAVE
behavior. This register is a bit odd. It's the only one of its type
(extended control register), and usually accessed via XGETBV/XSETBV.

Note that neither the existing CR or MSR APIs allow accessing XCR0. 

I think that modifying XCR0 can potentially change the BND config,
so I added a call to `cpu_sync_bndcs_hflags`.


riptl:
Related: https://github.com/unicorn-engine/unicorn/issues/1955

riptl:
Just saw the `dev` branch, rebasing.

wtdcode:
Looks good, thanks

PR #2305, opened 2026-03-05
Fixed tcg temp alloc without free in mips tcg translation

justdaniel-gh:
Fixes #2240

Calls to tcg_const_i32 end up calling tcg_temp_alloc (in tcg.c:tcg_temp_new_internal), which allocates a temp var in ctx->temps[]. tcg_temp_free_i32 should be called to free the allocated temp var.

Without it, ctx->temps will eventually overflow ctx->ops and SEGFAULT.

Issue #2240, opened 2025-09-26
Segfault in uc_ctl_request_cache with `UC_ARCH_MIPS`


hgarrereyn:
Hi, using `uc_ctl_request_cache` with `UC_ARCH_MIPS` results in a segfault.

Interestingly this seems to only happen on MIPS (both 32 and 64 bit).

Tested on the most recent commit `c24c9ebe`.

(found via automated fuzzing).

The following testcase reproduces the crash:

**testcase.cpp**
```cpp
#include <cstdint>
extern "C" {
#include "/fuzz/install/include/unicorn/unicorn.h"
#include "/fuzz/install/include/unicorn/mips.h"
}
int main() {
  uc_engine *uc = nullptr;
  if (uc_open(UC_ARCH_MIPS, UC_MODE_MIPS32, &uc) != UC_ERR_OK) return 0;
  uint64_t base = 0x1000;
  uc_mem_map(uc, base, 0x1000, UC_PROT_READ | UC_PROT_EXEC | UC_PROT_WRITE);
  unsigned char code[16]; for (int i=0;i<16;i++) code[i]=0x90;
  uc_mem_write(uc, base, code, sizeof(code));
  uc_tb tb{};
  uc_ctl_request_cache(uc, base, &tb);
  return 0;
}
```

**ASAN crash**
```
==12==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x55d18bf4af76 bp 0x000000000006 sp 0x7ffccff5a800 T0)
==12==The signal is caused by a WRITE memory access.
==12==Hint: address points to the zero page.
    #0 0x55d18bf4af76 in tcg_emit_op_mipsel /fuzz/src/qemu/tcg/tcg.c:2132:5
    #1 0x55d18bf77977 in tcg_gen_op2_mipsel /fuzz/src/qemu/tcg/tcg-op.c:51:17
    #2 0x55d18bf49400 in tcg_gen_op2i_i32 /fuzz/src/qemu/include/tcg/tcg-op.h:134:5
    #3 0x55d18bf49400 in tcg_gen_movi_i32 /fuzz/src/qemu/include/tcg/tcg-op.h:416:5
    #4 0x55d18bf49400 in tcg_const_i32_mipsel /fuzz/src/qemu/tcg/tcg.c:1104:5
    #5 0x55d18c0efa64 in mips_tr_translate_insn /fuzz/src/qemu/target/mips/translate.c:30951:19
[...]

PR #2307, opened 2026-03-07
Add error handling for tcg temps overflow

justdaniel-gh:
Related to #2240, this adds further error handling for cases where tcg_ctx->temps could be overflowed within `tcg_temp_alloc`. In addition, it adds a guard for a potential null ptr deref within `temp_load` when ts->mem_base is NULL, by allocating the frame.

Similar fixes were added in upstream qemu:
* https://github.com/qemu/qemu/commit/e139bc4b1772575e1f2dcf8e3dbe1df2b684ef1f 
* https://github.com/qemu/qemu/commit/ae30e86661b0f48562cd95918d37cbeec5d02262 
* https://github.com/qemu/qemu/commit/db6b7d0c6936cd209e3e8d95aea61ad29ceef5e6

When applying the fix for the temps overflow, it exposed a potential for a null ptr deref on `ts->mem_base->reg` (`ts->mem_base` is NULL):
```
[#0] 0x7ffff724bce5 → temp_load(s=0x55555556e040, ts=0x5555555703f0, desired_regs=0xffff, allocated_regs=0x30, preferred_regs=0x40)
[#1] 0x7ffff724c0ef → tcg_reg_alloc_mov(s=0x55555556e040, op=0x555555565240)
[#2] 0x7ffff724da14 → tcg_gen_code_mips64el(s=0x55555556e040, tb=0x7fffb6200040)
[#3] 0x7ffff72792a5 → tb_gen_code_mips64el(cpu=0x5555555a46e0, pc=0x1000, cs_base=0x0, flags=0x10000038, cflags=0xff000000)
[#4] 0x7ffff7277f09 → uc_gen_tb(uc=0x555555560770, addr=0x1000, out_tb=0x7fffffffd160)
[#5] 0x7ffff69e7848 → uc_ctl(uc=0x555555560770, control=3355443208)
[#6] 0x55555555b4ea → test_mips_overflow_tcg_temps()
[#7] 0x555555557c2f → acutest_do_run_(test=0x55555555ec00 <acutest_list_+160>, index=0x0)
[#8] 0x555555558060 → acutest_run_(test=0x55555555ec00 <acutest_list_+160>, index=0x0, master_index=0xa)
[#9] 0x5555555594ad → main(argc=0x2, argv=0x7fffffffd3a8)

                 // s=0x00007fffffffcd18  →  [...]  →  0x0000000000000000, ts=0x00007fffffffcd10  →  [...]  →  0x0000004000000200
 → 3065          tcg_out_ld(s, ts->type, reg, ts->mem_base->reg, ts->mem_offset);
 
 gef>  p *ts
$3 = {
  reg = TCG_REG_EAX,
  val_type = TEMP_VAL_MEM,
  base_type = TCG_TYPE_I32,
  type = TCG_TYPE_I32,
  fixed_reg = 0x0,
  indirect_reg = 0x0,
  indirect_base = 0x0,
  mem_coherent = 0x0,
  mem_allocated = 0x0,
  temp_global = 0x0,
  temp_local = 0x1,
  temp_allocated = 0x0,
  val = 0x0,
  mem_base = 0x0,
  mem_offset = 0x0,
  name = 0x0,
  state = 0x0,
  state_ptr = 0x555555670a24
}
```

I traced through the allocation of the temps, and I can see where `val_type=TEMP_VAL_MEM` and `mem_allocated=0`, as well as when `mem_allocated=1`, which happens within `tcg_global_mem_new_internal` or `temp_allocate_frame`. For whatever reason, those don't seem to be getting called in this case, but are being used. Not sure why this error hasn't already happened with some other test, but it was identified upstream and I applied the same fix, which does indeed solve the problem.

@PhilippTakacs I reused the test from #2305, with the proper `OK()` checks. When these two get merged, want to just overwrite the other one with this one while fixing the merge conflict? Or would you prefer to have two different tests?

### DIFF
--- a/qemu/accel/tcg/translate-all.c
+++ b/qemu/accel/tcg/translate-all.c
@@ -1752,6 +1752,11 @@ TranslationBlock *tb_gen_code(CPUState *cpu,
     tcg_ctx->tb_cflags = cflags;
  tb_overflow:
 
+    gen_code_size = sigsetjmp(tcg_ctx->jmp_trans, 0);
+    if (unlikely(gen_code_size != 0)) {
+        goto error_return;
+    }
+
     tcg_func_start(tcg_ctx);
 
     tcg_ctx->cpu = env_cpu(env);
@@ -1774,6 +1779,7 @@ TranslationBlock *tb_gen_code(CPUState *cpu,
 
     gen_code_size = tcg_gen_code(tcg_ctx, tb);
     if (unlikely(gen_code_size < 0)) {
+ error_return:
         switch (gen_code_size) {
         case -1:
             /*
@@ -1785,6 +1791,9 @@ TranslationBlock *tb_gen_code(CPUState *cpu,
              * flush the TBs, allocate a new TB, re-initialize it per
              * above, and re-do the actual code generation.
              */
+            qemu_log_mask(CPU_LOG_TB_OP | CPU_LOG_TB_OP_OPT,
+                          "Restarting code generation for "
+                          "code_gen_buffer overflow\n");
             goto buffer_overflow;
 
         case -2:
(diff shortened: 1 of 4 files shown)


qemu/accel/tcg/translate-all.c
```diff
@@ -1795,11 +1804,20 @@ TranslationBlock *tb_gen_code(CPUState *cpu,
...
              * Try again with half as many insns as we attempted this time.
+             * Use the TB instruction count as calculated within
+             * gen_intermediate_code, as long as it is > 0 (error case), otherwise
+             * use the last calculated max. The icount will never be greater than
+             * the last max_insns, so this will not invoke an infinite loop.
+             * 
              * If a single insn overflows, there's a bug somewhere...
              */
-            max_insns = tb->icount;
+            max_insns = tb->icount > 0 ? tb->icount : max_insns;
```
justdaniel-gh, line 1814: I put this as tb->icount > 0 instead of > 1 due to the "If a single insn overflows, there's a bug somewhere...". In the case of a temps overflow, it should never get here with an icount > 0, as that value is set within `translator_loop` after the overflow would happen. However, there is another area (tcg.c:3851) which returns -2 from `tcg_gen_code`, which could have an icount > 0.

PR #2308, opened 2026-03-10
Fixes for PPC64 CPU emulation

kukrimate:
The default POWER10 1.0 was pointing to the correct place in the array, but for manually set indices we still need to append the same offset in order to set the correct CPU.

Additionally MSR.HV bit is not actually respected for CPUs that should support it.


wtdcode:
LGTM, but you should target dev.

kukrimate:
@wtdcode Changed target and rebased.

PR #2310, opened 2026-03-13
allow using UC_MMIO_READ_TYPE, UC_MMIO_WRITE_TYPE

shuffle2:
fix a script importing:
from unicorn.unicorn import UC_MMIO_READ_TYPE, UC_MMIO_WRITE_TYPE

### DIFF
--- a/bindings/python/unicorn/unicorn_py3/unicorn.py
+++ b/bindings/python/unicorn/unicorn_py3/unicorn.py
@@ -1627,4 +1627,4 @@ UC_MMIO_READ_TYPE = Callable[[Uc, int, int, Any], int]
 UC_MMIO_WRITE_TYPE = Callable[[Uc, int, int, int, Any], None]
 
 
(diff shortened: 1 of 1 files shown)


wtdcode:
I do think somehow the design was intended because these types are just for type hinting, no?

shuffle2:
Right - but the fields they provide type hints for are exposed outside of the module. So currently a user can't properly apply the same hints to their own code (e.g. if they write a wrapper/reimpl of mmio_map)

wtdcode:
Those types could still be imported by names instead of "*", right? I mean if we should either put all such types (there are other type hint types too)  to "__all__" or only functional types like current status, but not just one or two specific types.

shuffle2:
uhm good question - i don't know :') 
I made this PR because some pre-existing code happened to import them by name. If that user code should be changed (to import `*`, if that works), or unicorn should be changed - I guess that's up to you.

PR #2317, opened 2026-04-15
qemu/tcg: fix UC_HOOK_MEM_READ/WRITE on riscv64 host

### LINKED ISSUE #2124: `test_arm64_mem_hook_read_write` fails on `riscv64` and `s390x`
I am experiencing the following `test_arm` and `test_arm64` errors after successful build from current `master` on `riscv64` and `s390x` arch under Alpine Linux 3.21:
```
Test test_arm64_mem_hook_read_write...          [ FAILED ]
  test_arm64.c:649: Check counter[0] == 4 && counter[1] == 4... failed
```
and
```
Test test_arm_mem_hook_read_write...            [ FAILED ]
  test_arm.c:864: Check counter[0] == 2 && counter[1] == 2... failed
```
Additionally `s390x` throws errors for `test_mem`, `test_riscv` and `test_x86`, I'll open a separate issue for that if necessary after this one.

Build logs are here: https://gitlab.alpinelinux.org/strophy/aports/-/pipelines/304520
I'm happy to push builds from specific commits through the Alpine CI system when you tag me, or you can do this yourself with a free account.

Xeonacid:
The RISC-V TCG backend's `tcg_out_tlb_load()` unconditionally used `BNE` to branch to the slow path only on TLB misses. When memory hooks (UC_HOOK_MEM_READ/WRITE) are installed, the slow path must always be taken so hooks are invoked on every memory access.

Use `BEQ x0,x0` when `tcg_uc_has_hookmem()` is true, matching the approach used in the ARM32, AArch64, and x86 backends.

Fixes #2124

Tested on riscv64 host and all tests passed.

### DIFF
--- a/qemu/tcg/riscv/tcg-target.inc.c
+++ b/qemu/tcg/riscv/tcg-target.inc.c
@@ -1013,7 +1013,12 @@ static void tcg_out_tlb_load(TCGContext *s, TCGReg addrl,
 
     /* Compare masked address with the TLB entry. */
     label_ptr[0] = s->code_ptr;
-    tcg_out_opc_branch(s, OPC_BNE, TCG_REG_TMP0, TCG_REG_TMP1, 0);
+    /* Unicorn: when memory hooks are present, always branch to slow path */
+    if (tcg_uc_has_hookmem(s)) {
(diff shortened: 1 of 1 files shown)


wtdcode:
Thanks!

Issue #1766, opened 2023-01-19
The code should return errors instead of aborting/exitting


disconnect3d:
Hey,

There are code paths when Unicorn calls `exit()` or `abort()` instead of returning an error, which is undesirable, especially if you consider the fact that the library is often included/pulled by various tools like [GDB plugins](https://github.com/pwndbg/pwndbg/).

One such path is when Unicorn fails to allocate a RWX memory mapping of 1GB. Instead of a graceful error return, we will get a `"Could not allocate dynamic translator buffer"` error print along with an exit:

https://github.com/unicorn-engine/unicorn/blob/56f3bdedb42d26bee1532cc01baf5eaf44a9aa23/qemu/accel/tcg/translate-all.c#L960-L963

Such a failure can happen for example if the user has low amount of RAM or uses a system with SELinux or custom kernel patches in place (like grsecurity) which disallows for W+X mappings. This results in a poor user experience for library users since the program that uses Unicorn crashes with a bogus error that does not say much. This is [actually a case in Pwndbg](https://github.com/pwndbg/pwndbg/issues/1534#issuecomment-1397081015) currently which motivated me to create this issue.

The original "bail out and exit since allocation failed" may have been good but since the allocation may fail due to different reasons I don't think it is valid anymore.

On a long term, this should be fixed by refactoring the code so that the error can be handled by the library users. Additionally, we could also see if we can allocate with `PROT_WRITE` first and then changing the permissions to `PROT_EXEC`. However, I have seen systems that does not allow for that as well without explicit permissions.

On a short term, we could apply the following improvements:
1) The library could make a check whether RWX memory pages are supported during initialization and return appropriate error if its not
2) The errors displayed could be more explicit, detailing that a RWX memory allocation done by Unicorn failed which can either be OOM or inability to use RWX pages

This way, the library users would be able to handle this case and also would have an easier time tracking down the issue origins.

wtdcode:
Thanks for your issue! I noticed this long time ago but lack time to improve it.

There are also a few tcg memory related issues on my TODO list and I will get them done asap.

vrubleg:
It is still worthy to implement it one day in the future.

vrubleg:
Still worthy to keep it.

disconnect3d:
For what is worth, if we don't have a better solution or don't have time to implement one, we could:
[...]

PR #2322, opened 2026-04-25
uc_context: validate save/restore against content flags and engine

### LINKED ISSUE #2319: `uc_context_alloc` leaves header fields uninitialized; misuse aborts the host
This is a robustness bug in the spirit of #1766 ("The code should return errors instead of aborting/exitting"):
`uc_context_alloc` (uc.c:2241) calls `g_malloc(size)` and only explicitly zeros `fv` (uc.c:2246). `snapshot_level`, `ramblock_freed`, and
`last_block` stay uninitialized. If a caller does `uc_context_alloc + uc_context_restore` without an intervening `uc_context_save`, restore happily writes those garbage values into the engine (uc.c:2567/2578/2579) and returns OK. The next emulation hits `g_assert(ret < cpu->num_ases && ret >= 0)` in `cpu_asidx_from_attrs` (qemu/include/hw/core/cpu.h:421) and aborts the host process.

A Python user sees the interpreter abort with no Python-side exception.

## Repro

```c
/* repro_ctx_restore_no_save.c */
#include <stdio.h>
#include <unicorn/unicorn.h>

int main(void) {
    uc_engine *uc;
    uc_open(UC_ARCH_X86, UC_MODE_64, &uc);

    uc_context *ctx;
    uc_context_alloc(uc, &ctx);
    uc_context_restore(uc, ctx);   /* returns UC_ERR_OK silently */

    uint8_t nop = 0x90;
    uc_mem_write(uc, 0x1000, &nop, 1);
    uc_emu_start(uc, 0x1000, 0x1001, 0, 1);
}
```

Build the same way as a stock ASan debug build:

```
CC=clang CFLAGS="-fsanitize=address,undefined -g -O1" \
  cmake .. -DUNICORN_ARCH=x86 -DBUILD_SHARED_LIBS=OFF -DCMAKE_BUILD_TYPE=Debug
cmake --build . -j
clang -fsanitize=address,undefined -g -O1 -I../include \
  repro_ctx_restore_no_save.c libunicorn.a -lpthread -o repro
./repro
```

Output:

```
calling uc_context_restore WITHOUT prior uc_context_save...
ret = OK (UC_ERR_OK)
Assertion failed: (ret < cpu->num_ases && ret >= 0),
  function cpu_asidx_from_attrs, file cpu.h, line 421.
```

Reproduces both with apple clang and homebrew clang ASan-only builds.

## Suggested fix

Change `g_malloc(size)` to `g_malloc0(size)` at uc.c:2241. One line. The explicit `fv = NULL` at 2246 then becomes redundant but harmless.

If you'd prefer to keep `g_malloc` for the variable-length body for any reason, zero just the header bytes (sizeof `uc_context` minus `data[]`)
before returning.

A more conservative variant is to set a "saved" flag in the header in `uc_context_save` and refuse `uc_context_restore` (return `UC_ERR_ARG`) if the flag is unset. That also closes #1766 for this entry point.

Happy to send the patch + a `tests/unit/test_ctl.c` regression test against `dev`.

### LINKED ISSUE #2320: `uc_context_restore` reads back fields from a struct the user owns; three crash classes follow
The `uc_context` struct returned by `uc_context_alloc` is a normal heap allocation handed to the caller. Its public layout (uc_priv.h:438) contains:
- `size_t context_size`     — drives a memcpy at uc.c:2585
- `void  *fv`               — passed straight to `flatview_copy` at uc.c:2568
- `void  *last_block`       — assigned to `uc->ram_list.last_block` at uc.c:2579

Nothing in `uc_context_restore` checks that any of these were produced by `uc_context_save` on the same engine. If the bytes get corrupted in place, the next `_restore` walks the corrupted fields. Mutating any of the three (which is trivial since the caller owns the allocation) produces a different host crash:

1. `context_size` -> heap-buffer-overflow inside the env memcpy
2. `fv` -> SEGV inside `flatview_copy`
3. `last_block` -> SEGV inside `find_ram_offset_last` on the next RAM op

I want to flag the framing carefully: the docs do not promise that a context blob is portable across processes, and the struct embeds live host pointers (`fv`, `last_block`) that obviously can't survive a serialize/deserialize round trip. So the realistic scenario isn't "attacker ships a malicious snapshot from the network"; it's "the public API hands the user a struct full of live host pointers and trusts those pointers on the way back in." Even an in-process bug that corrupts the heap region holding a `uc_context` turns into a controllable-pointer crash on the next restore.

This is defense-in-depth, not a CVE-class report. Filing here because the fix is small and uniform across all three.

## Repros

All three reproduce on a stock CMake debug build with ASan
(`CC=clang CFLAGS="-fsanitize=address,undefined -g -O1" cmake ...`).

### 1. `context_size` overflow (poc3)

```c
struct fake_ctx_hdr {
  size_t context_size; int mode; int arch; int snapshot_level;
  unsigned char ramblock_freed; void *last_block; void *fv;
};
uc_engine *uc; uc_open(UC_ARCH_X86, UC_MODE_64, &uc);
uc_context *ctx; uc_context_alloc(uc, &ctx); uc_context_save(uc, ctx);
((struct fake_ctx_hdr *)ctx)->context_size = 0x100000;
uc_context_restore(uc, ctx);
```

```
AddressSanitizer: heap-buffer-overflow
READ of size 1048576 ... uc_context_restore uc.c:2585
```

Smaller mismatches produce ASan "memcpy-param-overlap"; sign-bit-set sizes produce "negative-size-param". Same root cause.

### 2. `fv` deref (repro_ctx_fv_hijack)

```c
uc_open(UC_ARCH_X86, UC_MODE_64, &uc);
uc_ctl(uc, UC_CTL_WRITE(UC_CTL_CONTEXT_MODE, 1),
     UC_CTL_CONTEXT_CPU | UC_CTL_CONTEXT_MEMORY);
uc_context_alloc(uc, &ctx); uc_context_save(uc, ctx);
((struct fake_ctx_hdr *)ctx)->fv = (void *)0xdead0000c0debabeULL;
uc_context_restore(uc, ctx);
```

```
AddressSanitizer: SEGV on unknown address 0x...
#0 flatview_copy_x86_64 memory.c:929
#1 uc_context_restore uc.c:2568
```

### 3. `last_block` deref (repro_ctx_lastblock_hijack)

Save first so `fv` stays valid, then tamper only `last_block` and trigger any RAM-allocating op:

```c
uc_open(UC_ARCH_X86, UC_MODE_64, &uc);
uc_ctl(uc, UC_CTL_WRITE(UC_CTL_CONTEXT_MODE, 1),
     UC_CTL_CONTEXT_CPU | UC_CTL_CONTEXT_MEMORY);
uc_context_alloc(uc, &ctx); uc_context_save(uc, ctx);

struct fake_ctx_hdr h; memcpy(&h, ctx, sizeof h);
h.last_block = (void *)0xdeadbeefdeadbeefULL;
memcpy(ctx, &h, sizeof h);

uc_context_restore(uc, ctx);          /* returns OK */
uc_mem_map(uc, 0x10000, 0x1000, UC_PROT_ALL);  /* SEGV here */
```

```
AddressSanitizer: SEGV on unknown address 0x...
#0 find_ram_offset_last exec.c:963
#1 find_ram_offset exec.c:987
#2 ram_block_add exec.c:1057
#3 qemu_ram_alloc_from_ptr_x86_64 exec.c:1136
#4 memory_region_init_ram_x86_64 memory.c:1570
#5 memory_map_x86_64 memory.c:49
#6 uc_mem_map uc.c:1403
```

## Suggested fix

Two complementary changes, the first being the cheapest:

1. **Cheap defense in `uc_context_save`/`_restore`**: write a small header (magic + version + engine arch + engine mode + cpu_context_size) in `_save`, verify it in `_restore`, refuse non-matching blobs with `UC_ERR_ARG`. This catches all three repros above.

2. **Stop putting live pointers in the user-visible struct**: keep `fv` and `last_block` inside `uc_engine`'s side state, keyed by the blob's header sequence number, instead of embedding them in the struct returned to the caller. This removes the class entirely and also gets you closer to a future where contexts could be serialized.

I'd rather sync on the API direction with you before sending a PR for this one. (1) is a one-screen patch I can put up against `dev` quickly if you'd prefer a minimal change. (2) is a bigger refactor.

MarkLee131:
Two failure modes in `uc_context_save` / `uc_context_restore` could abort or corrupt the host process.                                                                                                                                    
                                                                                                                                                                                                                                            
  1. `uc_context_alloc()` returned a heap allocation with only `fv` explicitly zeroed; `snapshot_level` / `ramblock_freed` / `last_block` stayed garbage. A stray `uc_context_restore()` before any `uc_context_save()` propagated those    
  into the engine and tripped the `cpu_asidx_from_attrs()` assertion in `cpu.h` on the next `emu_start()`.                                                                                                                                  
                                                                                                                                                                                                                                            
  2. `uc_context_restore()` trusted `fv` / `last_block` from the user-visible struct without checking they came from a `uc_context_save()` on the same engine. Cross-engine restores dereferenced live host pointers belonging to a         
  different `uc_engine`.                                                                                                                                                                                                                    
                                                                                                                                                                                                                                            
  Changes:                        
                                                                                                                                                                                                                                            
  - `uc_context_alloc()` switches to `g_malloc0()` so untouched header fields read 0.
  - `struct uc_context` grows a `context_content` field set by `uc_context_save()` mirroring `uc->context_content`, and an `engine` pointer set only when `UC_CTL_CONTEXT_MEMORY` is included. `uc_context_restore()` refuses contexts with
  `context_content == 0` (covers #2319), arch/mode mismatching the destination engine, or an engine pointer mismatch on a memory restore (covers #2320). Each restore branch is gated on both sides advertising the matching content bit, so
   a CPU-only context cannot be replayed as a memory snapshot.                                                                                                                                                                            
  - `uc_context_reg_*()` refuses memory-only contexts (`context_content & MEMORY` without `CPU`), since their data buffer holds no CPU state.                                                                                               
                                                                             
  Approach follows the review feedback on this PR and on #2320.                                                                                                                                                                             
                                                                                                                                                                                                                                            
  Regression tests in `tests/unit/test_ctl.c`:
                                                                                                                                                                                                                                            
  - restore without save returns `UC_ERR_ARG`, engine still usable                                                                                                                                                                          
  - cross-engine restore: CPU-only context portable, memory-mode refused
  - `uc_context_reg_read` / `_reg_write` on memory-only snapshot returns `UC_ERR_ARG`                                                                                                                                                       
                                                                             
  Closes part of #1766. Fixes #2319. Fixes #2320. Supersedes #2325.

### DIFF
--- a/uc.c
+++ b/uc.c
@@ -2262,12 +2262,16 @@ uc_err uc_context_alloc(uc_engine *uc, uc_context **context)
 
     UC_INIT(uc);
 
-    *_context = g_malloc(size);
+    // Zero-initialise the whole allocation so a stray uc_context_restore()
+    // before any uc_context_save() sees context_content == 0 and bails out
+    // with UC_ERR_ARG instead of feeding garbage snapshot_level / fv /
+    // last_block into the engine and tripping the cpu_asidx_from_attrs()
+    // assertion in cpu.h. Covers #2319.
+    *_context = g_malloc0(size);
     if (*_context) {
         (*_context)->context_size = size - sizeof(uc_context);
         (*_context)->arch = uc->arch;
         (*_context)->mode = uc->mode;
-        (*_context)->fv = NULL;
         restore_jit_state(uc);
         return UC_ERR_OK;
     } else {
@@ -2302,6 +2306,16 @@ uc_err uc_context_save(uc_engine *uc, uc_context *context)
 {
     UC_INIT(uc);
     uc_err ret = UC_ERR_OK;
+
+    // Tag the context with what we are about to save and which engine
+    // owns the memory state. uc_context_restore() refuses contexts whose
+    // context_content is 0 (never saved, #2319), arch/mode mismatches
+    // the destination engine, or whose engine pointer doesn't match on
+    // a memory restore (#2320). uc_context_reg_*() refuses memory-only
+    // snapshots via context_content as well.
+    context->context_content = uc->context_content;
+    context->engine =
+        (uc->context_content & UC_CTL_CONTEXT_MEMORY) ? uc : NULL;
 
     if (uc->context_content & UC_CTL_CONTEXT_MEMORY) {
         if (!context->fv) {
@@ -2458,9 +2472,28 @@ static context_reg_rw_t find_context_reg_rw(uc_arch arch, uc_mode mode)
     return rw;
 }
 
+// Refuse register r/w on memory-only contexts: their data buffer
+// holds no CPU state, so the call would just shuffle zeros into the
+// caller's value (read) or be lost on the next restore (write).
+// Untouched contexts (context_content == 0) are left alone, so the
+// uc_context_alloc + uc_context_reg_write + uc_context_save pattern
+// keeps working.
+static inline uc_err uc_context_check_cpu_state(const uc_context *ctx)
+{
+    if ((ctx->context_content & UC_CTL_CONTEXT_MEMORY) &&
+        !(ctx->context_content & UC_CTL_CONTEXT_CPU)) {
+        return UC_ERR_ARG;
+    }
+    return UC_ERR_OK;
+}
+
 UNICORN_EXPORT
 uc_err uc_context_reg_write(uc_context *ctx, int regid, const void *value)
 {
+    uc_err err = uc_context_check_cpu_state(ctx);
+    if (err) {
+        return err;
+    }
     int setpc = 0;
     size_t size = (size_t)-1;
     return find_context_reg_rw(ctx->arch, ctx->mode)
@@ -2470,6 +2503,10 @@ uc_err uc_context_reg_write(uc_context *ctx, int regid, const void *value)
 UNICORN_EXPORT
 uc_err uc_context_reg_read(uc_context *ctx, int regid, void *value)
 {
+    uc_err err = uc_context_check_cpu_state(ctx);
+    if (err) {
+        return err;
+    }
     size_t size = (size_t)-1;
     return find_context_reg_rw(ctx->arch, ctx->mode)
         .read(ctx->data, ctx->mode, regid, value, &size);
@@ -2479,6 +2516,10 @@ UNICORN_EXPORT
 uc_err uc_context_reg_write2(uc_context *ctx, int regid, const void *value,
                              size_t *size)
 {
+    uc_err err = uc_context_check_cpu_state(ctx);
+    if (err) {
+        return err;
+    }
     int setpc = 0;
     return find_context_reg_rw(ctx->arch, ctx->mode)
         .write(ctx->data, ctx->mode, regid, value, size, &setpc);
@@ -2488,6 +2529,10 @@ UNICORN_EXPORT
 uc_err uc_context_reg_read2(uc_context *ctx, int regid, void *value,
                             size_t *size)
 {
+    uc_err err = uc_context_check_cpu_state(ctx);
+    if (err) {
+        return err;
+    }
     return find_context_reg_rw(ctx->arch, ctx->mode)
         .read(ctx->data, ctx->mode, regid, value, size);
 }
@@ -2496,6 +2541,10 @@ UNICORN_EXPORT
 uc_err uc_context_reg_write_batch(uc_context *ctx, int const *regs,
                                   void *const *vals, int count)
 {
+    uc_err err = uc_context_check_cpu_state(ctx);
+    if (err) {
+        return err;
+    }
     reg_write_t reg_write = find_context_reg_rw(ctx->arch, ctx->mode).write;
     void *env = ctx->data;
     int mode = ctx->mode;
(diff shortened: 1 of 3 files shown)


PhilippTakacs:
I would prefer to save `uc->context_content` in the context itself. This can also be used as a check in the `uc_context_reg_*` function to prevent register read/write on memory-only snapshots.

MarkLee131:
rewrote on dev. struct uc_context now carries context_content, engine (set only when MEMORY) and arch/mode; restore checks all four, and reg_*() refuses memory-only ctx as you suggested. closed #2325, this PR covers both #2319 and  #2320 now.

uc.c
```diff
@@ -2458,9 +2472,28 @@ static context_reg_rw_t find_context_reg_rw(uc_arch arch, uc_mode mode)
...
+// Refuse register r/w on memory-only contexts: their data buffer
+// holds no CPU state, so the call would just shuffle zeros into the
+// caller's value (read) or be lost on the next restore (write).
+// Untouched contexts (context_content == 0) are left alone, so the
+// uc_context_alloc + uc_context_reg_write + uc_context_save pattern
+// keeps working.
+static inline uc_err uc_context_check_cpu_state(const uc_context *ctx)
+{
+    if ((ctx->context_content & UC_CTL_CONTEXT_MEMORY) &&
+        !(ctx->context_content & UC_CTL_CONTEXT_CPU)) {
```
PhilippTakacs: I don't understand what you want to achieve with this. When we refuse to restore a context which was never saved writing to a register of the context is doing nothing. So why allow this?

MarkLee131: Carve-out was for alloc + reg_write + save, but uc_context_save overwrites the buffer from the engine so the reg_write gets clobbered anyway. Tightened to require UC_CTL_CONTEXT_CPU, symmetric with restore. Test extended.

PR #2325, opened 2026-04-25
uc_context: tag blobs with a tamper-evident magic

### LINKED ISSUE #2320: `uc_context_restore` reads back fields from a struct the user owns; three crash classes follow

MarkLee131:
The struct uc_context that uc_context_alloc() returns is a normal heap allocation handed to the caller. Its public layout exposes context_size, fv and last_block, all of which uc_context_restore() reads back at face value. If anything mutates those bytes between save and restore (which the caller can do trivially because they own the allocation), the restore path produces:

  - a heap-buffer-overflow inside the env memcpy (context_size mutated),
  - a SEGV inside flatview_copy() (fv mutated),
  - or a SEGV inside find_ram_offset_last() on the next RAM op (last_block mutated).

The docs don't promise that a context blob is portable across processes (and the embedded host pointers obviously can't survive that), so the realistic threat model is intra-process: another bug that corrupts the heap region holding the context, or a caller that hands the same context to a different uc_engine. Both turn into a controllable pointer crash on restore.

This commit adds a per-engine context_seed in uc_struct (set at uc_open()) plus a save_magic header field in uc_context. The magic is a hash of context_size / arch / mode / snapshot_level / fv / last_block / ramblock_freed mixed with the engine seed. uc_context_save() recomputes and stores it on success. uc_context_restore() recomputes it from the current (possibly mutated) fields and refuses with UC_ERR_ARG on mismatch.

Catches:
  - restore from a never-saved context (save_magic == 0)
  - cross-engine restore (different engine seed)
  - any mutation of context_size / fv / last_block / arch / mode

uc_context_alloc() now uses g_malloc0() so that the new save_magic field, plus the existing snapshot_level / ramblock_freed / last_block, all start at zero rather than stack/heap garbage.

Adds two regression tests in tests/unit/test_ctl.c covering the four tamper cases above plus the cross-engine case. The non-tampered save -> restore round trip still passes.

Fixes #2320.

### DIFF
--- a/include/uc_priv.h
+++ b/include/uc_priv.h
@@ -422,6 +422,10 @@ struct uc_struct {
                           // workaround to treat the IT block as a whole block.
     bool init_done;       // Whether the initialization is done.
 
+    // Per-engine seed for tagging uc_context blobs in uc_context_save()
+    // and verifying them in uc_context_restore(). Set in uc_open().
+    uint64_t context_seed;
+
     sigjmp_buf jmp_bufs[UC_MAX_NESTED_LEVEL]; // To support nested uc_emu_start
     int nested_level;                         // Current nested_level
 
@@ -451,6 +455,11 @@ struct uc_context {
     bool ramblock_freed;  // wheter there was a some ramblock freed
     RAMBlock *last_block; // The last element of the ramblock list
     FlatView *fv;         // The current flatview of the memory
+    // Tamper-evidence: set by uc_context_save() to a hash of all the
+    // critical fields above mixed with the engine's context_seed.
+    // uc_context_restore() recomputes the hash and refuses contexts
+    // whose fields have been mutated (or that were never saved).
+    uint64_t save_magic;
     char data[0];         // context
 };
 
--- a/tests/unit/test_ctl.c
+++ b/tests/unit/test_ctl.c
@@ -398,6 +398,97 @@ static void test_noexec(void)
     OK(uc_close(uc));
 }
 
+// Regression test for issue #2320. uc_context_restore() used to trust
+// context_size, fv and last_block from the user-visible struct. Mutating
+// any of them produced a heap-overflow / wild-pointer-deref crash. The
+// save_magic header tag now catches all three tamper cases up front.
+//
+// We mirror the public layout of struct uc_context so the test can poke
+// individual fields without including uc_priv.h.
+struct ctl_test_uc_context_layout {
+    size_t context_size;
+    int    mode;
+    int    arch;
+    int    snapshot_level;
+    unsigned char ramblock_freed;
+    void  *last_block;
+    void  *fv;
+    uint64_t save_magic;
+};
+
+static void test_uc_context_restore_rejects_tampering(void)
+{
+    uc_engine *uc;
+    uc_context *ctx;
+
+    OK(uc_open(UC_ARCH_X86, UC_MODE_64, &uc));
+    OK(uc_mem_map(uc, 0x1000, 0x1000, UC_PROT_ALL));
+    OK(uc_context_alloc(uc, &ctx));
+
+    // 1. Restore from a fresh context (save_magic == 0).
+    uc_assert_err(UC_ERR_ARG, uc_context_restore(uc, ctx));
+
+    // 2. Real save then untampered restore is still the happy path.
+    OK(uc_context_save(uc, ctx));
+    OK(uc_context_restore(uc, ctx));
+
+    // 3. Tamper with context_size between save and restore.
+    {
+        struct ctl_test_uc_context_layout *h =
+            (struct ctl_test_uc_context_layout *)ctx;
+        size_t saved = h->context_size;
+        h->context_size = 0x100000;
+        uc_assert_err(UC_ERR_ARG, uc_context_restore(uc, ctx));
+        h->context_size = saved;
+    }
+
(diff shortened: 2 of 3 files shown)


MarkLee131:
folded into #2322, closing in favor of that PR

PR #2327, opened 2026-04-28
Update COMPILE.md with symlink permissions note

kant2002:
Added note about running shell as administrator for symlink permissions.


PhilippTakacs:
can you change the target branch to dev

kant2002:
Change target to dev

wtdcode:
We try to avoid external links in such documents because we wish our tarball to be self-contained. Would you mind updating relevant details in the doc directly?

PR #2343, opened 2026-06-20
Honor emu_start until address across TB-chained runs (#2341)

### LINKED ISSUE #2341: emu_start does not honor `until` after TB chaining across runs (order-dependent fall-through)
### Summary

On a single `Uc` instance, the order in which two adjacent functions are emulated changes the result: running function B and then function A causes A's emulation to **run past its `until` address** into B's code, eventually faulting with `UC_ERR_FETCH_UNMAPPED`. Running A first works correctly.

### Environment

- Unicorn `master` @ `7c5db941` (built from source, x86-only config), Python bindings, `UC_ARCH_X86` / `UC_MODE_64`, Linux x86-64
- Verified the fault is independent of any local patches by building pristine `master` in a clean worktree and loading it via `LIBUNICORN_PATH`.

### Details

`func_a` and `func_b` are laid out back to back, so `func_a_end == func_b_start`. When `func_a` is emulated, its `ret` pops a seeded return address equal to `func_a_end`, which is also the `until` passed to `emu_start`, so emulation should stop there.

Emulating `func_b` first chains its translation blocks. On the subsequent `func_a` run, `func_a`'s terminal block is already chain-linked to the block at `func_a_end` (func_b's entry), so the chained jump bypasses the per-block `until` check. Instead of stopping, control falls through into `func_b`; its epilogue then returns to a poisoned stack slot (`0xDEADBEEF`) and faults.

Calling `uc.ctl_flush_tb()` between the two runs works around it (drops the stale chain), confirming TB chaining as the cause.

### Reproducer

A self-contained regress-style reproducer (3 cases: working order, faulting order, and the flush workaround):

https://github.com/retrocpugeek/unicorn/blob/8f5cb0d4/tests/regress/minimal_tb_chain_repro.py

retrocpugeek:
Fixes #2341.

Whether an address is an exit (`until`) is decided at translation time and baked into the cached TB, but the TB cache is keyed by address only. A block translated by an earlier `uc_emu_start()` — when the address was an ordinary instruction (e.g. the entry of an adjacent function emulated first) — stays cached as a normal block with no halting code. A later run that makes that address an exit chains straight into the stale block and runs past it. That's why the fault was order-dependent.

## Fix

Invalidate cached TBs at the moment the exit set changes, rather than only at run boundaries: when an exit is added (`uc_add_exit`), replaced (`UC_CTL_UC_EXITS`), or the default `until` address changes (`uc_emu_start`). `uc_exit_invalidate()` drops `[addr-1, addr+1)`, covering both the block starting at the exit and the block that halts on reaching it. `qemu/` is untouched.

The existing post-run cleanup in `resume_all_vcpus()` is kept as-is: it's the "exit deactivated" half, and it's also relied on to flush executed blocks for code-patching-after-run (removing it breaks `test_*_code_patching`).

## Testing

- New self-contained regression in `tests/unit/test_x86.c`: emulates the neighbour function first, then the target with `until` at the shared boundary, and asserts it stops there. Fails with `UC_ERR_FETCH_UNMAPPED` without the fix.
- Full unit suite passes (12/12).

### DIFF
--- a/include/uc_priv.h
+++ b/include/uc_priv.h
@@ -454,12 +454,27 @@ struct uc_context {
     char data[0];         // context
 };
 
+// Whether @addr is an exit address is baked into a TB at translation time, but
+// the TB cache is keyed by address only. So whenever an address is added to or
+// removed from the exit set, drop any cached TB whose translation depended on
+// the old state, forcing re-translation. We invalidate [addr - 1, addr + 1):
+// that covers both the block starting at addr (halt block, see translator.c)
+// and the block that halts on reaching it (ends at addr, see the target
+// translators' uc_addr_is_exit() check).
+static inline void uc_exit_invalidate(uc_engine *uc, uint64_t addr)
+{
+    if (addr != 0 && uc->uc_invalidate_tb) {
+        uc->uc_invalidate_tb(uc, addr - 1, 2);
+    }
+}
+
 // We have to support 32bit system so we can't hold uint64_t on void*
 static inline void uc_add_exit(uc_engine *uc, uint64_t addr)
 {
     uint64_t *new_exit = g_malloc(sizeof(uint64_t));
     *new_exit = addr;
     g_tree_insert(uc->ctl_exits, (gpointer)new_exit, (gpointer)1);
+    uc_exit_invalidate(uc, addr);
 }
 
 // This function has to exist since we would like to accept uint32_t or
--- a/tests/unit/test_x86.c
+++ b/tests/unit/test_x86.c
@@ -2201,6 +2201,59 @@ static void test_x86_mem_hooks_pc_guarantee(void)
     OK(uc_close(uc));
 }
 
+// Regression test for an order-dependent TB-chaining fall-through (#2341).
+// func_a (mov eax, 0xfffffffe; ret) and func_b are laid out back to back, so
+// func_a_end == func_b_start. func_a's `ret` returns to func_a_end, which is
+// also the emu_start `until`, so emulation must stop there. The bug: emulating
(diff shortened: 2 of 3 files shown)


retrocpugeek:
### Verification

**64-bit** (Linux x86-64, clean Debug build):
- Full C suite (`ctest`): 12/12 pass
- The added regression test (`tests/regress/minimal_tb_chain_repro.py`): pass

**32-bit** (x86, real `-m32` build): I didn't run the Python regress test here — a 64-bit interpreter can't load a 32-bit `libunicorn.so` — so I verified with a small 32-bit C harness mirroring the repro, linked against a genuine 32-bit build:
- pre-fix lib → reproduces the bug (`UC_ERR_FETCH_UNMAPPED`, `rip=0xdeadbeef`)
- with this fix → `until` honored, stops at the exit address, `rax=0xfffffffe`, no fault

Confirmed via trace that the pre-run invalidation resolves the exit address and invalidates the stale block (`addr=0x1006 → phys=0x6`) without faulting, on both 32-bit and 64-bit.

PhilippTakacs:
> 🤖 Generated with [Claude Code](https://claude.com/claude-code)

Not a fan, your pr is bloated and hard to read

I don't think this is the correct solution, better clear the tb when an exit is added or removed.

Even if this is the correct solution `uc_exit_invalidate_prerun_iter` is duplicated code you can just use `uc_exit_invalidate_iter`.

Put tests in `tests/unit`.

retrocpugeek:
Thanks @PhilippTakacs . Attempted to address your feedback, take another look when you get a chance.

PR #2346, opened 2026-06-23
target/mips: fix spurious EXCP_AdEL on ll/lld under UC_TLB_VIRTUAL

### LINKED ISSUE #2345: MIPS: ll/lld raise spurious EXCP_AdEL above useg under UC_TLB_VIRTUAL
Under `UC_TLB_VIRTUAL`, MIPS Load-Linked (`ll`/`lld`) raise a spurious Address Error (`EXCP_AdEL`, intno 12) for addresses at/above the useg limit (`>= 0x80000000`), even though an ordinary load (`ld`/`lw`) at the same address goes through the `UC_HOOK_TLB_FILL` hook and succeeds. This breaks emulating MIPS64 code that takes a lock above 2 GB (e.g. glibc).

Reproduced on 2.1.3 and current `dev`; MIPS64 BE+LE, independent of CPU model.

```python
import unicorn
from unicorn import (Uc, UC_ARCH_MIPS, UC_MODE_MIPS64, UC_MODE_BIG_ENDIAN,
                     UC_HOOK_TLB_FILL, UC_TLB_VIRTUAL, UC_PROT_ALL, UcError)
from unicorn.mips_const import UC_MIPS_REG_S0, UC_MIPS_REG_V1

ADDR = 0x120000000  # above the MIPS useg limit; requires the virtual TLB

def identity_tlb(uc, vaddr, access, entry, user_data):
    entry.paddr = vaddr & ~0xfff
    entry.perms = UC_PROT_ALL
    return True

for opcode, name in ((0xde030000, "ld $v1, 0($s0)"), (0xc2030000, "ll $v1, 0($s0)")):
    uc = Uc(UC_ARCH_MIPS, UC_MODE_MIPS64 | UC_MODE_BIG_ENDIAN)
    uc.ctl_set_tlb_mode(UC_TLB_VIRTUAL)
    uc.hook_add(UC_HOOK_TLB_FILL, identity_tlb)
    uc.mem_map(0x10000000, 0x1000)            # code page
    uc.mem_map(ADDR, 0x1000)                  # data page (the lock word)
    uc.mem_write(0x10000000, opcode.to_bytes(4, "big"))
    uc.reg_write(UC_MIPS_REG_S0, ADDR)
    try:
        uc.emu_start(0x10000000, 0x10000004, 0, 1)
        print("  %-16s OK" % name)
    except UcError as e:
        print("  %-16s ERROR: %s" % (name, e))
```

Output:

```
  ld $v1, 0($s0)   OK
  ll $v1, 0($s0)   ERROR: Unhandled CPU exception (UC_ERR_EXCEPTION)
```

**Root cause:** the `ll`/`lld` helpers (`HELPER_LD_ATOMIC` in `qemu/target/mips/op_helper.c`) perform the load via the soft-TLB (which honors the virtual TLB) but additionally call `do_translate_address()` to populate `CP0_LLAddr`. That walks the MIPS segments via `get_physical_address()` and returns `TLBRET_BADADDR` for an address outside useg (UX unset), raising `EXCP_AdEL` and bypassing the virtual TLB. `sc`/`scd` are unaffected (they use `tcg_gen_atomic_cmpxchg_tl`, which never calls `do_translate_address`).

Fix + regression test ready; PR to follow.

retrocpugeek:
Fixes #2345.

Under `UC_TLB_VIRTUAL`, MIPS `ll`/`lld` raised a spurious `EXCP_AdEL` for addresses above useg (`>= 0x80000000`) because `HELPER_LD_ATOMIC` (`qemu/target/mips/op_helper.c`) called `do_translate_address()` to populate `CP0_LLAddr`, which segment-checks the address and rejects anything outside useg. The load itself already goes through the soft-TLB, so in virtual-TLB mode this records the virtual address into `CP0_LLAddr` directly instead of segment-translating it.

The alignment check is preserved (a misaligned `ll` still raises AdEL); `sc`/`scd` are unaffected (they use `tcg_gen_atomic_cmpxchg_tl`, which never calls `do_translate_address`).

Adds a regression test in `tests/unit/test_mips.c` (high-address `ll` reads correctly; misaligned `ll` still faults).

---

Downstream tracking: qilingframework/qiling#1645 (Qiling MIPS64 real-ELF/glibc support gates its CI tests on this fix).

### DIFF
--- a/qemu/target/mips/op_helper.c
+++ b/qemu/target/mips/op_helper.c
@@ -312,7 +312,22 @@ target_ulong helper_##name(CPUMIPSState *env, target_ulong arg, int mem_idx)  \
         }                                                                     \
         do_raise_exception(env, EXCP_AdEL, GETPC());                          \
     }                                                                         \
-    env->CP0_LLAddr = do_translate_address(env, arg, 0, GETPC());             \
+    /*                                                                        \
+     * Unicorn: under UC_TLB_VIRTUAL the guest's MIPS segment layout does not \
+     * apply - the soft-TLB fill hook provides an identity/virtual mapping,   \
+     * so any address may be valid (e.g. MIPS64 useg above the 2GB limit).    \
+     * cpu_mips_translate_address() walks the MIPS segments and would raise a \
+     * spurious EXCP_AdEL for such addresses, even though the actual load      \
+     * below goes through the soft-TLB and succeeds. Skip the segment-based   \
+     * translation in virtual-TLB mode and record the (virtual) address; the  \
+     * real access still validates permissions via the TLB-fill hook.         \
+     */                                                                       \
+    if (CPU_GET_CLASS(env_cpu(env))->tlb_fill !=                              \
+        CPU_GET_CLASS(env_cpu(env))->tlb_fill_cpu) {                          \
+        env->CP0_LLAddr = arg;                                                \
+    } else {                                                                  \
+        env->CP0_LLAddr = do_translate_address(env, arg, 0, GETPC());         \
+    }                                                                         \
     env->lladdr = arg;                                                        \
     env->llval = do_cast cpu_##insn##_mmuidx_ra(env, arg, mem_idx, GETPC());  \
     return env->llval;                                                        \
--- a/qemu/target/mips/op_helper.c
+++ b/qemu/target/mips/op_helper.c
@@ -314,17 +314,23 @@ target_ulong helper_##name(CPUMIPSState *env, target_ulong arg, int mem_idx)  \
     }                                                                         \
     /*                                                                        \
      * Unicorn: under UC_TLB_VIRTUAL the guest's MIPS segment layout does not \
-     * apply - the soft-TLB fill hook provides an identity/virtual mapping,   \
-     * so any address may be valid (e.g. MIPS64 useg above the 2GB limit).    \
+     * apply - the soft-TLB fill hook supplies the mapping and may translate  \
+     * a page to any physical address (it is not necessarily identity).       \
      * cpu_mips_translate_address() walks the MIPS segments and would raise a \
-     * spurious EXCP_AdEL for such addresses, even though the actual load      \
-     * below goes through the soft-TLB and succeeds. Skip the segment-based   \
-     * translation in virtual-TLB mode and record the (virtual) address; the  \
-     * real access still validates permissions via the TLB-fill hook.         \
+     * spurious EXCP_AdEL for addresses outside useg (e.g. a MIPS64 address    \
+     * above the 2GB limit), even though the load below goes through the      \
+     * soft-TLB and succeeds. So in virtual-TLB mode skip the segment walk    \
+     * and recover the hook-mapped physical address from the soft-TLB, which  \
+     * is what CP0_LLAddr is architecturally meant to hold.                   \
      */                                                                       \
     if (CPU_GET_CLASS(env_cpu(env))->tlb_fill !=                              \
         CPU_GET_CLASS(env_cpu(env))->tlb_fill_cpu) {                          \
-        env->CP0_LLAddr = arg;                                                \
+        target_ulong paddr;                                                   \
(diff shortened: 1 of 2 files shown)


PhilippTakacs:
I had only a quick look, but you missed that the virtual tlb option also allows hooks to change the mapping.

retrocpugeek:
Thanks @PhilippTakacs
Hope the updated test and fix covers the case you identified.

wtdcode:
Looks good to me. @PhilippTakacs do you also approve this?

PhilippTakacs:
there is still an issue, because with the mips mmu the code raises an exeption (see `cpu_mips_translate_address` and exits the translation loop). this change just put the 1:1 mapping back in, but the vtlb mmu allows to have a translation failing. I would guess using `CPU_GET_CLASS(env_cpu(env))->tlb_fill()` direct and setting probe to false would be the correct implementation, but i'm not familiar with mips and have the next two weeks not the time to look at this.

retrocpugeek:
Thanks @PhilippTakacs, good catch — you're right that the previous version put a 1:1 mapping back in and swallowed a translation the vtlb hook is allowed to decline.

I tried your suggestion of calling `CPUClass::tlb_fill()` directly with `probe=false` first, but it doesn't compose with unicorn's soft-TLB fill: `unicorn_fill_tlb()` calls `cpu_restore_state()` unconditionally at its top, which is only safe when you're about to leave the CPU loop on a fault. Driving it inline from the `ll` helper and then *continuing* execution corrupts CPU state — it broke real MIPS64 workloads (the guest ended up taking an unexpected exception mid-instruction).

So instead I get the same faulting semantics through the supported path: under `UC_TLB_VIRTUAL`, do the soft-TLB **load first** (it goes through the normal softmmu slow path and raises the proper MMU fault if the hook declines the page), then recover the hook-mapped physical address from the now-populated soft-TLB for `CP0_LLAddr`. A genuinely-declined page now faults instead of recording a bogus 1:1 address, and the non-virtual-TLB path is unchanged (`cpu_mips_translate_address` before the load, as before).

Added a regression test asserting that `ll` on a page the hook declines raises `UC_ERR_MMU_READ`.

Pushed as 5573568a.

wtdcode:
> Thanks @PhilippTakacs, good catch — you're right that the previous version put a 1:1 mapping back in and swallowed a translation the vtlb hook is allowed to decline.
> 
> I tried your suggestion of calling `CPUClass::tlb_fill()` directly with `probe=false` first, but it doesn't compose with unicorn's soft-TLB fill: `unicorn_fill_tlb()` calls `cpu_restore_state()` unconditionally at its top, which is only safe when you're about to leave the CPU loop on a fault. Driving it inline from the `ll` helper and then _continuing_ execution corrupts CPU state — it broke real MIPS64 workloads (the guest ended up taking an unexpected exception mid-instruction).
> 
> So instead I get the same faulting semantics through the supported path: under `UC_TLB_VIRTUAL`, do the soft-TLB **load first** (it goes through the normal softmmu slow path and raises the proper MMU fault if the hook declines the page), then recover the hook-mapped physical address from the now-populated soft-TLB for `CP0_LLAddr`. A genuinely-declined page now faults instead of recording a bogus 1:1 address, and the non-virtual-TLB path is unchanged (`cpu_mips_translate_address` before the load, as before).
> 
> Added a regression test asserting that `ll` on a page the hook declines raises `UC_ERR_MMU_READ`.
> 
> Pushed as [5573568](https://github.com/unicorn-engine/unicorn/commit/5573568af06e7398433081447c463ccf936fd44e).

Hi @retrocpugeek, I appreciate your work on Unicorn, and honestly I'm not against using AI tools for coding. However, I would kindly request that you at least review all outputs and phrase the points yourself for high-quality conversations and communication. It is a bit offensive to always paste plain, unmodified AI-generated content. Unicorn, as a widely used security tool, focuses on code quality, and we wish we could get the intention of the pull requests right in the first place.

Thanks again for your contributions, and please consider making our communication more efficient.

retrocpugeek:
Fair call, thanks again @PhilippTakacs and @wtdcode
This issue is in a pretty niche part of the framework, and your expertise is very apparent.
I'll invest more human time in the future.

Issue #2342, opened 2026-06-16
vtlb hook and pc sync


PhilippTakacs:
In `unicorn_fill_tlb` there is a call to `cpu_restore_state`. This call ensures that the pc is synced before the vtlb callback. The problem is that this causes more or less a recalculation of the current TB. We spend quite some time to have our VTLB hook (and other hooks) fast, so we only depend on already known mappings stored via the data pointer. When the mapping is not known (or need some special handling) at the moment we return false and handle the error outside the emulation (this is the reason for #2270). A quick benchmark showed a huge speedup when removing this call.

@tbodt you mentioned that your hook depend on a correct pc. This was before it was possible to get the faulting address and exact reason when the hook failed. Would it possible for you to adjust your hook to return false when the correct pc is required and handle this case outside of the emulation?

tbodt:
My hook is the linux kernel, I combined a custom linux kernel arch and unicorn to make a sort of UML-type thing. I used the TLB fill hook to trigger page fault handling. Maybe there is a better way? I can't remember anymore why I didn't use UC_HOOK_*_UNMAPPED.

PhilippTakacs:
Thanks for the explanation, we do something similar.

The TLB hook is called on a tlb miss not on a page fault. So it might even get called twice for the same page without any direct change on the layout.

> Maybe there is a better way?

You could do the same as we do: add some page table implementation to your custom arch which stores the memory mappings in the user_data of the vtlb hook. When there is no mapping in the user_data return `false` from the hook. Then check the return code of `uc_emu_start` and call the page fault handler when `UC_ERR_MMU_*` is returned.

The `UC_HOOK_*_UNMAPPED` hooks might also work, but have some limits (i.e. can't have a page in one process mapped read and in an other read+write).

PhilippTakacs:
@wtdcode can you share your opinion on this before the next release #2356.

wtdcode:
I took some time to read the code. If I understand correctly, the only break change is that:

- inside the tlb hook
- reads the stale pc
- returns `true` (if it returns `false`, this breaks the translation loop and we sync pc in the end)

This makes sense because reading pc inside tlb hooks sounds useless and `UC_HOOK_*_UNMAPPED` seems the best alternative. If my understanding is correct, I agree removing that sync but explicity documentation is required, including pointers to `UC_HOOK_*_UNMAPPED`.

retrocpugeek:
[...]

PR #2347, opened 2026-06-23
target/mips: fix spurious EXCP_RI under UC_TLB_VIRTUAL after a delay-slot TLB fill

### LINKED ISSUE #2272: MIPS emulation goes into “Branch in delay / forbidden slot” scenario when virtual tlb is enabled

retrocpugeek:
Fixes #2272.

### Symptom

Under `UC_TLB_VIRTUAL`, a memory access in a branch **delay slot** raises a spurious
`EXCP_RI` ("Reserved Instruction / branch in delay-forbidden slot") on the branch
*target*, aborting emulation. The MIPS repro is glibc's `__libc_setup_tls` PT_TLS
program-header scan, whose loop body ends in a branch whose delay-slot load forces a
TLB fill.

### Root cause

`unicorn_fill_tlb()` called `cpu_restore_state(cs, retaddr, false)` unconditionally.
`cpu_restore_state()` → `restore_state_to_opc()` rolls `env` back to the instruction
boundary of the faulting access, and on MIPS that re-applies the per-insn *start*
hflags. For a delay-slot instruction those include the branch hflags
(`MIPS_HFLAG_B` / `MIPS_HFLAG_BDS32` / …), which are only ever meant to exist during
translation, never in the live runtime hflags.

That rollback is fine on the paths `cpu_restore_state()` was designed for — the CPU is
about to *leave* the run loop (raise an exception, or read the exact PC then exit), so
the transient hflags are re-derived on the next entry and never observed. But on a
**successful** TLB fill the faulting access is resumed *in place* inside the same TB.
Nothing clears the leaked bits, so the next TB — the branch target — is translated as
if it sat in a delay slot, and MIPS raises `EXCP_RI`.

### Fix

Remove the unconditional `cpu_restore_state()` from the fill path. The exception path
still restores via `cpu_loop_exit_restore()` (in `raise_mmu_exception()`), mirroring the
CPUs' own `*_cpu_tlb_fill` handlers, which only restore on the exception path. The
`UC_HOOK_TLB_FILL` callback is invoked with the faulting page/type and the entry to
fill; it does not consume rolled-back register state.

This also removes a per-fill `cpu_restore_state()` from the hot path, which is exactly
the change requested for performance reasons in #2342.

Note on the earlier wording: I originally justified this as "the TLB_FILL callback only
needs the faulting address." That over-generalized — per the FAQ and #2342/#2049, hooks
are normally expected to see a synced PC. The accurate framing is the one above: the
restore is *unconditional on a resume-in-place path*, where it corrupts live `env` state
regardless of whether any hook is registered.

Note on #2273: it also targets #2272 but changes only the no-hook fallback path's
permissions; it does not fix the case where a `UC_HOOK_TLB_FILL` hook is registered
(verified — the RI still fires). This fix covers both paths and is orthogonal to #2273.

### Answer to "can a `UC_HOOK_MEM_*_UNMAPPED` hook trigger a similar bug?"

Yes — and it is not specific to the virtual TLB. The MEM hook paths in `cputlb.c`
(`UC_HOOK_MEM_READ`/`WRITE` and the `*_UNMAPPED` variants) also call
`cpu_restore_state(uc->cpu, retaddr, false)` before invoking the hook, and after a
`*_UNMAPPED` hook maps the page and returns `true` the faulting access likewise resumes
*in place*. So the same delay-slot hflag leak occurs.

I reproduced it in the **ordinary soft-MMU (no `UC_TLB_VIRTUAL`)**: a delay-slot load
that faults, serviced by a `UC_HOOK_MEM_READ_UNMAPPED` callback that maps the page and
returns `true`, resumes with the leaked hflags and raises `EXCP_RI` (intno 20 via
`UC_HOOK_INTR`; `UC_ERR_EXCEPTION` otherwise). Pre-mapping the page so no hook fires
makes the same program complete correctly — isolating the cause to the hook-triggered
`cpu_restore_state()`. Standalone failing repro (kept out of this PR): #2352.

The important difference: those MEM paths call `cpu_restore_state()` *deliberately*, to
honour the "synced PC in hooks" contract, and are already gated (`!synced &&
!skip_sync_pc_on_exit`). They can't simply drop the call the way the vtlb fill can.
The proper fix for that class is lower-level — clearing the transient branch hflags
after a resume-in-place restore (in `restore_state_to_opc()` / its MIPS callers) so the
contract and correctness both hold. That's a broader change than this regression fix, so
I've kept it out of this PR; happy to open a follow-up if you agree with the direction.

### Test

Adds `test_mips_virtual_tlb_branch_in_delay_slot_2272` in `tests/unit/test_mips.c`,
exercising the `UC_HOOK_TLB_FILL` path (how a soft-MMU front-end such as Qiling drives
the virtual TLB).

---

Downstream tracking: qilingframework/qiling#1645 (Qiling MIPS64 real-ELF/glibc support gates its CI tests on this fix).

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -42,7 +42,7 @@ ldr x0, [x1] <--- exception here and PC sync-ed here
 <details>
   <summary>For version >= 2.1.4 </summary>
 
-We should always have valid PC all the time. Please report an issue for your case.
+We should always have valid PC all the time, with one documented exception: inside a `UC_HOOK_TLB_FILL` callback the CPU registers (including PC) are *not* rolled back to the faulting instruction. See [the `UC_TLB_VIRTUAL` question below](#is-there-anyway-to-disable-softmmu-to-speed-up-execution) for details. Otherwise, please report an issue for your case.
 
 </details>
 
@@ -134,6 +134,10 @@ Starting from 2.0.2, Unicorn will emulate the MMU depending on the emulated arch
 Therefore, if you still prefer the previous `paddr = vaddr` simple mapping, we have a simple experimental MMU implementation that can be switched on by: `uc_ctl_tlb_mode(uc, UC_TLB_VIRTUAL)`. With this mode, you could also add a `UC_HOOK_TLB_FILL` hook to manage the TLB. When a virtual address is not cached, the hook will be called. Besides, users are allowed to flush the tlb with `uc_ctl_flush_tlb`.
 
 In theory, `UC_TLB_VIRTUAL` will achieve better performance as it skips all MMU details, though not benchmarked.
+
+The `UC_HOOK_TLB_FILL` callback is given the faulting *virtual* address and the access type, and returns the mapping for that page (physical address and permissions), or `false` to reject it. Decide the mapping from the address argument passed to the hook — **not** from the CPU registers. Unlike most hooks, the registers (including PC) are not guaranteed to be rolled back to the faulting instruction while the callback runs: on a successful fill the faulting access is resumed in place, so rolling the state back here would leak into the resumed execution (on MIPS, for example, it re-applies branch-delay-slot flags and raises a spurious Reserved Instruction exception). Keeping this off the register-restore path also avoids per-fill overhead.
+
+When the hook returns `false` — or when no hook is installed and a virtual address cannot be translated — `uc_emu_start` stops and returns `UC_ERR_MMU_READ`, `UC_ERR_MMU_WRITE`, or `UC_ERR_MMU_FETCH` depending on the access type. You can then read the faulting virtual address with `uc_ctl_get_invalid_addr(uc, &addr)`.
 
 ## Something is wrong - I would like to dig deeper
 
--- a/qemu/softmmu/unicorn_vtlb.c
+++ b/qemu/softmmu/unicorn_vtlb.c
@@ -67,7 +67,26 @@ bool unicorn_fill_tlb(CPUState *cs, vaddr address, int size,
     struct hook *hook;
     HOOK_FOREACH_VAR_DECLARE;
 
-    cpu_restore_state(cs, retaddr, false);
+    /*
+     * Unicorn: Do NOT unconditionally restore CPU state here.
+     *
+     * cpu_restore_state() rolls env back to the instruction boundary of the
+     * faulting access (it calls the target restore_state_to_opc()). On a
+     * successful fill the faulting access is then resumed in place inside the
+     * same TB, so any env mutation done here leaks into the continued
+     * execution. On MIPS this is fatal: restore_state_to_opc() re-applies the
+     * branch-delay hflags (MIPS_HFLAG_B / MIPS_HFLAG_BDS32 etc.) that were
+     * saved for a delay-slot instruction. Those bits are normally never present
+     * in the runtime hflags (they live only in the per-insn start data), so
+     * nothing ever clears them again, and the next TB (the branch target) is
+     * then translated as if it were sitting in a delay slot, raising a spurious
+     * EXCP_RI ("branch in delay / forbidden slot").
+     *
+     * The CPU own tlb_fill handlers (e.g. mips_cpu_tlb_fill) only restore
+     * state on the exception path. We mirror that: the exception path below
+     * restores via cpu_loop_exit_restore(), and the TLB_FILL hooks only need
+     * the faulting address (passed explicitly), not rolled-back register state.
+     */
 
     HOOK_FOREACH(uc, hook, UC_HOOK_TLB_FILL) {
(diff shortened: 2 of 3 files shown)


PhilippTakacs:
Thanks for the detailed explanation of the bug.

> The UC_HOOK_TLB_FILL callbacks only need the faulting address, which is passed explicitly

This is not correct, see #2342 and #2049. The FAQ also says following:

> We should always have valid PC all the time. Please report an issue for your case.

So general it's assumed that the PC is synced (and all other registers) when a hook is called. I see that the vtlb hook is a candidate for an exception and would also like to remove the `cpu_restore_state()` call (for performance reasons).

By the way: Is it possible to trigger similar bug with an `UC_HOOK_MEM_*_UNMAPPED` hook?

retrocpugeek:
You're right — thanks, I've corrected the PR description. "Hooks only need the faulting address" over-generalized the contract; the accurate framing is that the restore here is *unconditional on a resume-in-place path*, so it corrupts live `env` regardless of whether a hook is registered. And since you'd like this call gone for perf (#2342) anyway, this fix and that goal line up.

On the `UC_HOOK_MEM_*_UNMAPPED` question: yes, the same bug is reachable — and it's not vtlb-specific. The MEM hook paths in `cputlb.c` also `cpu_restore_state()` before the callback and then resume the access in place after a `*_UNMAPPED` hook maps the page. I reproduced it in the plain soft-MMU (no `UC_TLB_VIRTUAL`): a delay-slot `lw` that faults, serviced by a `UC_HOOK_MEM_READ_UNMAPPED` hook that maps the page and returns `true`, raises `EXCP_RI` (intno 20). Pre-mapping the page so no hook fires makes the same program complete correctly. Standalone failing repro on a separate branch so it doesn't muddy this PR: #2352.

The real root cause is that MIPS `restore_state_to_opc()` re-applies the delay-slot branch hflags into the live `env`, which is only safe when the CPU is about to leave the run loop. The vtlb fill can simply not restore (this PR); the MEM paths restore on purpose for the synced-PC contract, so they'd need the transient branch hflags cleared after a resume-in-place restore instead. Happy to open a follow-up PR for that if you agree with the direction.

PhilippTakacs:
> And since you'd like this call gone for perf (https://github.com/unicorn-engine/unicorn/issues/2342) anyway, this fix and that goal line up

Yes, but I also want a correct documentation and good justification for such exceptions.

Can you add some documentation to `docs/FAQ.md` mentioning the exception with some hints to `UC_ERR_MMU_*` and `uc_ctl_get_invalid_addr`?

retrocpugeek:
Added to `docs/FAQ.md` in 2e985689: documented the `UC_HOOK_TLB_FILL` path as the one exception to the "valid PC all the time" guarantee (with the resume-in-place justification and the #2342 perf angle), and pointed at `UC_ERR_MMU_{READ,WRITE,FETCH}` plus `uc_ctl_get_invalid_addr()` for retrieving the faulting address. Let me know if you'd like more detail or a different framing.

PR #2353, opened 2026-07-04
target/i386: Decode group 3 /1 instructions as TEST

unknown-1-0:
According to table A-6 in Volume 3 of AMD64 Architecture Programmer's Manual, group 3 instructions that have ModRM.reg field value of either 0 or 1 are decoded as TEST instruction.

Although Volume 2 of Intel 64 and IA-32 Architectures Software Developer's Manual leaves the cell for ModRM.reg == 1 in group 3 in table A-6 blank, group 3 instructions with ModRM.reg == 1 seem to behave like TEST on Intel CPUs.

Currently, Unicorn does not recognize group 3 instruction with ModRM.reg == 1. As a result, it raises an #UD exception. This behavior does not match that of a real hardware.

Fix this issue by adding a missing `case 1:` to the group 3 decoder.

Upstream QEMU issue: https://gitlab.com/qemu-project/qemu/-/work_items/3580

### DIFF
--- a/qemu/target/i386/translate.c
+++ b/qemu/target/i386/translate.c
@@ -5277,6 +5277,7 @@ static target_ulong disas_insn(DisasContext *s, CPUState *cpu)
 
         switch(op) {
         case 0: /* test */
+        case 1:
             val = insn_get(env, s, ot);
             tcg_gen_movi_tl(tcg_ctx, s->T1, val);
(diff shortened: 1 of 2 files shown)


unknown-1-0:
Upstream QEMU patch: https://lore.kernel.org/qemu-devel/20260621032524.1138213-1-apolivodaa433@gmail.com/T/#u

PR #2354, opened 2026-07-04
target/i386: Decode 0x8f as POP only if ModRM.reg == 0

unknown-1-0:
According to both table A-6 in Volume 2 of Intel 64 and IA-32 Architectures Software Developer's Manual and table A-6 in Volume 3 of AMD64 Architecture Programmer's Manual, opcode 0x8f (group 1a) is decoded as POP only if ModRM.reg == 0.

Currently, Unicorn does not perform this check, which makes it happily execute every group 1a instruction as POP, no matter if it's actually POP or not.

Fix this bug by adding a missing check.

### DIFF
--- a/qemu/target/i386/translate.c
+++ b/qemu/target/i386/translate.c
@@ -5881,8 +5881,14 @@ static target_ulong disas_insn(DisasContext *s, CPUState *cpu)
         tcg_gen_movi_tl(tcg_ctx, s->T0, val);
         gen_push_v(s, s->T0);
         break;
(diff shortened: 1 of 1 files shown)


wtdcode:
Hi, thanks for your help. Did you port this commit from qemu? if so, what's their commit hash?

unknown-1-0:
No, I didn't port it from QEMU. I wrote this commit based on both Intel and AMD manuals.

wtdcode:
Took some time to validate this. Thanks!

Issue #2356, opened 2026-07-04
Next release


wtdcode:
Hi everyone,

Finally, I got some time for the next release and this will contain several key fixes to resolve segment fault and a few security issues.

This would probably be `2.1.5` but if I stall enough changes, I will make `2.2.0` instead. Since the backlog is very long, please ping me if you do want your changes to be included.

PR #2358, opened 2026-07-05
target/i386: Fix LOCK prefix handling with BTx instructions

unknown-1-0:
According to the Intel 64 and IA-32 Software Developer's Manual, LOCK BT and forms of LOCK BTS/BTR/BTC where the destination is not a memory operand raise an invalid opcode (#UD) exception when executed.

Currently, Unicorn does not check for these conditions when decoding BTx instructions. This behavior not only does not match the hardware's behavior, but also allows Unicorn to crash, since in register-only forms, the `s->A0` temp can be used uninitialized (allocated, but not written to), leading to a `tcg fatal error` later because that temp is marked as "dead" and is later used in `temp_load()`, which then calls `tcg_abort();`.

Fix this bug by adding the missing checks.

### DIFF
--- a/qemu/target/i386/translate.c
+++ b/qemu/target/i386/translate.c
@@ -7489,6 +7489,9 @@ static target_ulong disas_insn(DisasContext *s, CPUState *cpu)
         op = (modrm >> 3) & 7;
         mod = (modrm >> 6) & 3;
         rm = (modrm & 7) | REX_B(s);
+        /* LOCK BT and register-only forms of LOCK BTS/BTR/BTC are illegal */
+        if ((s->prefix & PREFIX_LOCK) && (mod == 3 || op == 4))
+            goto illegal_op;
         if (mod != 3) {
             s->rip_offset = 1;
             gen_lea_modrm(env, s, modrm);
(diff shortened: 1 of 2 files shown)


qemu/target/i386/translate.c
```diff
@@ -7501,11 +7507,6 @@ static target_ulong disas_insn(DisasContext *s, CPUState *cpu)
         tcg_gen_shl_tl(tcg_ctx, s->tmp0, s->tmp0, s->T1);
         if (s->prefix & PREFIX_LOCK) {
             switch (op) {
-            case 0: /* bt */
```
unknown-1-0, line 7546: Should not be reachable anymore, so `case 0:` is removed. `g_assert_not_reached();` may be used instead, if you wish.

unknown-1-0:
@wtdcode, could you please share your opinion on this fix before the next release (#2356)? Thank you!

wtdcode:
> @wtdcode, could you please share your opinion on this fix before the next release (#2356)? Thank you!

Spare me some time to review, or if possible, could you add a unit test for this? It would be better that you also add one for #2354 though already merged.

unknown-1-0:
Working on it

unknown-1-0:
Working on getting rid of `Merge branch 'dev' into i386-bt-fix-tcg-fatal-error`

wtdcode:
> Working on getting rid of `Merge branch 'dev' into i386-bt-fix-tcg-fatal-error`

#2358 seems not included? Can you get it into this PR (or a separate one?).

unknown-1-0:
Sorry, did you mean #2354? If so, will add tests for it in a separate PR

wtdcode:
> Sorry, did you mean #2354? If so, will add tests for it in a separate PR

Oh yes

unknown-1-0:
I see there's a merge conflict in the test_x86.c file. If changes on my end are necessary, please let me know when you are ready to merge this PR, and I'll resolve it as soon as I can to keep the commit history clean. :)

wtdcode (approved):

wtdcode:
I confirm that the fix is correct, please fix conflicts.

unknown-1-0:
Done

PR #2360, opened 2026-07-06
Fix __int128_t redefinition in int128.h !CONFIG_INT128 fallback on Clang/GCC

robertream:
## Problem

Building QEMU's `int128.h` fails on Clang/GCC whenever the build reaches the
`!CONFIG_INT128` fallback:

```
qemu/include/qemu/int128.h:150:16: error: typedef redefinition with different types
    ('Int128' (aka 'struct Int128') vs '__int128')
typedef Int128 __int128_t;
```

The fallback aliases the **reserved builtin name** `__int128_t` to `struct Int128`
so fallback code can name the type. That is only valid on compilers that do *not*
provide `__int128_t` as a builtin. On GCC/Clang the builtin always exists (when
128-bit ints are supported), so the alias is a redefinition error.

This bit two real setups:
- **Apple clang on arm64** (macOS), and
- the Rust **`unicorn-engine-sys`** crate's vendored build (its cmake path ends up
  without `CONFIG_INT128`, so it compiles the fallback — and fails on any modern
  clang/gcc).

#2251 added the current guard `#if !(defined(_MSC_VER) && defined(__clang__))`,
but that only covers **clang-cl**; regular Clang/GCC still hit the redefinition.

## Fix

Gate the alias on `!defined(__SIZEOF_INT128__)`. `__SIZEOF_INT128__` is the standard
macro that is defined **iff** the compiler provides the 128-bit builtin, so:
- GCC/Clang (incl. clang-cl): builtin present → alias skipped (uses the builtin) ✅
- MSVC: no builtin → alias emitted (as before) ✅

This subsumes #2251 and fixes the general case.

## Validation

- `tests/regress/int128_redefinition.c` — a self-contained reproduction of the
  fallback construct. It **fails to compile with the old guard** on any
  `__int128`-capable compiler and compiles cleanly with the fix (verified on Apple
  clang / arm64, producing the exact error above pre-fix).
- `.github/workflows/int128-regression.yml` — compiles + runs that test on
  `macos-14` (Apple Silicon, arm64 clang), `macos-13` (Intel) and Linux.

**Why a new job:** existing CI never exercises this fallback on arm64 clang — the
C-library CI builds on `macos-14` via cmake with `CONFIG_INT128` defined (native
path, fallback not compiled), and the Rust-crate CI (`Crate-publishing.yml`) does
not run on Apple Silicon.

Happy to drop the test/CI or fold it into an existing workflow if maintainers prefer.

### DIFF
--- a/.github/workflows/int128-regression.yml
+++ b/.github/workflows/int128-regression.yml
@@ -1,0 +1,41 @@
+name: int128 fallback regression
+
+# Guards the !CONFIG_INT128 fallback in qemu/include/qemu/int128.h against the
+# __int128_t redefinition that broke real Clang builds (e.g. Apple clang on
+# arm64, and the Rust unicorn-engine-sys crate). The existing C-library CI builds
+# on macos-14 (arm64) but via cmake with CONFIG_INT128 defined, so it uses the
+# native path and never compiles the fallback; the Rust-crate CI does not run on
+# Apple Silicon. This job compiles the fallback construct directly on every
+# platform whose compiler exposes the __int128_t builtin — including arm64
+# Apple clang, where the pre-fix guard fails.
+
+on:
+  push:
+    paths:
+      - 'qemu/include/qemu/int128.h'
+      - 'tests/regress/int128_redefinition.c'
+      - '.github/workflows/int128-regression.yml'
+  pull_request:
+    paths:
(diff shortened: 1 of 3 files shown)


wtdcode:
Hi please note our PR should go against the dev branch.

Also tests go to `tests/unit`.

PR #2361, opened 2026-07-07
target/i386: Fix decoding of 0xC6/0xC7 (group 11) instructions

unknown-1-0:
According to both table A-6 in Volume 2 of Intel 64 and IA-32 Architectures Software Developer's Manual and table A-6 in Volume 3 of AMD64 Architecture Programmer's Manual, group 11 instructions are decoded as MOV only if ModRM.reg == 0.

Although Unicorn checks for that, it handles some things incorrectly:
- In register-only operand form (ModRM.mod == 3), decoding of MOV instruction also succeeds if ModRM.reg == 7. However, according to table A-6 in Volume 2 of Intel SDM, this should encode XBEGIN/XABORT, not MOV!
- Before the check, ModRM.reg field is ORed with the value of REX.R << 3, leading to an invalid opcode (#UD) exception if REX.R bit is actually set. Real Intel hardware seems to ignore the REX.R bit with group 11 instructions.

This commit fixes both issues.

### DIFF
--- a/qemu/target/i386/translate.c
+++ b/qemu/target/i386/translate.c
@@ -6009,15 +6009,12 @@ static target_ulong disas_insn(DisasContext *s, CPUState *cpu)
         ot = mo_b_d(b, dflag);
         modrm = x86_ldub_code(env, s);
         mod = (modrm >> 6) & 3;
-        reg = ((modrm >> 3) & 7) | rex_r;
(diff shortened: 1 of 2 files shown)


unknown-1-0:
Updated the tests and the commit message, fixed a typo in one of test names

unknown-1-0:
Force-pushed to sign the commit

PR #2362, opened 2026-07-07
target/i386: Reject ModRM.reg == 7 early when decoding 0xFE/0xFF (group 4 and 5) instructions

unknown-1-0:
According to both table A-6 in Volume 2 of Intel 64 and IA-32 Architectures Software Developer's Manual and table A-6 in Volume 3 of AMD64 Architecture Programmer's Manual, the opcode for ModRM.reg == 7 is reserved for both group 4 (0xFE) and 5 (0xFF) instructions.

Although Unicorn already rejects ModRM.reg == 7 early with group 4 (0xFE) instructions, that is not the case with group 5 (0xFF) instructions. The latter case is eventually handled in the `default:` block of `switch(op)`, but by the time that code is reached, some micro-ops are already generated. These may include a memory load micro-op. As a result, an invalid memory access error may be produced instead of an expected invalid instruction error.

Fix this bug by rejecting group 4 and 5 instructions with ModRM.reg == 7 before generating any micro-operations.

### DIFF
--- a/qemu/target/i386/translate.c
+++ b/qemu/target/i386/translate.c
@@ -5493,7 +5493,7 @@ static target_ulong disas_insn(DisasContext *s, CPUState *cpu)
         mod = (modrm >> 6) & 3;
         rm = (modrm & 7) | REX_B(s);
         op = (modrm >> 3) & 7;
-        if (op >= 2 && b == 0xfe) {
+        if (op == 7 || (op >= 2 && b == 0xfe)) {
             goto unknown_op;
         }
         if (CODE64(s)) {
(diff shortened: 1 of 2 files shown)


wtdcode:
For such instruction semantic related fixes, a unit test could greatly speed up my reviewing process because I can debug and see the difference. Thanks in advance if you wish adding one.

unknown-1-0:
Noted. Working on it

unknown-1-0:
Added the unit test

unknown-1-0:
I see there's a merge conflict in test_x86.c across my open PRs. Please let me know when you are ready to merge each PR, and I'll resolve the conflicts one by one.

unknown-1-0:
Resolved the conflict. Moved the test between existing ones to prevent future conflicts. Also the commit is signed now.

PR #2370, opened 2026-07-22
target/i386: Clear `s->x86_64_hregs` when decoding LAHF/SAHF with a REX prefix

unknown-1-0:
According to both Volume 2 of Intel 64 and IA-32 Architectures Software Developer's Manual and Volume 3 of AMD64 Architecture Programmer's Manual, the LAHF (0x9f) and SAHF (0x9e) instructions always access the AH register. The presence of a REX prefix in 64-bit mode does not alter this behavior, and uniform byte register addressing (SPL) must not be applied.

The bug occurs because the translator blindly sets `s->x86_64_hregs` to `true` upon encountering any REX prefix. During translation of LAHF/SAHF, `gen_op_deposit_reg_v()` is called, which relies on the internal helper `byte_reg_is_xH()` to check whether to address a high byte register (AH). Since `s->x86_64_hregs` is `true`, `byte_reg_is_xH(s, 4)` returns `false`. This causes the translator to bypass the high-byte logic (`reg - 4` with an 8-bit shift) and instead maps index 4 directly to `tcg_ctx->cpu_regs[4]`, leading to SPL being accessed instead of AH.

Fix this by ensuring `s->x86_64_hregs` remains `false` when decoding LAHF/SAHF.

### DIFF
--- a/qemu/target/i386/translate.c
+++ b/qemu/target/i386/translate.c
@@ -5037,8 +5037,10 @@ static target_ulong disas_insn(DisasContext *s, CPUState *cpu)
             rex_r = (rex_byte & 0x4) << 1;
             s->rex_x = (rex_byte & 0x2) << 2;
             REX_B(s) = (rex_byte & 0x1) << 3;
-            /* select uniform byte register addressing */
-            s->x86_64_hregs = true;
+            /* select uniform byte register addressing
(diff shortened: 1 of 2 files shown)


unknown-1-0:
New tests were added between existing ones to prevent merge conflicts later.

unknown-1-0:
Force-pushed to sign the commit

PR #2372, opened 2026-07-23
tests/x86: add regression test for TCG temps overflow on dense x87 blocks

ioma8:
## Summary

**Reworked per review feedback** (originally this raised `TCG_MAX_TEMPS` to 1024 on 64-bit hosts): this PR is now **test-only**, rebased onto `dev`, and adds a real-world x86 regression test for the TCG temps overflow.

@PhilippTakacs was right that raising the limit only moves the threshold. I verified against #2307:

| Build | `test_x86_tcg_temp_limit_x87_block` |
|---|---|
| stock `dev` (`TCG_MAX_TEMPS` 512) | SIGSEGV in `tcg_emit_op` |
| `dev` + #2307, limit unchanged | passes, `UC_ERR_OK` |

So #2307's overflow error handling fixes the underlying problem; the limit raise is unnecessary and has been dropped.

## The bug this test covers

During translation of long basic blocks, `nb_temps` can exceed `TCG_MAX_TEMPS`. `tcg_temp_alloc()` only guards with `tcg_debug_assert`, a no-op in release builds, so `temps[]` silently overflows into the adjacent `ops` queue head in `TCGContext`, and the next `tcg_emit_op()` crashes on a NULL `tqh_last`. Same failure class as #2240 (MIPS), reproduced here on x86.

The test embeds a dense x87 basic block (matrix math from a 32-bit Windows game, found via [OpenSWE1R](https://github.com/OpenSWE1R/openswe1r) on Apple Silicon) that reliably drives `nb_temps` past 512 (`nb_temps == 513`, `nb_ops == 3830` at crash).

## Merge order

This complements #2307 with an x86 reproducer (it currently only has a MIPS test) and **must land after it** — on current `dev` the test SIGSEGVs, so CI will fail until #2307 is merged. Happy to have the test cherry-picked into #2307 directly instead, if that is preferred.

## Test plan

- [x] `./test_x86 test_x86_tcg_temp_limit_x87_block` on arm64 macOS: SIGSEGV on stock `dev`, passes on `dev` + #2307


wtdcode:
Hi, could you please send your patch to dev branch instead?

PhilippTakacs:
Just increasing the limit doesn't solve the problem, see  #2240 #2305 and #2307.

ioma8:
> Hi, could you please send your patch to dev branch instead?

Yes changed the target branch.

ioma8:
> Just increasing the limit doesn't solve the problem, see #2240 #2305 and #2307.

Yes I see. My workaround fixed it for the game to run well, but isnt root cause solution. #2307 seems to fix it at underlying level and I validated it works. Still I think my PR contains good real-data x86/x87 reproducer test. Do you want me to keep the PR open / do you wanna cherrypick it?

PR #2378, opened 2026-07-31
add block hook with instruction count

PhilippTakacs:
TODO disallow overlapping with block hook, write tests

### DIFF
--- a/uc.c
+++ b/uc.c
@@ -1927,6 +1927,30 @@ uc_err uc_hook_add(uc_engine *uc, uc_hook *hh, int type, void *callback,
 
     UC_INIT(uc);
 
+    int block_type = type & (UC_HOOK_BLOCK | UC_HOOK_BLOCK_ICOUNT);
+    if (block_type == (UC_HOOK_BLOCK | UC_HOOK_BLOCK_ICOUNT)) {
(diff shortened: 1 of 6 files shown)


PhilippTakacs:
@wtdcode it would be nice to have this in the next release.

uc.c
```diff
@@ -2013,7 +2013,24 @@ uc_err uc_hook_add(uc_engine *uc, uc_hook *hh, int type, void *callback,
         return UC_ERR_OK;
     }
 
-    if (type & UC_HOOK_CODE || type & UC_HOOK_BLOCK) {
+    if (type & UC_HOOK_BLOCK || type & UC_HOOK_BLOCK_ICOUNT) {
```
wtdcode: Wait `UC_HOOK_CODE` is missing from this branch?

PhilippTakacs: This is to check if a `UC_HOOK_BLOCK` and a `UC_HOOK_BLOCK_ICOUNT` overlaps. I use this because this way the hook is simpler to implement, but doesn't allow a `UC_HOOK_BLOCK` and `UC_HOOK_BLOCK_ICOUNT` at the same tb. Overlapping with a `UC_HOOK_CODE` is not a problem.

Issue #1477, opened 2021-11-02
[RISC-V] `pc` is not increased after execute `ecall`


cla7aye15I4nd:
I am trying to simulate a simple riscv helloworld example and hook the syscall, but it seems like the uc2 do not increase the pc after executing `ecall`.

Here is my uc code:
```python
import lief

from capstone import Cs, CS_ARCH_RISCV, CS_MODE_RISCV64

from unicorn.riscv_const import *
from unicorn import Uc, UC_ARCH_RISCV, UC_MODE_RISCV64, UC_HOOK_INTR

uc = Uc(UC_ARCH_RISCV, UC_MODE_RISCV64)
cs = Cs(CS_ARCH_RISCV, CS_MODE_RISCV64)

elf = lief.parse('hello')
code = bytearray(elf.get_section('.text').content)

uc.mem_map(0, 0x100000)
uc.mem_write(elf.entrypoint, bytes(code) + b'hello world\n!')

def hook_intr(uc, intno, user_data):
    print(f'Interrupt number: {intno} (pc = {hex(uc.reg_read(UC_RISCV_REG_PC))})')

for i in cs.disasm(code, elf.entrypoint):
    print("0x%x:\t%s\t%s" %(i.address, i.mnemonic, i.op_str))
    
uc.hook_add(UC_HOOK_INTR, hook_intr)
uc.emu_start(elf.entrypoint, 0, count=10)
```

and this is my riscv code:
```asm
.global _start

_start: addi  a0, x0, 1
        la    a1, helloworld
        addi  a2, x0, 13    
        addi  a7, x0, 64    
        ecall
        addi    a0, x0, 0
        addi    a7, x0, 93
        ecall

.data
helloworld:      .ascii "Hello RISCV!\n"
```

The output of the simulation is:
```
0x100b0:	addi	a0, zero, 1
0x100b4:	auipc	a1, 1
0x100b8:	addi	a1, a1, 0x20
0x100bc:	addi	a2, zero, 0xd
0x100c0:	addi	a7, zero, 0x40
0x100c4:	ecall	
0x100c8:	mv	a0, zero
0x100cc:	addi	a7, zero, 0x5d
0x100d0:	ecall	
Interrupt number: 8 (pc = 0x100c4)
Interrupt number: 8 (pc = 0x100c4)
Interrupt number: 8 (pc = 0x100c4)
Interrupt number: 8 (pc = 0x100c4)
Interrupt number: 8 (pc = 0x100c4)
[...]

PR #2384, opened 2026-08-09
fix: preserve riscv trap pc

Jvlegod:
This PR fixes RISC-V exception PC handling in interrupt hooks.

Previously, Unicorn unconditionally advanced the RISC-V PC by 4 before invoking `UC_HOOK_INTR`. That behavior was added to solve `ecall` resume problem reported in #1477, but it also made interrupt hooks observe the wrong PC for other exceptions, as reported in #1990.

This change preserves the faulting instruction address while the interrupt hook is running, and only advances PC to the saved next PC after a handled RISC-V `ecall`.

Note: `ebreak` will follow the `UC_ERR_INSN_INVALID` path is not affected by this change.

### DIFF
--- a/qemu/accel/tcg/cpu-exec.c
+++ b/qemu/accel/tcg/cpu-exec.c
@@ -390,10 +390,6 @@ static inline bool cpu_handle_exception(CPUState *cpu, int *ret)
         CPUMIPSState *env = &(MIPS_CPU(cpu)->env);
         env->active_tc.PC = uc->next_pc;
 #endif
-#if defined(TARGET_RISCV)
-        CPURISCVState *env = &(RISCV_CPU(uc->cpu)->env);
-        env->pc += 4;
-#endif
 #if defined(TARGET_SPARC)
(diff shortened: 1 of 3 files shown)


Kreijstal:
I closed #2400 in favor of this PR.

One part from it is still useful. For instruction, load, and store address faults, RISC-V places the faulting address in `mtval`. `UC_HOOK_INTR` should be able to read that value.

I split the fix into a commit based directly on this PR:

https://github.com/Kreijstal/unicorn/commit/c3cb1677900dbeb1d144fe904dd2386bd06e355b

It does not change this PR's PC handling. I built it with `UNICORN_ARCH=riscv`, and all 30 RISC-V tests pass. Feel free to take the commit, or I can submit it separately after this PR lands.

Jvlegod:
Hello @wtdcode.

#2400 was closed in favor of this PR, @Kreijstal extracted the remaining change into c3cb167. I have verified that it applies cleanly on top of this branch.  If you think this PR is suitable, I will `cherry-pick` the new code.

:>

PR #2394, opened 2026-08-12
riscv: reject odd instruction addresses at fetch time

### LINKED ISSUE #2390: RISC-V: execution starts from an odd PC (IALIGN=16 violation)
# RISC-V: execution starts from an odd PC (IALIGN=16 violation)

## Rechecked revision

## API boundary

Legal RISC-V control transfers obey IALIGN. This report targets the Unicorn
public entry boundary: a caller can pass an odd `begin` address to
`uc_emu_start` or write an odd `UC_RISCV_REG_PC`, after which the fetch path
currently decodes from that address. The reproduction therefore tests an
externally supplied invalid emulator state, not an odd target produced by a
legal guest branch. If maintainers prefer rejecting the state in an API setter
rather than at fetch time, the same public witness can be used for that design.

## Public reproducer

The minimal Python witness is available at
https://github.com/carlosqwqqwq/unicorn-riscv-repros/tree/main/u025.
It starts the public U54 model at `0x10001` and records the executed PCs and
the returned Unicorn error.

## Expected behavior

The first fetch from an odd address must raise the instruction-address-
misaligned exception before any guest instruction executes. An aligned entry
point and legal compressed/32-bit streams must remain unchanged.

`riscv_tr_translate_insn` fetches at `ctx->base.pc_next` without checking bit
0. The proposed change checks the address before fetch and raises the existing
instruction-address-misaligned exception. The focused regression is included
in [PR #2394](https://github.com/unicorn-engine/unicorn/pull/2394).

carlosqwqqwq:
Fixes #2390

## Problem

IALIGN=16 forbids instruction addresses with bit 0 set (even with the C extension). Unicorn accepts an externally supplied odd PC at the emulator entry and fetches at that address; branch-target checks in `trans_rvi.inc.c` do not cover the entry path.

## Change

`qemu/target/riscv/translate.c` `riscv_tr_translate_insn`: when `ctx->base.pc_next & 1`, raise `RISCV_EXCP_INST_ADDR_MIS` before any fetch. The focused unit test is `test_riscv64_odd_pc` in `tests/unit/test_riscv.c`.

## Validation

- Rebased onto `dev`.
- The public U54 witness traps before the first fetch.
- Aligned entry and legal compressed/32-bit controls remain functional.

Public reproducer:
https://github.com/carlosqwqqwq/unicorn-riscv-repros/tree/main/u025

### DIFF
--- a/qemu/target/riscv/translate.c
+++ b/qemu/target/riscv/translate.c
@@ -849,7 +849,13 @@ static void riscv_tr_translate_insn(DisasContextBase *dcbase, CPUState *cpu)
     bool insn_hook = false;
 
     // Unicorn: end address tells us to stop emulation
-    if (uc_addr_is_exit(uc, ctx->base.pc_next)) {
+    if (ctx->base.pc_next & 1) {
+        /*
+         * IALIGN=16: an odd instruction address is always misaligned,
+         * even with the C extension.
+         */
+        gen_exception_inst_addr_mis(ctx);
+    } else if (uc_addr_is_exit(uc, ctx->base.pc_next)) {
         // Unicorn: We have to exit current execution here.
         dcbase->is_jmp = DISAS_UC_EXIT;
     } else {
--- a/tests/unit/test_riscv.c
+++ b/tests/unit/test_riscv.c
@@ -55,6 +55,19 @@ static void test_riscv64_nop(void)
     OK(uc_close(uc));
 }
 
+static void test_riscv64_odd_pc(void)
+{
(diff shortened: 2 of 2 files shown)


carlosqwqqwq:
Thank you very much for taking the time to review this PR. We have rebased it onto dev and added focused coverage for odd instruction addresses. Could you please let us know what further code, test, or PR changes you would recommend? We will gladly adjust the patch to match the project guidelines.

PR #2395, opened 2026-08-12
riscv: gate D-width atomic translators on the A extension

### LINKED ISSUE #2389: RISC-V: D-width atomic instructions retire on harts without the A extension
# RISC-V: D-width atomic instructions retire without the A extension

## Rechecked revision

## Problem

The eleven RV64 D-width atomic translators omit the A-extension gate used by
their W-width siblings. On a hart without A, `lr.d`, `sc.d`, and `amo*.d`
therefore retire and may perform memory effects instead of trapping.

## Public reproducer

The witness, controls, and the explicit A-clear `riscv_any` build-profile patch
are available at
https://github.com/carlosqwqqwq/unicorn-riscv-repros/tree/main/u030.
Apply `patches/no-a-any.patch` to a clean public Unicorn checkout, build it,
set `LIBUNICORN_PATH` to that build, and run `u030/run.sh`. The D-width
witness must trap; the W-width control must retain its existing trap behavior.
The profile patch is public and only constructs the no-A test configuration;
it does not hide any source change used by the report.

## Expected behavior

Every `LR`/`SC`/`AMO` encoding requires A unless another explicitly supported
extension owns that operation. A-enabled models remain legal.

Add `REQUIRE_EXT(ctx, RVA);` to each D-width atomic translator before its
memory helper runs. The change is included in [PR #2395](https://github.com/unicorn-engine/unicorn/pull/2395).

carlosqwqqwq:
Fixes #2389

## Problem

The eleven RV64 D-width atomic translators in `trans_rva.inc.c` (`trans_lr_d`, `trans_sc_d`, `trans_amoswap_d`, `trans_amoadd_d`, `trans_amoxor_d`, `trans_amoand_d`, `trans_amoor_d`, `trans_amomin_d`, `trans_amomax_d`, `trans_amominu_d`, `trans_amomaxu_d`) emit their memory operations without the `REQUIRE_EXT(ctx, RVA)` check that every W-width sibling performs, so they retire and modify memory on harts without A instead of raising an illegal-instruction exception.

## Change

Add `REQUIRE_EXT(ctx, RVA);` as the first statement of each of the eleven D-width translators. The check traps at the first D-width atomic before any memory effect, matching the W-width translators and upstream QEMU. The focused unit test is `test_riscv64_dwidth_atomic_without_a` in `tests/unit/test_riscv.c`.

## Validation

- Rebased onto `dev` and built the RISC-V targets.
- The public `u030` suite includes the A-clear profile patch, a D-width witness, and a W-width control.
- The D-width witness traps on the A-clear profile; the W-width control retains its existing behavior.

Public reproducer and explicit profile setup:
https://github.com/carlosqwqqwq/unicorn-riscv-repros/tree/main/u030

### DIFF
--- a/qemu/target/riscv/insn_trans/trans_rva.inc.c
+++ b/qemu/target/riscv/insn_trans/trans_rva.inc.c
@@ -172,56 +172,67 @@ static bool trans_amomaxu_w(DisasContext *ctx, arg_amomaxu_w *a)
 
 static bool trans_lr_d(DisasContext *ctx, arg_lr_d *a)
 {
+    REQUIRE_EXT(ctx, RVA);
     return gen_lr(ctx, a, MO_ALIGN | MO_TEQ);
 }
 
 static bool trans_sc_d(DisasContext *ctx, arg_sc_d *a)
 {
+    REQUIRE_EXT(ctx, RVA);
     return gen_sc(ctx, a, (MO_ALIGN | MO_TEQ));
 }
 
 static bool trans_amoswap_d(DisasContext *ctx, arg_amoswap_d *a)
 {
+    REQUIRE_EXT(ctx, RVA);
     return gen_amo(ctx, a, &tcg_gen_atomic_xchg_tl, (MO_ALIGN | MO_TEQ));
 }
 
 static bool trans_amoadd_d(DisasContext *ctx, arg_amoadd_d *a)
 {
+    REQUIRE_EXT(ctx, RVA);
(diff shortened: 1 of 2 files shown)


carlosqwqqwq:
Thank you very much for taking the time to review this PR. We have rebased it onto dev and applied the extension check consistently to the D-width atomic translators, with focused validation of the affected behavior. Could you please let us know what further code, test, or PR changes you would recommend? We will gladly adjust the patch to match the project guidelines.

PR #2396, opened 2026-08-12
riscv: keep mstatus.MPP=M after mret on harts without U-mode

### LINKED ISSUE #2391: RISC-V: mret sets mstatus.MPP=U on harts without U-mode
# RISC-V: `mret` sets `mstatus.MPP=U` without U-mode

## Rechecked revision

## Problem

`helper_mret` always writes `mstatus.MPP = PRV_U`. On a hart whose ISA omits
U-mode, M is the least-privileged supported mode, so the post-`mret` MPP state
must remain M rather than becoming U.

## Public reproducer

The witness and the explicit U-clear `riscv_any` build-profile patch are
available at
https://github.com/carlosqwqqwq/unicorn-riscv-repros/tree/main/u034.
Apply `patches/no-u-any.patch` to a clean public Unicorn checkout, build it,
set `LIBUNICORN_PATH` to that build, and run `u034/run.sh`. The no-U run must
read back `MPP=3`; a U-enabled control remains unchanged.

The profile patch is public and only constructs the no-U test configuration;
it does not hide any source change used by the proposed fix.

## Expected behavior

`mret` must select the least-privileged mode implemented by the hart: U when
U-mode exists, otherwise M.

Select `PRV_U` only when `RVU` is present and select `PRV_M` otherwise. The
focused change is included in [PR #2396](https://github.com/unicorn-engine/unicorn/pull/2396).

carlosqwqqwq:
Fixes #2391

## Problem

`helper_mret` unconditionally writes `mstatus.MPP = PRV_U`. On a hart whose MISA does not include U, the least-privileged supported mode is M, so after `mret` the MPP field must read 3, not 0. The reproducer runs on a locally built no-U model library (default `riscv_any` with RVU cleared; stock E51/U54 keep U as the control).

## Change

`qemu/target/riscv/op_helper.c` `helper_mret` selects `PRV_U` only when `RVU` is present and `PRV_M` otherwise. The focused unit test is `test_riscv64_mret_without_u` in `tests/unit/test_riscv.c`.

## Validation

- Rebased onto `dev` and built the RISC-V targets.
- The public `u034` suite includes the U-clear profile patch and an explicit U-enabled control.
- The no-U witness reads back MPP=M; U-enabled behavior remains unchanged.

Public reproducer and explicit profile setup:
https://github.com/carlosqwqqwq/unicorn-riscv-repros/tree/main/u034

### DIFF
--- a/qemu/target/riscv/op_helper.c
+++ b/qemu/target/riscv/op_helper.c
@@ -149,7 +149,8 @@ target_ulong helper_mret(CPURISCVState *env, target_ulong cpu_pc_deb)
         MSTATUS_MIE : MSTATUS_UIE << prev_priv,
         get_field(mstatus, MSTATUS_MPIE));
     mstatus = set_field(mstatus, MSTATUS_MPIE, 1);
-    mstatus = set_field(mstatus, MSTATUS_MPP, PRV_U);
+    mstatus = set_field(mstatus, MSTATUS_MPP,
+                        riscv_has_ext(env, RVU) ? PRV_U : PRV_M);
 #ifdef TARGET_RISCV32
     env->mstatush = set_field(env->mstatush, MSTATUS_MPV, 0);
 #else
--- a/tests/unit/test_riscv.c
+++ b/tests/unit/test_riscv.c
@@ -55,6 +55,30 @@ static void test_riscv64_nop(void)
     OK(uc_close(uc));
 }
 
+static void test_riscv64_mret_without_u(void)
+{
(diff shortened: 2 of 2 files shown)


carlosqwqqwq:
Thank you very much for taking the time to review this PR. We have rebased it onto dev and updated the mret MPP handling for harts without U-mode, with focused validation of the affected behavior. Could you please let us know what further code, test, or PR changes you would recommend? We will gladly adjust the patch to match the project guidelines.

Issue #1902, opened 2023-10-27
Add UC_HOOK_INSN support for UC_X86_INS_RDTSC and UC_X86_INS_RDRAND


Z4ee:
The current version of the Unicorn Engine does not provide hook support for the `UC_X86_INS_RDTSC` and `UC_X86_INS_RDRAND` instructions. These instructions are crucial for applications requiring precise time measurement or random number generation. This proposal suggests adding hook support for these instructions, allowing users to have more flexible control over code execution.

```
uc_hook_add(uc, &hook, UC_HOOK_INSN, hook_rdtsc, nullptr, 1, 0, UC_X86_INS_RDTSC); // Return: UC_ERR_HOOK
uc_hook_add(uc, &hook, UC_HOOK_INSN, hook_rdrand, nullptr, 1, 0, UC_X86_INS_RDRAND); // Return: UC_ERR_HOOK
```

PR #2398, opened 2026-08-14
feat: add x86 RDRAND instruction hook

Jvlegod:
Hello maintainers,

This PR close https://github.com/unicorn-engine/unicorn/issues/1902

We have support `UC_X86_INS_RDTSC`, now this PR adds `UC_X86_INS_RDRAND`'s hook.


PhilippTakacs:
The API you propose is a bit limited. I would expect a way to change the generated random. Your change just call the hook and after that use the normal implementation. So there is no simple way to change the result.

Jvlegod:
> I would expect a way to change the generated random.

I have now added a hook that executes after the instruction successfully completes.

This hook allows for controlling the `CF` flag (indicating whether the instruction succeeded) and modifying the random value.

> The CF flag is set according to the result (see the “Operation” section above). The OF, SF, ZF, AF, and PF flags are set to 0.

- https://www.felixcloutier.com/x86/rdrand

PR #2375, opened 2026-07-28
Pr2349 rebased

wtdcode:
#2349 Upgrade our QEMU fork to 7.1.22.

We will finally chase up to the upstream QEMU.

PR #2399, opened 2026-08-17
target/i386: Mark VPERMILPS as not valid with prefix 0

unknown-1-0:
Original commit message below:
There are a small set of binary SSE insns which have no MMX equivalent, which we create the gen functions for with the BINARY_INT_SSE() macro.  This forwards to gen_binary_int_sse() with a NULL pointer for 'mmx'.

For almost all of these insns we correctly mark them in the decode table as not permitting a zero prefix byte; however we got this wrong for VPERMILPS, with the result that a bogus instruction would get through the decode checks and end up in gen_binary_int_sse() trying to call a NULL pointer.

Correct the decode table entry for VPERMILPS so that we get the expected #UD exception.

In the x86 SDM, table A-4 "Three-byte Opcode Map: 08H-FFH (First Two Bytes are 0F 38H)" confirms that there is no pfx 0 version of VPERMILPS.

This is a backport of the following upstream QEMU commit: qemu/qemu@ebd9ea2947d88f237e20333fe547ca8817d0b0ee
Link to the original patch: https://lore.kernel.org/qemu-devel/20251120001649.1085913-4-pbonzini@redhat.com/
Link to my original issue: https://gitlab.com/qemu-project/qemu/-/work_items/3199


unknown-1-0:
TL;DR: Fixes a crash in x86 decoder when decoding VPERMILPS instruction without any prefixes, which was introduced in #2375

PR #2400, opened 2026-08-19
fix: riscv PC advance for compressed instructions and fetch faults

Kreijstal:
`cpu_handle_exception` unconditionally does `env->pc += 4` for RISC-V.  This corrupts the PC for 16-bit compressed instructions and for instruction access/page faults (where `env->pc` already holds the faulting address and must not be advanced).  It also leaves the faulting address unavailable to `UC_HOOK_INTR` callbacks.

Changes:

- Advance `env->pc` by the real instruction size (2 or 4 bytes, determined by the low two opcode bits) for execution-time exceptions.
- Skip the advance for `INST_ADDR_MIS`, `INST_ACCESS_FAULT`, `INST_PAGE_FAULT`.
- Mirror `riscv_cpu_do_interrupt` and set `env->mbadaddr = env->badaddr` for every address-bearing exception, so the callback can recover from e.g. PMP faults.

Verified by booting a RISC-V firmware under Unicorn whose syscall/return path is full of compressed instructions and relies on lazy PMP fault recovery.

### DIFF
--- a/qemu/accel/tcg/cpu-exec.c
+++ b/qemu/accel/tcg/cpu-exec.c
@@ -391,8 +391,8 @@ static inline bool cpu_handle_exception(CPUState *cpu, int *ret)
         env->active_tc.PC = uc->next_pc;
 #endif
 #if defined(TARGET_RISCV)
-        CPURISCVState *env = &(RISCV_CPU(uc->cpu)->env);
-        env->pc += 4;
+        riscv_cpu_prepare_exception_pc(&(RISCV_CPU(uc->cpu)->env),
+                                       cpu->exception_index);
 #endif
 #if defined(TARGET_SPARC)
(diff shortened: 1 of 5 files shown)


Kreijstal:
Closing this as superseded by #2384. That PR preserves the faulting PC while the hook runs and advances to the translator-recorded next PC only after a handled `ecall`, which also covers compressed instructions.

One part is missing from #2384. For instruction, load, and store address faults, RISC-V places the faulting address in `mtval`. `UC_HOOK_INTR` should be able to read that value. I split the fix into a patch on top of #2384:

https://github.com/Kreijstal/unicorn/commit/c3cb1677900dbeb1d144fe904dd2386bd06e355b

I built that commit with RISC-V enabled. All 30 tests from #2384 pass.

PR #2403, opened 2026-08-20
RISC-V: add optional architectural trap entry

### LINKED ISSUE #2402: RISC-V: allow guest trap handlers to receive exceptions
Unicorn does not perform RISC-V trap entry when an instruction raises an exception.

For example:

```c
uc_open(UC_ARCH_RISCV, UC_MODE_RISCV64, &uc);

uc_mem_write(uc, 0x1000, "\x73\x00\x00\x00", 4); /* ecall */

uint64_t mtvec = 0x1800;
uc_reg_write(uc, UC_RISCV_REG_MTVEC, &mtvec);

uc_emu_start(uc, 0x1000, 0, 0, 1);
```

After an M-mode `ecall`, RISC-V requires:

```text
pc     = 0x1800
mepc   = 0x1000
mcause = 11
mode   = M
```

Unicorn instead calls `UC_HOOK_INTR`, advances `pc` to `0x1004`, and does not enter the handler at `mtvec`.

The reason is that `cpu_handle_exception()` handles the exception without calling QEMU's `riscv_cpu_do_interrupt()`. That function already implements the required CSR, privilege, and PC changes.

Could Unicorn provide an opt-in control such as:

```c
uc_ctl_set_architectural_exceptions(uc, true);
```

For RISC-V, enabling it would call `riscv_cpu_do_interrupt()` and then call `UC_HOOK_INTR` with the original exception number. The default would stay unchanged for compatibility.

Should this be a generic control, with RISC-V as its first implementation, or a RISC-V-specific control?

I have a patch and unit test for both the existing default behavior and the opt-in behavior.

Related: #1805

Kreijstal:
RISC-V exceptions currently call `UC_HOOK_INTR` and advance `pc` without updating the trap CSRs or jumping to `mtvec`.

This adds an opt-in control:

```c
uc_ctl_set_architectural_exceptions(uc, true);
```

When enabled on RISC-V, `cpu_handle_exception()` calls QEMU's existing `riscv_cpu_do_interrupt()` before invoking `UC_HOOK_INTR`. The hook receives the original exception number.

The option is disabled by default, so existing RISC-V behavior does not change. Other architectures return `UC_ERR_ARCH` when the control is used.

The new unit test checks that an M-mode `ecall` sets `pc`, `mepc`, `mcause`, and the privilege mode. The existing `ecall` test still checks the default PC-advance behavior.

Fixes #2402

### DIFF
--- a/include/uc_priv.h
+++ b/include/uc_priv.h
@@ -335,6 +335,7 @@ struct uc_struct {
     TargetPageBits *init_target_page;
     int target_bits; // User defined page bits by uc_ctl
     int cpu_model;
+    bool architectural_exceptions;
     BounceBuffer bounce;                // qemu/cpu-exec.c
     volatile sig_atomic_t exit_request; // qemu/cpu-exec.c
     /* qemu/accel/tcg/cpu-exec-common.c */
--- a/include/unicorn/unicorn.h
+++ b/include/unicorn/unicorn.h
@@ -605,6 +605,11 @@ typedef enum uc_control_type {
     // controle if context_save/restore should work with snapshots
     // Write: @args = (int)
     UC_CTL_CONTEXT_MODE,
+    // Let the target CPU perform architectural exception entry before
+    // UC_HOOK_INTR is invoked. Disabled by default for compatibility.
+    // Write: @args = (int)
+    // Read: @args = (int*)
+    UC_CTL_UC_ARCHITECTURAL_EXCEPTIONS,
 } uc_control_type;
 
 /*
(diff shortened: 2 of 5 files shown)


wtdcode:
This looks good, but honestly, the `uc_ctl` does not seem like the correct place. What's your use case?

Kreijstal:
> This looks good, but honestly, the `uc_ctl` does not seem like the correct place. What's your use case?

trying to emulate  full-system riscv32 firmware, It simulates SiFive HiFive1 Rev B development board, using the SiFive FE310.

QEMU cannot do this, so I thought I could it it with unicorn.

wtdcode:
> > This looks good, but honestly, the `uc_ctl` does not seem like the correct place. What's your use case?
> 
> trying to emulate full-system riscv32 firmware, It simulates SiFive HiFive1 Rev B development board, using the SiFive FE310.
> 
> QEMU cannot do this, so I thought I could it it with unicorn.

Sorry I mean why it should not advance PC?

Kreijstal:
"When a trap is taken into M-mode, mepc is written with the virtual address of the instruction that was
interrupted or that encountered the exception." page 49 https://docs.riscv.org/reference/isa/_attachments/riscv-privileged.pdf

so the +4 happens in the m trap handler which happens on the firmware

unicorn assumes you have no firmware, so it does it for you, but if you want to emulate the firmware, unicorn should allow you to do the +4  (in the firmware), or more strictly I want to model m-trap handling in riscv itself

PR #2412, opened 2026-09-14
feat: Add support for x86 MMX Registers

06needhamt:
The Python binding already uses a 64-bit value for MM0-MM7, but the native x86 register dispatch lacks their read/write cases. Access the physical FPReg MMX union without applying x87 TOP, changing tags, or altering the upper x87 bits. This fixes scalar, batch, and context access for all bindings.

Validated on Debian Linux x86-64 / Python 3.12


unknown-1-0:
Hi, could you please rebase this to the `dev` branch? According to the README, all pull requests should target `dev`. Thanks!

06needhamt:
@unknown-1-0 I have rebased the PR against dev. I must have missed that in the README.

tests/unit/test_x86.c
```diff
@@ -2732,6 +2732,138 @@ static void test_x86_lock_btc_reg(void)
     OK(uc_close(uc));
 }
 
+static const uc_mode mmx_modes[] = {
```
unknown-1-0: I think this should be called `x86_modes`, since CPU mode is an x86 setting, not a MMX one
```suggestion
static const uc_mode x86_modes[] = {
```

06needhamt:
Is there anything else I can do to move this pull request forward?

PR #2413, opened 2026-09-16
cmake: only pass -latomic when libatomic actually exists

pkubaj:
ATOMIC_LINKAGE_FIX unconditionally adds -latomic on arm, mips, sparc,
ppc, riscv and s390 hosts.  On systems that do not ship libatomic
(e.g. FreeBSD, where clang inlines the 128-bit atomics), -latomic makes
the "can build an executable" link test in qemu/configure fail, so
configure aborts before writing config-host.mak.  CONFIG_INT128 then
stays undefined and the struct Int128 fallback breaks the build:

  qemu/include/qemu/int128.h:150:16: error: typedef redefinition with
  different types ('Int128' (aka 'struct Int128') vs '__int128')

Probe for libatomic with check_library_exists() (the same check LLVM
uses) and drop ATOMIC_LINKAGE_FIX when it is absent.  Fixes the build
on FreeBSD/powerpc64 and FreeBSD/powerpc64le.


unknown-1-0:
Hi, could you please rebase this to the `dev` branch? According to the README, all pull requests should target `dev`. Thanks!

